PR #9, opened 2019-01-31
review current scripts

SophieHerbst:
moving towards a MWE of the pipeline, using `sample`


massich:
Thats fine with me. Merging this when green.

PR #11, opened 2019-01-31
rename files

SophieHerbst:
renamed files according to #10


massich:
we didn't change the `Makefile` this will break the CIs.
Can you fix it? thx

SophieHerbst:
can you do it?

PR #15, opened 2019-02-04
update script 01–02 and config

SophieHerbst:
add option to generate plots
read bad channels from config
revise description
add plot to 02-maxfilter


01-import_and_filter.py
```diff
@@ -44,7 +50,21 @@ def run_filter(subject):
...
+        # XXX if we add multiple runs, this should probably plot an appended
+        # version of the data
+        if config.plot:
+            # plot raw data
+            figure = raw.plot(n_channels = 50,butterfly=True, group_by='position') 
+            figure.show()
+            
+            # plot power spectral densitiy
+            figure = raw.plot_psd(area_mode='range', tmin = 10.0, tmax = 100.0,
+                                  fmin = 0., fmax = 50., average=True)
```
agramfort: you need to install a pep8 checker in your editor.

02-apply_maxwell_filter.py
```diff
@@ -2,24 +2,24 @@
...
-XXX: this needs update: congig.mf_reference_run, config.mf_st_duration
-config.mf_cal_fname, config.mf_ctc_fname
-
-The data are Maxwell filtered using tSSS and movement compensation.
-
+  
+The data are Maxwell filtered using SSS or tSSS (if config.mf_st_duration is not None)
+and movement compensation.
 Using tSSS with a short duration can be used as an alternative to highpass
 filtering. Here we will use the default (10 sec) and a short window (1 sec).
```
agramfort: I don't think this is true here anymore.

02-apply_maxwell_filter.py
```diff
@@ -2,24 +2,24 @@
...
+config.mf_ctc_fname  and config.mf_cal_fname. 
 
-The data are also lowpass filtered at 40 Hz using linear-phase FIR filter with
-delay compensation. The transition bandwidth is automatically defined. See
-`Background information on filtering <http://mne-tools.github.io/dev/auto_tutorials/plot_background_filtering.html>`_
-for more. The filtered data are saved to separate files to the subject's'MEG'
-directory.
+# XXX – do we?
+Here for consistency we exploit the MaxFilter log files for
+determining the bad channels.
```
agramfort: do we don't as we don't have this information unless we did run maxfilter from elekta before.

02-apply_maxwell_filter.py
```diff
@@ -57,7 +57,13 @@ def run_maxwell_filter(subject):
             destination=destination)
 
         raw_sss.save(raw_fname_out, overwrite=True)
-
+        
+        # XXX if we add multiple runs, this should probably plot an appended
+        # version of the data
+        if config.plot:
+            # plot maxfiltered data
+            figure = raw_sss.plot(n_channels = 50,butterfly=True, group_by='position')
```
agramfort: pep8

Makefile
```diff
@@ -3,7 +3,7 @@ PYTHON = python
 help:
 	@echo "Please use \`make <target>' where <target> is one of"
 	@echo "  fetch         to fetch the data"
-	@echo "  preproces      to run all scripts after anatomy. LFREQ must be set for highpass cutoff"
+	@echo "  preproces       run sensor space preprocessing from import to epochs"
```
agramfort: preproces -> preprocess

also fix indent

SophieHerbst: I tried to rename every instance of preproces in the makefile but that broke the CI. Is there another file I need to edit?

agramfort:
git grep preproces

should help you

>

02-apply_maxwell_filter.py
agramfort: can you clean this one up too?

SophieHerbst: Yes, if config.mf_st_duration ~= None it runs tSSS

agramfort:
but it says 'here we will use ..." I don't think we do tSSS by default here

>

PR #17, opened 2019-02-06
 make 01-import_and_filter take multiple runs

SophieHerbst:
- if there are several raw fils with 'runXX'  in the filename they are all read
still works if there is only one file without 'run'
(can be extended to the following scripts if ok, concatenate after maxfilter)
- added plot for events in 03-extract_events


SophieHerbst:
Crap. Why does it show all the commits from my last PR, too?  
Only 1172840 and 2e4766c are new.

agramfort:
you have a conflict in your file 01
you should have branched from current master. it's ok except that you opened a PR from master and not from a feature branch :(

To fix it just do

git pull upstream master

git push origin master

let me know if you need help

01-import_and_filter.py
```diff
@@ -3,32 +3,65 @@
...
-    # XXX : put the study-specific names in the config file
+
     meg_subject_dir = op.join(config.meg_dir, subject)
-    raw_fnames_in = [op.join(meg_subject_dir, '%s_audvis_raw.fif' % subject)]
-    raw_fnames_out = [op.join(meg_subject_dir, '%s_audvis_filt_raw.fif' % subject)]
 
+    # try to find multiple runs
+    raw_fnames_in = sorted(glob.glob(op.join(meg_subject_dir,
+                                             '%s_%s_run??_raw.fif'
+                                             % (subject, config.study_name))))
```
agramfort: maybe this is dangerous as if you run twice the scripts, the second time you'll have more files that end with raw.fif

How about using modern python string formatting:

```
In [1]: subject = 'sub01'
In [2]: run = 'run1'
In [3]: '{subject}_{run}_raw.fif'.format(run=run, subject=subject)
Out[3]: 'sub01_run1_raw.fif'
```

basically the format `'{subject}_run{run}_raw.fif'` would be part of the config file
as well as the list of runs.

Something like this:


```
subject = 'sub01'

runs = ['run1', 'run2', 'run3']
base_raw_fname = '{subject}_{run}_raw.fif'

# then every script does:

for run in runs:
    raw_fname = base_raw_fname.format(**locals())
    print(raw_fname)
```

basically I find the use of glob too dangerous

SophieHerbst: I see. 
shall we make 
``
runs = ['run1', 'run2', 'run3']
``
a dict to allow for different number of runs per participant (real world users problems)

agramfort:
if it's just missing runs we could test that file is present and ignore it
with a warning.
I find it simpler than a dict per subject

>

SophieHerbst:
@agramfort: I implemented what we said yesterday: define number of expected runs in config, warn if less are found. 
Can you suggest how to format the warning more nicely? Right now it prints all of this: 
```

/Users/sophie/repos/mne-study-template/01-import_and_filter.py:73: UserWarning: Found only 2 runs warn('Found only ' + str(runs_found) + ' runs')
```

01-import_and_filter.py
```diff
@@ -18,20 +18,58 @@
...
+        try:
+            raw = mne.io.read_raw_fif(op.join(meg_subject_dir, raw_fname_in),
+                                      preload=True, verbose='error')
+        except FileNotFoundError:
+            if not raws:
+                raise ValueError('Cannot find ' + raw_fname_in)
+            if raws:
+                runs_found = len(raws)
+                warn('Found only ' + str(runs_found) + ' runs')
+            break
```
agramfort: I would raise a warning for each raw not found. Nothing fancy.

```
raw_fname_path = op.join(meg_subject_dir, raw_fname_in)
if not op.exists(raw_fname_path):
     warn('Run %s not found for subject %s ' % (raw_fname_in, subject))
raw = mne.io.read_raw_fif(raw_fname_path, preload=True, verbose='error')
```

this avoids a try which is better

agramfort: code above updated.

SophieHerbst: Ok, done. Still, I kept the error when no file can be found. 
`raise ValueError('Cannot find ' + raw_fname_path)`

The warning message still looks murky on the command line. I don't get why it prints the command again:
> /Users/sophie/repos/mne-study-template/01-import_and_filter.py:68: UserWarning: Run sample_audvis_run02_raw.fif not found for subject sample 
>   warn('Run %s not found for subject %s ' % (raw_fname_in, subject))

PR #18, opened 2019-02-09
Multiple runs

agramfort:


agramfort:
@SophieHerbst I closed #17 by mistake so here is the new PR with my last commit that aims to simplify stuff. Let me know what you think.

01-import_and_filter.py
```diff
@@ -28,45 +28,19 @@ def run_filter(subject):
...
-                raise ValueError('Cannot find ' + raw_fname_path)
-            else:
-                warn('Run %s not found for subject %s ' %
-                     (raw_fname_in, subject))
-                continue
+            warn('Run %s not found for subject %s ' %
+                 (raw_fname_in, subject))
+            continue
 
         raw = mne.io.read_raw_fif(raw_fname_path,
```
SophieHerbst: Won't this throw an error anyways if the path doesn't exist (so we don't need lines 70-72)?

SophieHerbst:
Looks good to me!

PR #19, opened 2019-02-10
Bads per run

SophieHerbst:
allow for definition of bad channels per run
if plot = True, plot single runs

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,74 +8,66 @@ file is meant to contain study specific parameters:
 
 - `plot` : if True, the single scripts generate plots
 
-DIRECTORIES
--------------------
+#### DIRECTORIES
 - `study_path` : path to the directory that contains your data.
 - `subjects_dir` : path pointing to the antomical files for all subjects
 - `meg_dir` : path pointing to the MEG files for all subjects
 
-SUBJECTS / RUNS
--------------------
+#### SUBJECTS / RUNS
 - `study_name` : the name of your experiment
 - `subjects` : a list of the subject names
 - `exclude_subjects` : the list of subjects to exclude from the above
 - `runs` : list of strings that specifies how your runs are named. For example `runs = ['run01', 'run02']`. If you have only one run it should be `runs = ['']`.
 - `base_raw_fname` : string that describes how your files are named. For example: `'{subject}_audvis_{run}_raw.fif'`
 
-BAD CHANNELS
--------------------
+#### BAD CHANNELS
 - `bads` : to be removed before maxfilter is applied (defined per run)
 
-DEFINE ADDITIONAL CHANNELS
--------------------
-
+#### DEFINE ADDITIONAL CHANNELS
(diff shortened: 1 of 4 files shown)


SophieHerbst:
Two questions: 

1. Should we rename the variable `run`? 
> `for run in config.runs:`

Executing `run` in debug mode breaks the script. 

2. At what stage to concatenate the runs and save one file? In `02-maxfilter` or in `03-extract_events`?

README.md
```diff
@@ -6,42 +6,69 @@ Study analysis template with MNE
...
+
+#### AUTOMATIC REJECTION OF ARTIFACTS
 - `reject` : the default rejection limits to make some epochs as bads. This allows to remove strong transient artifacts. **Note**: these numbers tend to vary between subjects.
+
+#### EPOCHING
 - `tmin`: float that gives the start time before event of an epoch.
 - `tmax` : float that gives the end time after event of an epochs.
 - `baseline` : tuple that specifies how to baseline the epochs. If None, then no baseline applied.
 - `event_id` : python dictionary that maps events (trigger/marker values) to conditions. E.g. `event_id = {'Auditory/Left': 1, 'Auditory/Right': 2}`
 - `runica` : boolean that says if ICA should be used or not.
```
agramfort: Remove as below

config.py
```diff
@@ -30,43 +30,52 @@
...
-# naming should be consistant over subjects
-# put the number of runs you ideally expect to have per subject
-# the scripts will issue a warning if there are less
-# leave empty if there is just one file
-runs =  [''] # ['run01', 'run02']
+# The MEG-data need to be stored in a folder
+# named my_study_path/MEG/my_subject/
 
+# This is the name of your experimnet
+study_name = 'audvis'
```
agramfort: Where do you use it?

agramfort:
>
> Two questions:
>
>    1. Should we rename the variable run?
>
> for run in config.runs:
>
>
run is the naming used by BIDS-MEG. See:

https://hal.archives-ouvertes.fr/hal-01591080/document

and look for what we mean by session and run.

> Executing run in debug mode breaks the script.
>
how does it break it?

>
>    1. At what stage to concatenate the runs and save one file? In
>    02-maxfilter or in 03-extract_events?
>
> I would not concatenate the raws but just at the epochs level. otherwise
it's going to require too much memory I fear.

config.py
SophieHerbst: `base_raw_fname = '{subject}_' + study_name + '{run}_raw.fif'`

SophieHerbst:
@agramfort, when I put a breakpoint in the function and then execute ` run`  (to check which run I am in) from the command line, it says: 

```

  File "/Users/sophie/anaconda/envs/mne/lib/python3.6/pdb.py", line 1028, in do_run
    raise Restart

Restart
```

agramfort:
from ipython the magic command should be

%run

not just

run

>

SophieHerbst:
Hm. Ok. Not sure how to fix this. 
But if it is not a general problem, let's keep run.

PR #20, opened 2019-02-12
review extract events

massich:


03-extract_events.py
```diff
@@ -38,8 +38,8 @@ def run_events(subject):
 
         if config.plot:
             # plot events
-            figure = mne.viz.plot_events(events)
-            figure.show()
+            fig = mne.viz.plot_events(events, show=True)
```
agramfort: you should be using:

`mne.viz.plot_events(events, sfreq=raw.info['sfreq'], first_samp=raw.first_samp)`

sappelhoff:
@massich this can be closed, because the changes have already been incorporated (except `figure.suptitle(run)`)

PR #21, opened 2019-02-12
reorder and review 

massich:
This PR moves `make_epochs.py` earlier in the pipeline and does another pass in the script.

@SophieHerbst feel free to take it from here, and ping me when its ready to review/merge


04-make_epochs.py
```diff
@@ -57,6 +55,8 @@ def run_epochs(subject):
     # Construct metadata from the epochs
     # Add here if you need to attach a pandas dataframe as metadata
     # to your epochs object.
+    
+    # XXX: nice example missing
```
SophieHerbst: @agramfort: I remember I used an example on how to implement metadata, but could not find it. Do you have a suggestion to what we could point?

agramfort: https://martinos.org/mne/dev/auto_tutorials/plot_metadata_epochs.html

PR #22, opened 2019-02-13
ica on epoched data for meg and eeg

SophieHerbst:


05a-run_ica.py
```diff
@@ -14,50 +14,56 @@
...
+                               eog=False, stim=False, exclude='bads')
+    all_picks = dict({'meg': picks_meg, 'eeg': picks_eeg})
+
+    for ch_type in ['meg', 'eeg']:
+        print('Running ICA for ' + ch_type)
+        
+        if ch_type == 'meg': 
+            
+            # XXX this does not work on epochs: 
+            # meg_maxfilter_rank = epochs_for_ICA.copy().pick_types(meg=True).estimate_rank()
```
SophieHerbst: @massich how to check the rank for epoched data?

05a-run_ica.py
```diff
@@ -14,50 +14,56 @@
...
     print("Processing subject: %s" % subject)
-    
+
     meg_subject_dir = op.join(config.meg_dir, subject)
-   
+
     epochs_fname = op.join(meg_subject_dir,
-                            config.base_epochs_fname.format(**locals()))
+                           config.base_epochs_fname.format(**locals()))
     epochs_for_ICA_fname = op.splitext(epochs_fname)[0] + '_for_ICA.fif'
```
SophieHerbst: @massich I know that the file should end with epo, but I don't know how to do that.

04-make_epochs.py
```diff
@@ -67,9 +67,24 @@ def run_epochs(subject):
...
-    epochs.save(op.join(meg_subject_dir, '%s-epo.fif' % subject))
-
+    print('  Writing epochs to disk')
+    epochs_fname = op.join(meg_subject_dir,
+                            config.base_epochs_fname.format(**locals()))
+
+    epochs.save(epochs_fname)
+    
+    # produce high-pass filtered version of the data for ICA    
+    epochs_for_ICA = mne.Epochs(raw.copy().filter(l_freq=1., h_freq=None),
```
agramfort: don't use capital letters in variable names

04-make_epochs.py
```diff
@@ -67,9 +67,24 @@ def run_epochs(subject):
...
+                                events, config.event_id, config.tmin, 
+                                config.tmax, proj=True, 
+                                picks=picks, baseline=config.baseline,
+                                preload=False, decim=config.decim,
+                                reject=config.reject)
+    
+    
+    epochs_for_ICA_fname = op.splitext(epochs_fname)[0] + '_for_ICA.fif'
```
agramfort: I think the ICA code should filter the data in memory and the otuput of the ICA script should only be a -ica.fif file. No need to duplicate epochs etc. on disk.

SophieHerbst: But since we are running the ICA on epoched data, there is no raw data to be filtered in memory. And the epochs are too short for 1Hz hp.

config.py
```diff
@@ -61,7 +61,11 @@
 
 # This generates the name for the raw files
 # Note that there is no underscore between study_name and run.
+
+# XXX this can likely be done more efficiently
 base_raw_fname = '{subject}_' + study_name + '{run}_raw.fif'
+base_epochs_fname = '{subject}_' + study_name + '_epo.fif'
+base_ave_fname = '{subject}_' + study_name + '_ave.fif'
```
agramfort: convention is to name it -epo.fif and -ave.fif (not with _)

05a-run_ica.py
```diff
@@ -14,51 +14,55 @@
...
+
     meg_subject_dir = op.join(config.meg_dir, subject)
-    raw_fnames = op.join(meg_subject_dir, '%s_audvis_filt_sss_raw.fif' % subject)
+
+    epochs_fname = op.join(meg_subject_dir,
+                           config.base_epochs_fname.format(**locals()))
+    
+    epochs_for_ica_fname = op.splitext(epochs_fname)[0] + '_for_ica.fif'
+
+    epochs_for_ica = mne.read_epochs(epochs_for_ica_fname, preload=True)
```
agramfort: I would recompute the epochs here and not rely on epochs that have been saved elsewhere.
It's fast to recompute and avoids storing big files just to get a tiny -ica.fif file.

05a-run_ica.py
```diff
@@ -14,51 +14,55 @@
...
-          % (subject, (' (tSSS=%d)' % tsss) if tsss else ''))
+    print("Processing subject: %s" % subject)
+
     meg_subject_dir = op.join(config.meg_dir, subject)
-    raw_fnames = op.join(meg_subject_dir, '%s_audvis_filt_sss_raw.fif' % subject)
+
+    epochs_fname = op.join(meg_subject_dir,
+                           config.base_epochs_fname.format(**locals()))
+    
+    epochs_for_ica_fname = op.splitext(epochs_fname)[0] + '_for_ica.fif'
```
agramfort: this should disappear but anyway file name should have ended -epo.fif

config.py
SophieHerbst: Ok. We should find a more consistent/ efficient way for the filenames anyways (see issue #16). 
Ideally, we just have one base-name and add the endings/ extensions in every script.

05a-run_ica.py
SophieHerbst: Ok, we can do that. I was thinking it is good to do it in one script to keep the epochs for ICA as similar to the 'real' ones as possible. But since all specifications come from config, this should still work.

PR #23, opened 2019-02-17
Make file naming more consistent over scripts

SophieHerbst:
define only the base name in config, add extensions in every file


agramfort:
looks good

PR #27, opened 2019-04-12
[WIP] contribute back upstream CBD changes

agramfort:


massich:
we should remove 8c982d2 and rebase (after merging #29) that should be green.

PR #31, opened 2019-05-28
update conda link, fix pip caching, always upgrade mne

sappelhoff:
- Try stretch and python 3.6.8 docker image
- update link to install Miniconda (continuum is now Anaconda, and they provide miniconda3 or miniconda2, which is more explicit than "miniconda")
- Fix pip-caching:
    - the pip cache was never restored, because circle only restores the first matching key (see: [here](https://circleci.com/docs/2.0/caching/#restoring-cache)), thus only the data-cache was being restored
    - with a second `restore_cache` call, this should be fixed
    - I also make the pip-cache contingent on the upstream MNE-Python `environment.yml` ... in case there are changes upstream, we would want to update our cache ...
    - lastly: This involved also adding a `pip install -U mne`, because otherwise our cache might make us stuck with an older version of MNE-Python

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,8 +2,7 @@ version: 2
 jobs:
     build_docs:
       docker:
-        # 3.6-jessie is too new for conda
-        - image: circleci/python:3.6-jessie
+        - image: circleci/python:3.6.8-stretch
       steps:
         - checkout
         - run:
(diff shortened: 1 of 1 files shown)


sappelhoff:
# question about the testing

Why do we only run `make test` and not `make all`? --> `make test` does not generate reports:

```Makefile
test: fetch_test sensor source

all: fetch sensor source report
```

... and without reports, our circle calls to upload artifacts are meaningless:

```yml
        # Save the outputs
        - store_artifacts:
            path: doc/_build/html/
            destination: dev
        - store_artifacts:
            path: doc/_build/html_stable/
            destination: stable


```

agramfort:
report was crashing on travis due to mayavi ... So I just skipped it temporarily

agramfort:
thx @sappelhoff

PR #60, opened 2019-11-14
Bdm pipeline

SophieHerbst:


SophieHerbst:
That was just a test we did with @agramfort 
Can be deleted.

PR #68, opened 2020-03-17
test report

agramfort:


agramfort:
we're getting there https://app.circleci.com/pipelines/github/mne-tools/mne-study-template/180/workflows/dc120f5d-95fc-4cec-b0eb-674b36dcbe2f/jobs/181/artifacts

agramfort:
let's move on from here. 3d does not work yet but I have hope

PR #69, opened 2020-04-19
WIP: NF: Allow auto-detection of flat & noisy channels

hoechenberger:
This happens via `mne.preprocessing.find_bad_channels_maxwell` and requires MNE-Python 0.20 or newer.

### DIFF
--- a/01-import_and_maxfilter.py
+++ b/01-import_and_maxfilter.py
@@ -161,7 +161,7 @@ def run_maxwell_filter(subject, session=None):
             bads = sorted(set(bads))
             raw.info['bads'] = bads
             print(f'Marked {len(raw.info["bads"])} channels as bad.')
-            del bads, auto_flat, auto_noisy, msg
+            del bads, auto_flat_chs, auto_noisy_chs, msg
 
         if config.use_maxwell_filter:
             print('Applying maxwell filter.')
--- a/01-import_and_maxfilter.py
+++ b/01-import_and_maxfilter.py
@@ -3,9 +3,9 @@
 01. Apply Maxwell filter
 ========================
 
-The data are imported from the BIDS folder. 
+The data are imported from the BIDS folder.
 
-If you chose to run Maxwell filter (config.use_maxwell_filter = True), 
+If you chose to run Maxwell filter (config.use_maxwell_filter = True),
 the data are Maxwell filtered using SSS or tSSS (if config.mf_st_duration
 is not None) and movement compensation.
 
@@ -35,6 +35,7 @@ import os.path as op
 import glob
 import itertools
 import mne
+from mne.preprocessing import find_bad_channels_maxwell
 from mne.parallel import parallel_func
 from mne_bids.read import reader as mne_bids_readers
 from mne_bids import make_bids_basename, read_raw_bids
@@ -126,11 +127,47 @@ def run_maxwell_filter(subject, session=None):
         if hasattr(raw, 'fix_mag_coil_types'):
             raw.fix_mag_coil_types()
 
+        if config.find_flat_channels or config.find_noisy_channels:
+            if (config.find_flat_channels and
+                    not config.find_noisy_channels):
+                msg = 'Finding flat channels.'
+            elif (config.find_noisy_channels and
+                  not config.find_flat_channels):
+                msg = 'Finding noisy channels using Maxwell filtering.'
+            else:
+                msg = ('Finding flat channels, and noisy channels using '
+                       'Maxwell filtering.')
+
+            print(msg)
+            raw_lp_filtered_for_maxwell = (raw.copy()
+                                           .filter(l_freq=None,
+                                                   h_freq=40,
+                                                   verbose=True))
+            auto_noisy_chs, auto_flat_chs = find_bad_channels_maxwell(
+                raw=raw_lp_filtered_for_maxwell,
+                calibration=config.mf_cal_fname,
+                cross_talk=config.mf_ctc_fname,
+                verbose=True)
+            del raw_lp_filtered_for_maxwell
+
+            bads = raw.info['bads'].copy()
+            if config.find_flat_channels:
+                print(f'Found {len(auto_flat_chs)} flat channels.')
+                bads.extend(auto_flat_chs)
+            if config.find_noisy_channels:
+                print(f'Found {len(auto_noisy_chs)} noisy channels.')
+                bads.extend(auto_noisy_chs)
+
+            bads = sorted(set(bads))
+            raw.info['bads'] = bads
+            print(f'Marked {len(raw.info["bads"])} channels as bad.')
+            del bads, msg
+
         if config.use_maxwell_filter:
             print('Applying maxwell filter.')
 
             # Warn if no bad channels are set before Maxfilter
-            if raw.info['bads'] is None or len(raw.info['bads']) == 0:
+            if not raw.info['bads']:
                 print('\n Warning: Found no bad channels. \n ')
 
             if run_idx == 0:
--- a/01-import_and_maxfilter.py
+++ b/01-import_and_maxfilter.py
@@ -127,12 +127,12 @@ def run_maxwell_filter(subject, session=None):
         if hasattr(raw, 'fix_mag_coil_types'):
             raw.fix_mag_coil_types()
 
-        if config.find_flat_channels or config.find_noisy_channels:
-            if (config.find_flat_channels and
-                    not config.find_noisy_channels):
+        if config.find_flat_channels_meg or config.find_noisy_channels_meg:
+            if (config.find_flat_channels_meg and
+                    not config.find_noisy_channels_meg):
                 msg = 'Finding flat channels.'
-            elif (config.find_noisy_channels and
-                  not config.find_flat_channels):
+            elif (config.find_noisy_channels_meg and
+                  not config.find_flat_channels_meg):
                 msg = 'Finding noisy channels using Maxwell filtering.'
             else:
                 msg = ('Finding flat channels, and noisy channels using '
@@ -151,17 +151,17 @@ def run_maxwell_filter(subject, session=None):
             del raw_lp_filtered_for_maxwell
 
             bads = raw.info['bads'].copy()
-            if config.find_flat_channels:
+            if config.find_flat_channels_meg:
                 print(f'Found {len(auto_flat_chs)} flat channels.')
                 bads.extend(auto_flat_chs)
-            if config.find_noisy_channels:
+            if config.find_noisy_channels_meg:
                 print(f'Found {len(auto_noisy_chs)} noisy channels.')
                 bads.extend(auto_noisy_chs)
 
             bads = sorted(set(bads))
             raw.info['bads'] = bads
             print(f'Marked {len(raw.info["bads"])} channels as bad.')
-            del bads, msg
+            del bads, auto_flat, auto_noisy, msg
 
         if config.use_maxwell_filter:
             print('Applying maxwell filter.')
--- a/config.py
+++ b/config.py
@@ -163,7 +163,7 @@ set_channel_types = None
 # ``mf_cal_fname`` above.
 #
 # "Bad", i.e. flat and overly noisy channels, can be automatically detected
-# using a procedure inspired by the commercial MaxFilter by Electa. First,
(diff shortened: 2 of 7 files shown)


agramfort:
Any idea why CIs fail ?

hoechenberger:
@agramfort Yes, will fix.

hoechenberger:
So now we're essentially not testing this bc per default it's all disabled… What shall I do? Add a config file to `tests/configs/` that _does_ test this functionality?

hoechenberger:
Also: Should the results of flat and noisy channel detection be added to the Report? Probably yes?

hoechenberger:
To implement a proper test, I will need a test dataset with from which I can acquire the MEG crosstalk and calibration files. If you have any suggestions which dataset may be appropriate, please to let me know.

agramfort:
you should have it for mne sample data (ds000248) in the derivative folder
although I don't think it should be there....

hoechenberger:
@agramfort Thx, simply using `None` for now :)

hoechenberger:
So now the only thing left is to add an entry to the `Report`, I suppose. Also wondering if I should crop the data to reduce testing time!!

agramfort:
fair enough for None

I would not crop files as I want to see the quality of the produced results with real quality data. 

This has not been carefully done so far

hoechenberger:
Ok, so please don't merge until I've added a `Report` entry.

hoechenberger:
Since the `Report` is only constructed at the very end of the processing pipeline, I need to somehow store the information about which channels were automatically detected as being `bad`. There's not designated data structure in MNE for this, right? Wonder if I should just add this info to the `.comment` field? Although that's going to be annoying, as MNE rewrites it on certain operations (like averaging `Epochs`). Any suggestions welcome.

SophieHerbst:
@hoechenberger [This](https://github.com/brainthemind/CogBrainDyn_MEG_Pipeline/blob/master/01-import_and_maxwell_filter.py) is how we added bad channels to info['bads‘] (manually from annotations / config) before turning everything into bids. 


 ```
# add bad channels, using the annotations if found
        annotations_fname_out = op.join(meg_subject_dir,
                                        config.base_fname.format(**locals()))
        pre, ext = op.splitext(annotations_fname_out)
        annotations_fname_in = pre + '.csv'

        if op.exists(annotations_fname_in):
            annotations = mne.read.annotations(annotations_fname_in)
            raw.set_annotations(annotations)
            print("Reading bads from annotations")
            print("bads: ", raw.info['bads'])
        else:
            # read bad channels for run from config
            if run:
                bads = config.bads[subject][run]
            else:
                bads = config.bads[subject]
                raw.info['bads'] = bads
            print("no annotations found, reading bads from config")
            print("bads: ", raw.info['bads'])
```

**Comment by hoechenberger** [comment]
Thanks @SophieHerbst!

>            annotations = mne.read.annotations(annotations_fname_in)
>            raw.set_annotations(annotations)

Does that actually do what you thought it would? :) Because `set_annotations()`, to my knowledge, wouldn't touch `info['bads']` – `Annotations` are used for marking bad "time periods" in your data, while `info['bads']` is uses for "spatial" selection of bad data (i.e., corrupt sensors). I would expect `set_annotations()` not to add any channels to `info['bads']`

**Comment by agramfort** [comment]
bads should be written in the bids electrodes.tsv file or detected
automatically by the pipeline.
I am not sure the config.py is the right place to make the bads

>

**Comment by hoechenberger** [comment]
@agramfort @SophieHerbst said this is how they did it in the pre-BIDS era :)

**Comment by SophieHerbst** [comment]
Saw that too, @hoechenberger we probably did not test the annotations part because the interactive figure from which we wanted to get the annotations does not work (aka 'the eternal bug'). We add the bads from config.

**Comment by SophieHerbst** [comment]
But whatever you do for bids, it will be important to allow the user to vizualize the data offline. We do note bads during recording, but crucial to double-check.

**Comment by hoechenberger** [comment]
> But whatever you do for bids, it will be important to allow the user to visualize the data offline. We do note bads during recording, but crucial to double-check.

Ok, but just so we agree upon the use case here:
this pipeline is for data _that is already in BIDS format_. All channels marked as bad "in BIDS", e.g. directly after recording and before converting to BIDS, will be treated as "bad" by the pipeline too. My PR now allows users to additionally run an "automated bad-channel detection", which will potentially discover additional "bad" channels. I had been planning to add the latter to the report.

If you say you want _all_ bad channels in the report, then this would make things even easier for me, actually. Because I wouldn't necessarily have to state, "ok so these channels were marked as bad before running the pipeline, and the pipeline additionally discovered the following flat channels: ... and the following noisy channels: ...", but instead one could simply state, "The following channels were marked as bad: ..." and simply list `info['bads']`… might even be the case already, need to check!

**Comment by hoechenberger** [comment]
> one could simply state, "The following channels were marked as bad: ..." and simply list `info['bads']`… might even be the case already, need to check!

This information already appears in the report! What one potentially could add is a diff between pre- and post-pipeline `info['bads']`!

**Comment by hoechenberger** [comment]
> But whatever you do for bids, it will be important to allow the user to visualize the data offline.

After a second read, I'm not sure what exactly you mean here. Are you talking about interactive viz while the pipeline is running? Specifically, after the automated bad channel detection has done its job, so one could intervene and say, "Nope, you've got it all wrong, bro! Now look at me go!"

**Comment by SophieHerbst** [comment]
from my user perspective (small-n studies, individual treatment of data), this is what it would like to get from a pipeline: 

1. manually note bad sensors somewhere my code (config / subject file)
2. start my pipeline with visualizing data to verify bad sensors (before: manually, now: your bad channel detector + viz for double checking) -> add sensors to the code above
3. run Maxfilter (bad MEG sensors are interpolated and do not appear as bads anymore)
4. keep notes of excluded / interpolated sensors for the final paper

1) and 2) could potentially be done together. 
I don't know at what stage you would like to convert to BIDS here, but my main point was that the user should be enabled / encouraged to double-check the bad sensors by visualizing the data. 
If you miss 1 bad sensor (for instance the spiky one we exchanged about) in 1 subject, it can mess up a group average (n ~ 20), especially in time-frequency data.

**Comment by hoechenberger** [comment]
Thanks for the clarification, @SophieHerbst 

@agramfort This is probably something we should handle in `mne-bids` instead, then?

**Comment by hoechenberger** [comment]
@SophieHerbst Is a "bad" channel _always_ bad? Or does the classification change with your use case and preprocessing / filtering etc?

File: 01-import_and_maxfilter.py [hunk 1]
```diff
@@ -126,11 +127,47 @@ def run_maxwell_filter(subject, session=None):
...
+                    not config.find_noisy_channels_meg):
+                msg = 'Finding flat channels.'
+            elif (config.find_noisy_channels_meg and
+                  not config.find_flat_channels_meg):
+                msg = 'Finding noisy channels using Maxwell filtering.'
+            else:
+                msg = ('Finding flat channels, and noisy channels using '
+                       'Maxwell filtering.')
+
+            print(msg)
```
agramfort, line 141: we should consider using a proper logging system (later...)

01-import_and_maxfilter.py
```diff
@@ -126,11 +127,47 @@ def run_maxwell_filter(subject, session=None):
...
+                                                   h_freq=40,
+                                                   verbose=True))
+            auto_noisy_chs, auto_flat_chs = find_bad_channels_maxwell(
+                raw=raw_lp_filtered_for_maxwell,
+                calibration=config.mf_cal_fname,
+                cross_talk=config.mf_ctc_fname,
+                verbose=True)
+            del raw_lp_filtered_for_maxwell
+
+            bads = raw.info['bads'].copy()
```
agramfort, line 153: raw.info['bads'] can typically be a list. I would do:
```suggestion
            bads = list(raw.info['bads'])
```

config.py
```diff
@@ -155,18 +155,42 @@
 set_channel_types = None
 
 ###############################################################################
-# MAXFILTER PARAMETERS
-# --------------------
+# MAXFILTER AND MAXWELL FILTER PARAMETERS
```
agramfort, line 158: what's the difference? MAXFILTER is more an abbrev of MAXWELL FILTER

config.py
```diff
@@ -155,18 +155,42 @@
...
+# ---------------------------------------
 # done in 01-import_and_maxfilter.py
 #
+# Note: For any of this to work, you must set ``mf_ctc_fname`` and
+# ``mf_cal_fname`` above.
+#
+# "Bad", i.e. flat and overly noisy channels, can be automatically detected
+# using a procedure inspired by the commercial MaxFilter by Electa. First,
```
agramfort, line 166:
```suggestion
# using a procedure inspired by the commercial MaxFilter by Elekta. First,
```

tests/download_test_data.py
```diff
@@ -17,6 +17,7 @@ def _provide_testing_data(dataset=None):
         'somato': ('https://github.com/sappelhoff/MNE-somato-data'),
         'ds000246': 'https://github.com/OpenNeuroDatasets/ds000246',
         'ds000248': 'https://github.com/OpenNeuroDatasets/ds000248',
+        'ds000248_autobads': 'https://github.com/OpenNeuroDatasets/ds000248',
```
agramfort, line 20: I would not create a new dataset. It will blow up the testing time if we start doing this

01-import_and_maxfilter.py
hoechenberger: Absolutely!

01-import_and_maxfilter.py
hoechenberger: How's that better? I find `list.copy()` more readable

config.py
hoechenberger: To me, MaxFilter to me is the Elekta thing, Maxwell filter is a more generic term. Is that not the case?

tests/download_test_data.py
hoechenberger: Ok, will try to solve it by re-using the `ds000248` dataset. Creating a new dataset was just the quickest solution I could come up with. But I totally see your point.

config.py
hoechenberger: @SophieHerbst I would be interested in your opinion here too :)

01-import_and_maxfilter.py
hoechenberger: Opened an issue: #70

01-import_and_maxfilter.py
agramfort: I did not know you could do a .copy() on a list :)

config.py
agramfort: but we don't have maxfilter support here. All we do is what mne calls maxwell_filter

hoechenberger: We do this `find_bad_channels_maxwell` which is supposed to be _kind of_ like `MaxFilter`, but yes maybe I should simply change this to read "MAXWELL FILTER PARAMETERS"

agramfort:
+1 for MAXWELL FILTER PARAMETERS

>

config.py
hoechenberger: solved in 2c06272

01-import_and_maxfilter.py
hoechenberger: It's been possible since Python 3.3 or so :)

tests/download_test_data.py
hoechenberger: Resolved in 92b0090

hoechenberger:
@SophieHerbst Have been trying to triage things a bit with @agramfort. Current idea is

- to keep the auto-bad detection via `find_bad_channels_maxwell()` in this pipeline, but turned off by default
- keeping this pipeline as automated as possible, i.e. ideally it should run unsupervised
- bad channels should be marked as such in the BIDS dataset before it gets thrown into this pipeline. I will provide a simple tool that allows users to run bad channel detection, data visualization, and manual bad channel (un)selection (i.e., will just call `Raw.plot()` – which consequently will also allow users to select _time periods_ as bad (i.e., add `Annotations`), and write this information back to the BIDS data. 

Priority, at the time of writing, is getting this pipeline ready and tested first, also to finalize our ideas of an appropriate workflow, and only then start working on the last point. WDYT? 

[btw until this feature / tool has been implemented, one can easily –– albeit a little inconveniently — edit the BIDS `channels.tsv` file to select channels as bad. But I know that this is really not a great solution without any viz.)

agramfort:
Thx @hoechenberger

PR #71, opened 2020-04-19
BF: Fix epochs.plot_image call

hoechenberger:
The `group_by` parameter used here is intended for sth else (and doesn't work like this); I believe  `picks` is what we want instead!


hoechenberger:
Can this be merged?

PR #73, opened 2020-04-19
BF: Extract the data rank from the info structure when estimating noise covariance

hoechenberger:
<strike>- Shuffle the data before the cross-validation split. This is required to get any meaningful CV results.</strike>

- Extract the data rank from the `info` structure. At least on my  test data this was necessary to get correct results.

<strike>- Logging and documentation improvements</strike>

### DIFF
--- a/12-make_cov.py
+++ b/12-make_cov.py
@@ -57,7 +57,10 @@ def run_covariance(subject, session=None):
     epochs = mne.read_epochs(fname_epo, preload=True)
 
     print('  Computing regularized covariance')
-    cv = KFold(3, random_state=config.random_state, shuffle=True)
+    # Do not shuffle the data before splitting into train and test samples.
+    # Perform a block cross-validation instead to maintain autocorrelated
+    # noise.
+    cv = KFold(3, shuffle=False)
     cov = mne.compute_covariance(epochs, tmax=0, method='shrunk', cv=cv,
                                  rank='info')
     cov.save(fname_cov)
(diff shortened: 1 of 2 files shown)


12-make_cov.py
```diff
@@ -57,8 +58,21 @@ def run_covariance(subject, session=None):
...
+               f'the user-defined value: {config.random_state}')
+    else:
+        msg = 'Using random number generator with default settings'
+
+    print(msg)
+    del msg
+
+    cv = KFold(3, random_state=config.random_state, shuffle=True)
+    cov = mne.compute_covariance(epochs, tmax=0, method='shrunk', cv=cv,
+                                 rank='info')
```
agramfort, line 75: rank='info' is bug fix but the rest is just doc right?

hoechenberger: and `shuffle=True` in the `KFold` instantiation.

FYI this PR contains two commits, one for the bugfix and one for the docs & logging:
bugfix: 2069391
doc: e928ea9

agramfort: arff I the present case (with temporally correlated data) I wouldn't not shuffle. You don't want in the test set samples that are "just" like train because of autocorrelated noise. Please keep shuffle=False and make a note. This technique is also known as "block cross-validation" in stats.

hoechenberger: Wow I didn't know that! Thanks for the clarification, I'm glad you caught this mistake!

I've reverted those commits (including the documentation & logging updates, which are unnecessary now). I've added `shuffle=False` (which is the default in `sklearn`, but I think being explicit here is a good thing), and added a comment trying to explain _why_ we don't want to shuffle the data.

 Can you please review again?

agramfort:
thx @hoechenberger !

PR #75, opened 2020-04-27
MRG: Allow specification of root path in config.py

hoechenberger:
The BIDS root directory could previously only be specified using the environment variable `BIDS_ROOT`. However, users told me this is cumbersome and doesn't really align with what they're used to in their daily work. This commit makes it obvious how to specify a BIDS root path
directly in `config.py`. By default, we will continue to use `BIDS_ROOT` environment variable, but it should now be easy for users to change this behavior.

Also raises a `ValueError` now instead of a `RuntimeError` if no BIDS root has been specified, I think this is more appropriate here.

Let me know what you think – haven't added tests yet because I first wanted to hear your thoughts on this one.

### DIFF
--- a/config.py
+++ b/config.py
@@ -16,11 +16,21 @@ PIPELINE_NAME = 'mne-study-template'
 VERSION = '0.1.dev0'
 CODE_URL = 'https://github.com/mne-tools/mne-study-template'
 
-# Get the bids_root from an environment variable, raise an error if not found
-bids_root = os.getenv('BIDS_ROOT', False)
+# Speficy the BIDS root directory. Pass an empty string or `None` to use the
+# value specified in the ``BIDS_ROOT`` environment variable instead. Raises
+# an exception if the BIDS root has not been specified.
+#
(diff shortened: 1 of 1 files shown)


config.py
```diff
@@ -16,11 +16,21 @@
 VERSION = '0.1.dev0'
 CODE_URL = 'https://github.com/mne-tools/mne-study-template'
 
-# Get the bids_root from an environment variable, raise an error if not found
-bids_root = os.getenv('BIDS_ROOT', False)
+# Speficy the BIDS root directory. Pass an empty string or `None` to use the
```
agramfort, line 19:
```suggestion
# Specify the BIDS root directory. Pass an empty string or `None` to use the
```

agramfort:
current mne-bids code with deal of sex/gender broke the study template

mne-bids needs to be fixed or dataset

hoechenberger:
Ok was already wondering what that CI issue's about.

hoechenberger:
@agramfort I've opened a bug upstream
https://github.com/mne-tools/mne-bids/issues/396

PR #78, opened 2020-04-27
MRG: BF: Sanitize paths in group-average analysis

hoechenberger:
This analysis step would fail for me if the condition names contains a forward slash. Need to sanitize the generated filenames.


hoechenberger:
Oh damn, I just realized this hasn't been reviewed / approved yet. If you're not happy, I can revert, @agramfort @jasmainak

agramfort:
yes all good !

PR #79, opened 2020-04-27
WIP: BF: Do not try to render alignment, STC if we have no valid 3d backend

hoechenberger:
Do not render alignment or STC when no valid MNE 3d backend has been set.

### DIFF
--- a/99-make_reports.py
+++ b/99-make_reports.py
@@ -66,7 +66,8 @@ def run_report(subject, session=None):
 
     rep.add_figs_to_section(figs, captions)
 
-    if op.exists(fname_trans):
+    if (op.exists(fname_trans) and
+            mne.viz.backends.renderer.backend is not None):
         fig = mne.viz.plot_alignment(evoked.info, fname_trans,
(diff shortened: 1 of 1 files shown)


99-make_reports.py
```diff
@@ -66,7 +66,8 @@ def run_report(subject, session=None):
 
     rep.add_figs_to_section(figs, captions)
 
-    if op.exists(fname_trans):
+    if (op.exists(fname_trans) and
+            mne.viz.backends.renderer.backend is not None):
```
agramfort, line 70: we don't hvae a get_backend method?

hoechenberger: We do but it will raise an exception if it fails to detect a valid backend. Maybe this should be changed:

https://github.com/mne-tools/mne-python/blob/master/mne/viz/backends/renderer.py#L140-L141

hoechenberger: Opened an issue upstream:
https://github.com/mne-tools/mne-python/issues/7683

PR #93, opened 2020-05-03
Remove 03-make_events.py

### LINKED ISSUE #91: Should 03-extract_events.py be dropped?
It seems this file is a leftover from the time before we had `Annotations` (or before `mne-bids` started storing events as `Annotations`).

For example, the docstring says:
https://github.com/mne-tools/mne-study-template/blob/280bde6e98d78c7b888e878e292d30cf867e20f2/03-extract_events.py#L6-L7

Now, we don't even have or use `config.stim_channel` anymore. And the events file created by this script is only ever mentioned again in `05a-run_ica.py`, albeit not _used_.

The produced event file is picked up by the `Report.parse_folder` call in `99-make_reports.py`, which creates an event plot; however, these events do not necessarily correspond to `config.event_id` in the sense that
- _all_ events are included here, not just the ones specified in `config.event_id`, and
- the events are  not labeled (i.e., just have an event number and not a "human-readable" name)

#### Proposal

We should remove `03-extract_events.py` altogether and instead change `99-make_reports.py` such that it reads events and event labels from annotations, and passes these values to `mne.viz.plot_events()` so we get properly labeled event plots of only the events the user deemed relevant (via `config.event_id`).

hoechenberger:
Closes #91


config.py
```diff
@@ -490,7 +490,7 @@
 # >>> min_event_duration = 0.002  # 2 miliseconds
 
 # XXX not needed if bids events are present
-# min_event_duration = 0.002
+min_event_duration = 0.002
```
agramfort, line 493: revert?

hoechenberger: Thought I could sneak in this small little default change here ;) But yes can revert if you want

agramfort:
why is this needed? does it fix some issue or pb in one of our tested
dataset?

>

hoechenberger:
> why is this needed? does it fix some issue or pb in one of our tested dataset?

I have to set this for all NeuroMag datasts I've worked with so far (Cam-CAN, and the Localizer dataset). I also don't understand why this is necessary really, but didn't have time to investigate.

hoechenberger:
Ok I know why CI is failing. We need to discuss: for the events plot, do we want one figure per run, or should I concat all events from all runs and plot them into a single figure?

SophieHerbst:
> > why is this needed? does it fix some issue or pb in one of our tested dataset?
> 
> I have to set this for all NeuroMag datasts I've worked with so far (Cam-CAN, and the Localizer dataset). I also don't understand why this is necessary really, but didn't have time to investigate.

@hoechenberger In the localizer data there are some 'ghost' triggers that usually last 1 sample (1ms). By setting the minimal event duration to be longer than that, we are able to get only the 'real trigger events', because the trigger function we uses sends triggers that last 4-8ms.

agramfort:
I would write one event file per run and store it to disk in the epochs
script. Basically when you use/need them.

>

99-make_reports.py
```diff
@@ -53,8 +55,17 @@ def run_report(subject, session=None):
                      subjects_dir=subjects_dir)
     rep.parse_folder(fpath_deriv, verbose=True)
 
-    evokeds = mne.read_evokeds(fname_ave)
+    # Visualize events.
+    raw_filt = mne.io.read_raw_fif(fname=fname_raw_filt,
+                                   allow_maxshield=config.allow_maxshield)
+    events = mne.find_events(raw=raw_filt,
+                             min_duration=config.min_event_duration)
```
agramfort, line 62: don't use find_events but use the events from bids file.

config.py
agramfort: this is not the default in mne-python and I am sure this will break other datasets.

99-make_reports.py
hoechenberger: That's in fact another issue I had been planning to bring up in the future, but since you're mentioning it now…

Currently the pipeline writes all derivative data via vanilla MNE functions, and doesn't use `mne-bids`. This should be changed, IMHO – and you seem to agree. I can amend this PR to do this (for all pipeline results), or open a new PR for this – what do you think?

agramfort:
let's talk in the afternoon. I call you asap.

>

99-make_reports.py
hoechenberger: (mind you, the events I process here are those from the filtered raw data, i.e., after the 2nd step of the pipeline has been run)

hoechenberger:
> let's talk in the afternoon. I call you asap.

But but but… it's a Sunday! Maybe I should join the union after all…

No but seriously, drop me a message on Slack and we can arrange something :)
ttyl!!

config.py
agramfort: please revert

02-frequency_filter.py
```diff
@@ -92,13 +92,54 @@ def run_filter(subject, run=None, session=None):
...
 def main():
     """Run filter."""
     parallel, run_func, _ = parallel_func(run_filter, n_jobs=config.N_JOBS)
     parallel(run_func(subject, run, session) for subject, run, session in
              itertools.product(config.get_subjects(), config.get_runs(),
                                config.get_sessions()))
 
+    # Do not parallelize concatenation to avoid memory issues.
+    for subject, session in zip(config.get_subjects(), config.get_sessions()):
+        concatenate_filtered_raws(subject=subject, session=session)
```
agramfort, line 141: saving GBs of data to keep the annotations seems really an overkill. It you go this on cam-can you will blow up the disk space.

hoechenberger: > saving GBs of data to keep the annotations seems really an overkill. It you go this on cam-can you will blow up the disk space.

I just checked an there's an `Annotations.save()`, which I can certainly use instead.

agramfort:
+1

>

hoechenberger:
@agramfort There is no good way to work with `Annotations` directly (e.g., you cannot pass them directly + an `info` struct` to `events_from_annotatios()`, but instead you have to attach them to a `Raw` first), I decided _not_ to save the Annotations to disk. Instead, I moved the "filtered raw concatenation" to the reports script, concatenate in-place, and extract the events there for plotting. Felt the cleanest.

config.py
```diff
@@ -445,48 +445,6 @@
...
-
-# ``min_event_duration`` : float
-#    The minimal duration of the events you want to extract (in seconds).
-#
-# Example
-# ~~~~~~~
-# >>> min_event_duration = 0.002  # 2 miliseconds
-
-# XXX not needed if bids events are present
-# min_event_duration = 0.002
```
agramfort, line 466: why removing this one?

config.py
```diff
@@ -445,48 +445,6 @@
...
 
-# ``stim_channel`` : str
-#    The name of the stimulus channel, which contains the events.
-#
-# Example
-# ~~~~~~~
-# >>> stim_channel = 'STI 014'  # or 'STI101'
-
-# XXX not needed if bids events are present
-# stim_channel = 'STI 014'
```
agramfort, line 456: ok to remove this one

config.py
```diff
@@ -445,48 +445,6 @@
...
-#
-# Allow for grouping of all ``auditory`` conditions, regardless of
-# stimulation side:
-# >>> event_id = {'auditory/left': 1, 'auditory/right': 2}`
-#
-# Allow for grouping into ``auditory``, ``visual``, ``left``, and ``right``:
-# >>> event_id = {'auditory/left': 1, 'auditory/right': 2,
-#                 'visual/left': 3, 'visual/right': 4}`
-
-# event_id = {'auditory/left': 1, 'auditory/right': 2}
```
agramfort, line 488: how do you specify your contrast now? why remving this?

agramfort:
we should have mne.viz.plot_annotations() and annotations.plot()
functions/methods.

ok?

>

hoechenberger:
@agramfort 
Oh, sorry, forgot to comment on the removal of `min_event_duration`, `stim_channel`, and `event_id`. Good thing you're asking.

It's actually very simple: If we only work with BIDS data – which is my premise here – we don't need to handle event extraction anymore, except via `events_from_annotations()` (bc `mne-bids` writes events as annotations).

- `min_event_duration` is a parameter of `mne.find_events()`. Since we don't use `find_events()` anymore in the pipeline, this parameter can go.
- `stim_channel` is obvious
- `event_id` can also go bc during the conversion to BIDS, one can (should) specify the `event_id`, which are then stored as the name of the respective annotations. So when you want to have your `events` and `event_id` back from an `mne-bids` dataset, you simply do `events, event_id = mne.events_from_annotations(raw)`. No need to require a separate `event_id` in the config file.

hoechenberger:
> we should have mne.viz.plot_annotations() and annotations.plot() functions/methods. ok?

I agree. I can open an issue upstream.

agramfort:
good points

what bothers me about dropping event_id is that the convention with / is
really an MNE thing
and we should be able to process data converted with other tools like
fieldtrip.

you have an event_id parameter in events_from_annotations. It's "auto" by
default so we can
adopt this standard and set event_id="auto" in config and pass
config.event_id in events_from_annotations.

ok?

hoechenberger:
> > we should have mne.viz.plot_annotations() and annotations.plot() functions/methods. ok?
> 
> I agree. I can open an issue upstream.

https://github.com/mne-tools/mne-python/issues/7720

hoechenberger:
> you have an event_id parameter in events_from_annotations. It's "auto" by
> default so we can
> adopt this standard and set event_id="auto" in config and pass
> config.event_id in events_from_annotations.
> 
> ok?

Wonder if we shouldn't offer a direct way to rename events instead? So instead of asking the user to assign new names based on event codes, we could offer to allow them to map old names to new names, e.g., 

```
auditory_left -> auditory/left
auditory_right -> auditory/right
```
Thoughts?

hoechenberger:
Also… hang on a second. Currently we're relying on the presence of `Annotations` in many places of the pipeline. We wouldn't be able to support FieldTrip etc easily anyway without some significant modifications, no?

agramfort:
see API of event_id in events_from_annotations
if you name it auditory_left than no way you can do auditory for cond in mne

hoechenberger:
I guess what I'm saying is this:

No matter where or how the data was processed originally: if it's BIDS, we will want to read it via `mne-bids`, right?

Now `mne-bids` parses `events.tsv` and adds `Annotations` to the data accordingly. In this process, we lose the original event IDs. But we don't need them anyway since we have our fine annotations now.

To rename an experimental event, you would do sth like
```python
[a.description = 'new' for a in raw.annotations if a.description == 'old']
```
and if you need `events`, you use our friend `events_from_annotations()`.

I therefore see  no need for `config.event_id`… Also I should mention that even before this very PR, `config.event_id` wasn't used anywhere in the pipeline anymore anyway, since everything had been switched over to using `raw.annotations`.

hoechenberger:
Note to self after our brief phone call:
- create a dict mapping `old event name in BIDS dataset -> new event name for pipeline`
- do the mapping early on (read raw, change annotation description)
- profit!

hoechenberger:
@agramfort I think I've implemented everything we've discussed 👍

hoechenberger:
Also I've started to use f-strings in the last commit. Makes certain string formatting operations so much more readable…

agramfort:
thx @hoechenberger 

yet looking at some reports in artifacts I see rather weird stuff eg

https://294-167957354-gh.circle-artifacts.com/0/reports/ds001810/report_average_task-attentionalblink.html
https://294-167957354-gh.circle-artifacts.com/0/reports/ds000246/report_task-AEF.html
https://294-167957354-gh.circle-artifacts.com/0/reports/ds001810/report_task-attentionalblink.html

hoechenberger:
@agramfort Now this is weird, all these reports contain `-eve.fif` files, which are not supposed to be there. Probably an issue related to caching??? We may need to clean the `derivatives` folders before / after tests.

Regarding the artifacts themselves, yes, looks like we'll have some more work to do there :(

agramfort:
derivative folders should not be cached. Maybe delete the folders before
stroing cache

yes nobody really checked quality of results beyond sample and maybe somato

>

PR #95, opened 2020-05-04
TESTS: Retrieve somato dataset from OSF

hoechenberger:
This is MUCH faster than downloading it from GitHub and should, therefore, speed up CI testing quite a bit.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -66,6 +66,11 @@ jobs:
               git config --global user.email "user@mne.com"
               git config --global user.name "mne user"
 
+        - run:
+            name: Install Python `wget` to retrieve somato dataset from OSF
+            command: |
+              pip install wget
+
         # Save Python cache
         - save_cache:
             key: pip-cache-0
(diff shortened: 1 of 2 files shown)


agramfort:
data should be cached by circleci anyway it's a price to pay only once

hoechenberger:
> data should be cached by circleci anyway it's a price to pay only once

But I can see how it's taking a long time to download almost every single time we run those tests… and when I run the test data download locally or on `drago4`, it starts out fast but then after a while starts crawling super slowly, I think this might be an issue with `git-annex` on GitHub with many files small files or something

agramfort:
circle caching does not work?

>

hoechenberger:
> data should be cached by circleci anyway it's a price to pay only once

Caching seems to only happen within any given Circle run, not across subsequent runs. Since we're only using each dataset once during each CI run, we don't profit from caching

hoechenberger:
> circle caching does not work?

Not as far as I can tell, it's downloading all data in each new run

agramfort:
then it's the first thing to do. Enable caching in circleci like we do in
MNE.

hoechenberger:
> then it's the first thing to do. Enable caching in circleci like we do in
> MNE.

Ok will look into this. But this would still not help for local testing :)

PR #100, opened 2020-05-05
TESTS: Clear derivatives dir before and after running tests

hoechenberger:


tests/run_tests.py
```diff
@@ -98,16 +106,22 @@ def run_tests(test_suite):
...
+        clean_derivatives(os.environ['BIDS_ROOT'])
 
         # run the pipelines
         for pipeline in test_tuple[1::]:
             pipeline()
 
+        # Delete derivatives after testing is finished. Skip this step on
+        # circle, as we intend to retain the generated reports.
+        if os.environ['CI'] != 'true':
+            clean_derivatives(os.environ['BIDS_ROOT'])
```
agramfort, line 121: I think your changes to cricle.yml are enough

hoechenberger: I thought this here might be nice to have for running tests locally. Thoughts?

agramfort:
for me I would add a "make clean" target in the makefile

>

hoechenberger:
@agramfort Ok I've reverted the change to `run_tests.py`. Regarding the Makefile suggestion, I would look into this some other time.

PR #103, opened 2020-05-05
BF: Set rank='info' during make_inverse_operator()

hoechenberger:
We already use this during noise cov calculation, so we should use the very same setting here.


agramfort:
thx @hoechenberger

PR #106, opened 2020-05-05
ENH: Allow for ICA algorithm selection

### LINKED ISSUE #94: More ICA iterations
The low default number of MNE's ICA iterations should be increased, at least for EEG, since in my experience MNE's default is often not sufficient for ICA to converge; and / or the algorithm should be switched to Picard.

hoechenberger:
Closes #94

### DIFF
--- a/04a-run_ica.py
+++ b/04a-run_ica.py
@@ -110,8 +110,17 @@ def run_ica(subject, session=None):
     msg = f'Running ICA for {kind}'
     logger.info(gen_log_message(message=msg, step=4, subject=subject,
                                 session=session))
-    ica = ICA(method='fastica', random_state=config.random_state,
-              n_components=n_components[kind])
+
+    if config.ica_algorithm == 'picard':
+        fit_params = dict(fastica_it=5)
+    elif config.ica_algorithm == 'extended_infomax':
+        fit_params = dict(extended=True)
+    else:
+        fit_params = None
+
+    ica = ICA(method=config.ica_algorithm, random_state=config.random_state,
+              n_components=n_components[kind], fit_params=fit_params,
+              max_iter=config.ica_max_iterations)
 
     picks = all_picks[kind]
     if picks.size == 0:
--- a/04a-run_ica.py
+++ b/04a-run_ica.py
@@ -115,7 +115,7 @@ def run_ica(subject, session=None):
         fit_params = dict(fastica_it=5)
     elif config.ica_algorithm == 'extended_infomax':
         fit_params = dict(extended=True)
-    else:
+    elif config.ica_algorithm == 'fastica':
         fit_params = None
 
     ica = ICA(method=config.ica_algorithm, random_state=config.random_state,
--- a/config.py
+++ b/config.py
@@ -525,7 +525,7 @@ ica_algorithm = 'picard'
 #   Maximum number of iterations to decompose the data into independent
 #   components. A low number means to finish earlier, but the consequence is that the
 #   algorithm may not have finished converging. To ensure
-#   convergence, pick a high number here (e.g. 3000); the algorithm will
+#   convergence, pick a high number here (e.g. 3000); yet the algorithm will
 #   terminate as soon as it determines that is has successfully converged, and
 #   not necessarily exhaust the maximum number of iterations. Note that the
 #   default of 200 seems to be sufficient for Picard in many datasets, because
--- a/config.py
+++ b/config.py
@@ -523,7 +523,7 @@ ica_algorithm = 'picard'
 
 # ``ica_max_iterations`` : int
 #   Maximum number of iterations to decompose the data into independent
-#   components. A low number means to finish earlier, but a the cost that the
+#   components. A low number means to finish earlier, but the consequence is that the
 #   algorithm may not always have finished converging. That means to ensure
 #   convergence, pick a high number here (e.g. 3000); the algorithm will
 #   terminate as soon as it determines that is has successfully converged, and
--- a/config.py
+++ b/config.py
@@ -805,8 +805,9 @@ if use_ssp and use_ica:
     raise ValueError('Cannot use both SSP and ICA.')
 
 if use_ica and ica_algorithm not in ('picard', 'fastica', 'extended_infomax'):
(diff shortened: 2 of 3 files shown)


04a-run_ica.py
```diff
@@ -105,8 +105,8 @@ def run_ica(subject, session=None):
 
     kind = config.get_kind()
     print('Running ICA for ' + kind)
-    ica = ICA(method='fastica', random_state=config.random_state,
-              n_components=n_components[kind])
+    ica = ICA(method='picard', random_state=config.random_state,
```
agramfort, line 108: I would actually expose this as a parameter in config as not everyone (sadly) will expect picard to be run by default

hoechenberger: > not everyone (sadly)

hehe. Hey you need to promote this thing more… and it's not even the default in MNE!!

hoechenberger: Also I'm taking suggestions here: which parameters should picard receive that are not the defaults? for example here I'm passing `fastica_it=5`, not sure if that is really useful… but maybe you have some ideas about what could make sense?

agramfort:
It's better if other people than me say it's faster. It will feel less
biased :)

>

hoechenberger:
@agramfort Currently we only pass those PCA components that explain 99.9% of the variance to ICA; I'm not sure if that's common practice in MEG? Never seen anything like that with EEG tbh…

hoechenberger:
@agramfort this is what I'm talking about:
https://github.com/mne-tools/mne-study-template/blob/6b741cc33b63ccfa83f2be36468e4bb845341536/04a-run_ica.py#L105-L107

04a-run_ica.py
```diff
@@ -110,8 +110,17 @@ def run_ica(subject, session=None):
...
     logger.info(gen_log_message(message=msg, step=4, subject=subject,
                                 session=session))
-    ica = ICA(method='fastica', random_state=config.random_state,
-              n_components=n_components[kind])
+
+    if config.ica_algorithm == 'picard':
+        fit_params = dict(fastica_it=5)
+    elif config.ica_algorithm == 'extended_infomax':
+        fit_params = dict(extended=True)
+    else:
```
agramfort, line 118: if it means fastica can you be explicit?

config.py
```diff
@@ -500,23 +500,47 @@
...
 
-# ``ica_decim`` : int
+# ``ica_algorithm`` : 'picard' | 'fastica' | 'extended_infomax'
+#   The ICA algorithm to use.
+
+ica_algorithm = 'picard'
+
+# ``ica_max_iterations`` : int
+#   Maximum number of iterations to decompose the data into independent
+#   components. A low number means to finish earlier, but a the cost that the
```
agramfort, line 526:
```suggestion
#   components. A low number means to finish earlier, but the consequence is that the
```

config.py
```diff
@@ -500,23 +500,47 @@
...
+#   algorithm may not always have finished converging. That means to ensure
```
agramfort, line 527:
```suggestion
#   algorithm may not have finished converging. To ensure
```

config.py
```diff
@@ -500,23 +500,47 @@
...
+#   convergence, pick a high number here (e.g. 3000); the algorithm will
```
agramfort, line 528:
```suggestion
#   convergence, pick a high number here (e.g. 3000); yet the algorithm will
```

config.py
```diff
@@ -780,6 +804,11 @@ def default_reject_comps_factory():
 if use_ssp and use_ica:
     raise ValueError('Cannot use both SSP and ICA.')
 
+if use_ica and ica_algorithm not in ('picard', 'fastica', 'extended_infomax'):
+    msg = ("Invalid ICA algorithm requested. Valid values for ica_algorithm "
+           "are: 'picard', 'fastica', and 'extended_infomax'.")
+    raise ValueError(msg)
```
agramfort, line 810: say what you got

04a-run_ica.py
```diff
@@ -110,8 +110,17 @@ def run_ica(subject, session=None):
...
                                 session=session))
-    ica = ICA(method='fastica', random_state=config.random_state,
-              n_components=n_components[kind])
+
+    if config.ica_algorithm == 'picard':
+        fit_params = dict(fastica_it=5)
+    elif config.ica_algorithm == 'extended_infomax':
+        fit_params = dict(extended=True)
+    elif config.ica_algorithm == 'fastica':
+        fit_params = None
```
agramfort, line 119: else error?

config.py
```diff
@@ -500,23 +500,47 @@
...
+
+# ``ica_decim`` : None | None
 #    The decimation parameter to compute ICA. If 5 it means
 #    that 1 every 5 sample is used by ICA solver. The higher the faster
-#    it is to run but the less data you have to compute a good ICA.
+#    it is to run but the less data you have to compute a good ICA. Set to
+#    ``1`` ``None`` to not perform an decimation.
 
-ica_decim = 11
+ica_decim = None
```
agramfort, line 543: maybe for later but we should have an auto mode to look for let's say 10e6 time samples using decim. It's not necessary to run ICA for too many samples. It's just slower. 

my 2c

04a-run_ica.py
hoechenberger: No need to error, because the input values should have been sanitized at this point already:

https://github.com/mne-tools/mne-study-template/pull/106/files#diff-fdf313ed2303059c44a9ebfffbef8eebR807-R811

hoechenberger: Hmm link doesn't work properly… anyway, in `config.py` we do:

```python
if use_ica and ica_algorithm not in ('picard', 'fastica', 'extended_infomax'):
     msg = (f"Invalid ICA algorithm requested. Valid values for ica_algorithm "
            f"are: 'picard', 'fastica', and 'extended_infomax', but received "
            f"{ica_algorithm}.")
     raise ValueError(msg)
```

config.py
hoechenberger: Ok interesting, I didn't know that!

agramfort:
Merge @hoechenberger if done here

PR #107, opened 2020-05-05
DOC: Remove 03-make_events.py from readme

hoechenberger:


agramfort:
good catch @hoechenberger !

PR #110, opened 2020-05-08
BF: Still render if we don't have a 3d backend

hoechenberger:
Do not render alignment when no valid MNE 3d backend has been set. **STCs will berendered** via Matplotlib in this case.

Depends on https://github.com/mne-tools/mne-python/pull/7708 and will therefore currently only work with `mne-python:master`

### DIFF
--- a/99-make_reports.py
+++ b/99-make_reports.py
@@ -101,11 +101,13 @@ def run_report(subject, session=None):
     rep.add_figs_to_section(figs, captions)
 
     if op.exists(fname_trans):
-        fig = mne.viz.plot_alignment(evoked.info, fname_trans,
-                                     subject=subject,
-                                     subjects_dir=config.get_subjects_dir(),
-                                     meg=True, dig=True, eeg=True)
-        rep.add_figs_to_section(fig, 'Coregistration')
+        # We can only plot the coregistration if we have a valid 3d backend.
+        if mne.viz.get_3d_backend() is not None:
+            fig = mne.viz.plot_alignment(evoked.info, fname_trans,
+                                         subject=subject,
+                                         subjects_dir=config.subjects_dir,
+                                         meg=True, dig=True, eeg=True)
+            rep.add_figs_to_section(fig, 'Coregistration')
 
         for evoked in evokeds:
             method = config.inverse_method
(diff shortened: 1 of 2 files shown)


config.py
```diff
@@ -21,6 +21,36 @@
...
+        msg = ('The custom configuration file specified in the '
+               'MNE_BIDS_STUDY_CONFIG environment variable could not be '
+               'found: {cfg_path}'.format(cfg_path=cfg_path))
+        raise ValueError(msg)
+
+    # Import configuration from an arbitrary path without having to fiddle
+    # with `sys.path`.
+    spec = importlib.util.spec_from_file_location(name='custom_config',
+                                                  location=cfg_path)
+    custom_cfg = importlib.util.module_from_spec(spec)
```
hoechenberger, line 50: Well this stuff really shouldn't be here, must have done something wrong while rebasing…

config.py
```diff
@@ -42,7 +72,7 @@
 # or
 # >>> bids_root = None  # Make use of the ``BIDS_ROOT`` environment variable.
 
-bids_root = None
+bids_root = None  # XXX
```
hoechenberger, line 75: this should also be reverted

99-make_reports.py
```diff
@@ -101,11 +101,13 @@ def run_report(subject, session=None):
     rep.add_figs_to_section(figs, captions)
 
     if op.exists(fname_trans):
-        fig = mne.viz.plot_alignment(evoked.info, fname_trans,
-                                     subject=subject,
-                                     subjects_dir=config.get_subjects_dir(),
-                                     meg=True, dig=True, eeg=True)
-        rep.add_figs_to_section(fig, 'Coregistration')
+        # We can only plot the coregistration if we have a valid 3d backend.
+        if mne.viz.get_3d_backend() is not None:
```
hoechenberger, line 105: wonder if we should add a log msg if we're not rendering? a warning, maybe?

99-make_reports.py
```diff
@@ -118,10 +120,27 @@ def run_report(subject, session=None):
...
+                    brain = stc.plot(views=['lat'], hemi='both',
+                                     initial_time=peak_time,  backend='mayavi')
+                    rep.add_figs_to_section(brain._figures[0],
+                                            evoked.comment)
+                else:
+                    import matplotlib.pyplot as plt
+                    fig_lh = plt.figure()
+                    fig_rh = plt.figure()
+
+                    brain_lh = stc.plot(views='lat', hemi='lh',
```
hoechenberger, line 136: btw these matplotlib plots look kinda awful (but at least better than nothing)
any ideas how to increase their resolution?

hoechenberger:
@agramfort I will merge once CI goes green, unless you have some more ideas regarding making the matplotlib STCs prettiert :)

PR #111, opened 2020-05-09
add debug with pdb option

agramfort:
have a look at @hoechenberger 

it does not jump me to the line where it crashed but it's a start

### DIFF
--- a/config.py
+++ b/config.py
@@ -809,7 +809,7 @@ if 'eeg' in ch_types:
 
 if on_error not in ('continue', 'abort', 'debug'):
     msg = (f"on_error must be one of 'continue' or 'abort' or 'debug', but "
-            "received {on_error}")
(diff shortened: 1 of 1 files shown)


config.py
```diff
@@ -806,9 +807,9 @@ def default_reject_comps_factory():
                'To turn it on, set use_ica=True.')
         logger.info(msg)
 
-if on_error not in ('continue', 'abort'):
-    msg = (f"on_error must be one of 'continue' or 'abort', but received "
-           f"{on_error}")
+if on_error not in ('continue', 'abort', 'debug'):
+    msg = (f"on_error must be one of 'continue' or 'abort' or 'debug', but "
+            "received {on_error}")
```
hoechenberger, line 812:
```suggestion
           f"received {on_error}")
```

PR #113, opened 2020-05-18
ENH: Add config.deriv_root

hoechenberger:
Allows specification of the root folder of the derivatives. Currently not meant to be user-settable.
Removes redundant code across the entire pipeline.

### DIFF
--- a/01-import_and_maxfilter.py
+++ b/01-import_and_maxfilter.py
@@ -227,7 +227,8 @@ def run_maxwell_filter(subject, session=None):
                 destination=destination)
 
             # Prepare a name to save the data
-            raw_fname_out = op.join(deriv_path, bids_basename + '_sss_raw.fif')
+            raw_fname_out = op.join(config.deriv_root, subject_path,
+                                    bids_basename + '_sss_raw.fif')
             raw_sss.save(raw_fname_out, overwrite=True)
 
             if config.plot:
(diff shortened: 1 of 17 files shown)


01-import_and_maxfilter.py
```diff
@@ -91,8 +91,7 @@ def run_maxwell_filter(subject, session=None):
 
         if run_idx == 0:  # XXX does this work when no runs are specified?
             # Prepare the pipeline directory in /derivatives
-            deriv_path = op.join(config.bids_root, 'derivatives',
-                                 config.PIPELINE_NAME)
+            deriv_path = config.deriv_root
```
agramfort, line 94: I would do:

deriv_path = config.deriv_path

or 

deriv_root = config.deriv_root

pick your poison

hoechenberger: I made it that way because in all scripts except the first one we do:
```python
    deriv_path = op.join(config.deriv_root, subject_path)
```
I could rename `deriv_path` to `deriv_root` in the first script?

hoechenberger: Also just to mention, once https://github.com/mne-tools/mne-bids/pull/410 has been merged, we can simplify this code further, as we won't have to manually construct so many paths anymore :)

hoechenberger: Addressed in ac19b7a

hoechenberger: I've unified things now in the `01` script to work like in the others: 95c7d09

hoechenberger: Haha wow what a mess :) Should be good now!

agramfort:
thx !

PR #116, opened 2020-05-19
DOC: Rework docstrings; no need to exclude emptyroom

hoechenberger:
Empty-room recordings are now automatically excluded from regular analysis, so there's no need to list `emptyroom` as excluded subject anymore.

Also update some more docstrings.

### DIFF
--- a/config.py
+++ b/config.py
@@ -94,16 +94,28 @@ rec = None
 
 space = None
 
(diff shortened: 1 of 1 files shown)


hoechenberger:
Tests failing because it needs 143bc1d to be merged first. Will rebase after that.

PR #119, opened 2020-05-19
Plot PSD based on data after measurement start only

hoechenberger:
Skip the (uninformative?) pre-measurement recording segment when plotting PSD of the filtered data. Would like to hear your opinion on that, @SophieHerbst & @agramfort

### DIFF
--- a/02-frequency_filter.py
+++ b/02-frequency_filter.py
@@ -94,7 +94,7 @@ def run_filter(subject, run=None, session=None):
         # plot power spectral density of the measurement data – skip the
         # time period before data acquisition.
         raw.plot_psd(tmin=raw.first_time,
-                     n_fft=len(raw.times[raw.first_samp:]),
+                     n_fft=len(raw.times[raw.first_samp:]) + 1,
                      average=True, area_mode='range')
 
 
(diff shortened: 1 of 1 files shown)


02-frequency_filter.py
```diff
@@ -91,9 +91,11 @@ def run_filter(subject, run=None, session=None):
         # plot raw data
         raw.plot(n_channels=50, butterfly=True)
 
-        # plot power spectral densitiy
-        raw.plot_psd(area_mode='range', tmin=10.0, tmax=100.0,
-                     fmin=0., fmax=50., average=True)
+        # plot power spectral density of the measurement data – skip the
+        # time period before data acquisition.
+        raw.plot_psd(tmin=raw.first_time,
```
agramfort, line 96: are you sure it's necessary? there is no data before `raw.first_time` in the fif file.

hoechenberger: But there is? With Elekta data, there often is like 20–30 secs or so. I guess that's the reason why `first_time` and `first_samp` were introduced in the first place…
And no I don't know if it _is_ necessary, but I thought it's something we should talk about :)

hoechenberger: Also raised this question upstream, https://github.com/mne-tools/mne-python/issues/7799

hoechenberger:
Discussion with @agramfort showed that my concept about `raw.first_samp` and `raw.first_time` was incorrect: _there is no data_ before `raw.first_samp`.

PR #120, opened 2020-05-19
RF: Factor out Maxwell filter step into own function

hoechenberger:
The reason I'm doing all of this refactoring is to make the code easier to understand and maintain, and also I hope I can reuse more "measurement processing" code when doing the "emptyroom processing".


agramfort:
thx @hoechenberger

agramfort:
thx @hoechenberger

PR #125, opened 2020-05-22
Add config.get_subject_path(), config.get_subject_deriv_path()

hoechenberger:
The common task of creating the path to the directory where the data is stored for a particular participant & recording kind has now been moved into the `config.get_subject_path()` function.


hoechenberger:
Now looking at this again, we could even consider introducing a function `get_deriv_path()`...

agramfort:
do what you feel is best

hoechenberger:
Ok let's see what CI says

hoechenberger:
I love it when the diff shows a negative net difference of the line count :-)

PR #128, opened 2020-05-26
Add recording kind to fwd, cov, inv, stc filenames

hoechenberger:
If a dataset includes both, MEG and EEG data, we need a way to tell the different forward models, noise covariance matrices, inverse operators, and STCs apart by their filenames.

In line with most of the other existing code, I'm using `config.get_kind()` here (instead of `config.ch_types`), so currently, only separation by **kind** is possible and usability is limited. We will need to address this at a later time – it's beyond the scope of this PR, and currently not too urgent, I believe.

### DIFF
--- a/12-make_inverse.py
+++ b/12-make_inverse.py
@@ -24,9 +24,10 @@ logger = logging.getLogger('mne-study-template')
 
 @failsafe_run(on_error=on_error)
 def run_inverse(subject, session=None):
+    kind = config.get_kind()
     deriv_path = config.get_subject_deriv_path(subject=subject,
                                                session=session,
-                                               kind=config.get_kind())
+                                               kind=kind)
 
     bids_basename = make_bids_basename(subject=subject,
                                        session=session,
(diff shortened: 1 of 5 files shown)


hoechenberger:
@SophieHerbst @dnacombo
Do you have to handle data with simultaneous MEG & EEG recordings, saved to the same file, regularly? Or is this a rare thing for you? Just asking to help me prioritize things.

12-make_inverse.py
```diff
@@ -38,9 +39,9 @@ def run_inverse(subject, session=None):
                                        space=config.space)
 
     fname_ave = op.join(deriv_path, bids_basename + '-ave.fif')
-    fname_fwd = op.join(deriv_path, bids_basename + '-fwd.fif')
-    fname_cov = op.join(deriv_path, bids_basename + '-cov.fif')
-    fname_inv = op.join(deriv_path, bids_basename + '-inv.fif')
+    fname_fwd = op.join(deriv_path, f'{bids_basename}_{kind}-fwd.fif')
+    fname_cov = op.join(deriv_path, f'{bids_basename}_{kind}-cov.fif')
```
agramfort, line 43: why also for the cov?

hoechenberger: To handle the case where the user asked us to compute the MEG cov based on an empty-room recording. The EEG cov will still be computed from the measurement data.
(Currently working on this in another branch)

Or do you have better idea?

agramfort: you must use only one noise cov for source imaging. As long as EEG is present you must use prestim. let's talk it will be easier.

hoechenberger:
After discussion with @agramfort, this is not the way to do it, closing PR.

dnacombo:
> @SophieHerbst @dnacombo
> Do you have to handle data with simultaneous MEG & EEG recordings, saved to the same file, regularly? Or is this a rare thing for you? Just asking to help me prioritize things.

Happens, but not so often.

hoechenberger:
> Happens, but not so often.

Ok, thanks for letting me know!

PR #131, opened 2020-05-28
[MRG] Rename "plot" option to more generic "interactive"

hoechenberger:
In expectation of more interactive features to land soon :-)


01-import_and_maxfilter.py
```diff
@@ -273,8 +273,11 @@ def run_maxwell_filter(subject, session=None):
         os.makedirs(os.path.dirname(raw_fname_out), exist_ok=True)
 
         raw_out.save(raw_fname_out, overwrite=True)
-        if config.plot:
-            raw_out.plot(n_channels=50, butterfly=True)
+        if config.interactive:
+            (raw_out
+             .copy()
```
agramfort, line 278: why do you copy? you don't want to let user set bad channels or set annotations?

01-import_and_maxfilter.py
```diff
@@ -273,8 +273,11 @@ def run_maxwell_filter(subject, session=None):
...
+             .pick_types(meg=True)
```
agramfort, line 279: why only meg?

01-import_and_maxfilter.py
hoechenberger: Very good point. I copy() because I want to only plot MEG channels, because only they are touched during Maxwell filtering. Apparently there is no `pick` parameter in `Raw.plot()`, therefore I copy'd before picking (see your next question). But you're right, this won't allow users to set bads manually. My idea was to later extend it such that we extract the bads from the copy's info and inject them into the original data that will be saved to disk… Yes, seems awkward…

01-import_and_maxfilter.py
agramfort: honestly I would say 

`raw_out.plot(n_channels=50, butterfly=True)`

is a wiser default. You see very quickly issues when using butterfly mode.

I am now wondering if making a copy of raw here is really a wise thing.
Can it potentially double the memory usage of the script?

hoechenberger: I will revert this to use butterfly mode. I think users can switch to the other view (what is it called?) by pressing `b` anyway, right? :)

agramfort:
yes

>

hoechenberger:
@agramfort good to merge from my end

agramfort:
thx @hoechenberger

PR #135, opened 2020-06-04
Pass subjects_dir to plot_alignment() & stc.plot()

hoechenberger:
Otherwise the required files may not be found in some cases.

### DIFF
--- a/99-make_reports.py
+++ b/99-make_reports.py
@@ -124,7 +124,7 @@ def run_report(subject, session=None):
         if mne.viz.get_3d_backend() is not None:
             fig = mne.viz.plot_alignment(evoked.info, fname_trans,
                                          subject=subject,
-                                         subjects_dir=config.subjects_dir,
+                                         subjects_dir=subjects_dir,
                                          meg=True, dig=True, eeg=True)
             rep.add_figs_to_section(figs=fig, captions='Coregistration',
                                     section='Coregistration')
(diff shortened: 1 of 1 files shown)


99-make_reports.py
```diff
@@ -124,7 +124,7 @@ def run_report(subject, session=None):
         if mne.viz.get_3d_backend() is not None:
             fig = mne.viz.plot_alignment(evoked.info, fname_trans,
                                          subject=subject,
-                                         subjects_dir=config.subjects_dir,
+                                         subjects_dir=subjects_dir,
```
hoechenberger, line 127: We're now correctly using the `subjects_dir` variable, which is generated further up via `config.get_fs_subjects_dir()`.

99-make_reports.py
```diff
@@ -164,12 +164,17 @@ def run_report(subject, session=None):
...
+                        views='lat', hemi='lh',
+                        initial_time=peak_time,
+                        backend='matplotlib',
+                        subjects_dir=subjects_dir,
+                        figure=fig_lh)
+                    brain_rh = stc.plot(
+                        views='lat', hemi='rh',
+                        initial_time=peak_time,
+                        subjects_dir=subjects_dir,
+                        backend='matplotlib', figure=fig_rh)
```
agramfort, line 177: I would not had code matplotlib here. It should use matplotlib only because mayavi and pyvista are not available

hoechenberger: That's what we do :) You're in the `if mne.viz.get_3d_backend() is None` branch here :)

PR #137, opened 2020-06-04
Only epoch conditions the user requested to analyze; improve documentation

hoechenberger:
- Only epoch conditions we intend intend to process: Previously, we'd epoch based on **all** events in the data, and only filter down to the conditions in `config.conditions` when constructing Evokeds. Now, we only create the required epochs in the first place.
- Move `conditions` config setting up (first setting of the "Epoch" section – makes more sense that wat IMHO!), and improve (and slightly correct) its documentation


hoechenberger:
Whoops this was opened in error, will re-open or re-submit when ready

PR #143, opened 2020-06-08
ENH: Contrast Evokeds in sensor space

hoechenberger:
This PR adds support for sensor-space contrasts on the `Evoked` level: one can now calculate the difference waves of two conditions.

This introduces the following changes:

- **New configuration setting** `contrasts`, which is a list of tuples in the form `[(cond_1, cond_2), (cond_3, cond_4), ...]`. Each tuple specifies one contrast.
- `config.decoding_conditions` was **removed**; instead, the new `contrasts` setting is used to define which conditions to use in decoding. This ensures consistency within the analysis.
- A **new setting** `decode`, which takes a boolean value, was added, to turn on or off decoding.

This PR focuses on **sensor-space only** to keep the diff smaller and easier to review.

Please note that step 99 (reports), in addition to difference ERF/ERP plotting, already received some small changes in preparation for the coming source-space contrasts.

Example from the report generated from `ds000246`:
Screenshot 2020-06-08 at 12 22 11


agramfort:
thx @hoechenberger

PR #144, opened 2020-06-08
Use MNE's default cross-validation strategies in decoding, noise covariance estimation

hoechenberger:
Do not construct custom cross-validation objects anymore.

- Use MNE's default CV strategy while decoding: only pass the number of splits, but otherwise leave everything else to MNE to decide.
- Use MNE's default CV strategy (incl. number of splits) when estimating the noise covariance.

Current user-facing behavior remains unchanged.

I believe the only reason the CV objects were initially created was to allow for shuffling (which MNE does not do by default). Following our discussions in #141 and #73, we explicitly want to **avoid shuffling**: MNE does the right thing for us already. So let's drop the manually created CV objects.


agramfort:
thx @hoechenberger !

PR #147, opened 2020-07-05
[MRG] Treat grand-averaged data as if coming from "sub-average" & make derivatives naming more consistent

hoechenberger:
As discussed in https://github.com/mne-tools/mne-bids/issues/454#issuecomment-653858194

This also allows us to use BIDSPath in more (all?) places now when creating paths.

Also cleans up / unifies some naming schemes.

Example derivatives folder after running the study-template on the somato dataset:

```
.
├── dataset_description.json
├── sub-01
│   └── meg
│       ├── sub-01_task-somato_ave.fif
│       ├── sub-01_task-somato_cleaned-epo.fif
│       ├── sub-01_task-somato_cond-somato-event1_inverse-dSPM_hemi-lh.stc
│       ├── sub-01_task-somato_cond-somato-event1_inverse-dSPM_hemi-rh.stc
│       ├── sub-01_task-somato_cond-somato-event1_inverse-dSPM_morph-fsaverage_hemi-lh.stc
│       ├── sub-01_task-somato_cond-somato-event1_inverse-dSPM_morph-fsaverage_hemi-rh.stc
│       ├── sub-01_task-somato_cov.fif
│       ├── sub-01_task-somato_epo.fif
│       ├── sub-01_task-somato_filt_raw.fif
│       ├── sub-01_task-somato_fwd.fif
│       ├── sub-01_task-somato_inv.fif
│       ├── sub-01_task-somato_nosss_raw.fif
│       ├── sub-01_task-somato_report.html
│       ├── sub-01_task-somato_ssp-proj.fif
│       └── sub-01_task-somato_trans.fif
└── sub-average
    └── meg
        ├── sub-average_task-somato_ave.fif
        ├── sub-average_task-somato_cond-somato-event1_inverse-dSPM_morph-fsaverage_hemi-lh.stc
        ├── sub-average_task-somato_cond-somato-event1_inverse-dSPM_morph-fsaverage_hemi-rh.stc
        └── sub-average_task-somato_report.html

```

**Note that this depends on https://github.com/mne-tools/mne-bids/pull/470, and tests will fail until this has been merged upstream.**

PR #148, opened 2020-07-06
[MRG] BF: Include STC plots in reports

hoechenberger:
Fix a bug introduced in GH-147 that prevented STC rendering in the reports (except for grand average reports).


agramfort:
thx @hoechenberger

PR #149, opened 2020-07-06
TESTS: Try setting resource_class to "large"

hoechenberger:
As suggested in the Circle docs
May speed up testing


hoechenberger:
"This job has been blocked because resource-class:large is not available on your plan. Please upgrade to continue building."

Never mind…

PR #151, opened 2020-07-08
better ds000246

agramfort:


agramfort:
looks better https://514-167957354-gh.circle-artifacts.com/0/reports/ds000246/sub-0001_task-AEF_report.html

OLD: https://512-167957354-gh.circle-artifacts.com/0/reports/ds000246/sub-0001_task-AEF_report.html

PR #154, opened 2020-07-09
Add Raw PSD plots

hoechenberger:


hoechenberger:
Circle is having some issues right now…

This PR results in the emptyroom PSD plotted twice: first, on top of the report with all the other Raw info; and secondly, at the bottom in the Empty-Room section. This is something I will look into at a later time: getting away from `parse_folders` (which compiled the stuff at the top) and constructing all sections manually for better layout & UX.

hoechenberger:
Closing in favor of #155

PR #155, opened 2020-07-09
add psd to report for raw

agramfort:


hoechenberger:
I restarted Circle twice and it still keeps failing. I wonder if we somehow broke something?

agramfort:
we blow up the memory here @larsoner 

reason is this:

https://github.com/mne-tools/mne-python/blob/master/mne/time_frequency/psd.py#L20

we compute the FFTs for all frames and only average in the end.

do you think you can have a look?

Memory usage for report with PSD:
Screenshot 2020-07-10 at 09 07 48

Memory usage for report without PSD:
Screenshot 2020-07-10 at 09 10 43

hoechenberger:
Thanks @agramfort!

PR #157, opened 2020-07-14
WIP: Pick the requested channel type early

hoechenberger:
So we don't have to worry about it during later processing steps!

### DIFF
--- a/01-import_and_maxfilter.py
+++ b/01-import_and_maxfilter.py
@@ -208,22 +208,6 @@ def load_data(bids_basename):
                         extra_params=extra_params,
                         kind=config.get_kind())
 
-    # Select only the channel types we wish to analyze.
-    if config.get_kind() == 'meg' and ('mag' in config.ch_types or
-                                       'grad' in config.ch_types):
-        picks = mne.pick_types(raw.info, eog=True, ecg=True)
-
-        if 'mag' in config.ch_types:
-            picks += mne.pick_types(raw.info, meg='mag')
-        if 'grad' in config.ch_types:
-            picks += mne.pick_types(raw.info, meg='grad')
-    elif config.get_kind() == 'meg':
-        picks = mne.pick_types(raw.info, meg=True, eog=True, ecg=True)
-    else:
-        picks = mne.pick_types(raw.info, eeg=True, eog=True, ecg=True)
-
-    raw.pick(picks)
-
     # XXX hack to deal with dates that fif files cannot handle
     if config.daysback is not None:
         raw.anonymize(daysback=config.daysback)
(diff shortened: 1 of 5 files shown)


hoechenberger:
Oh wow this went smoother than I had expected!

01-import_and_maxfilter.py
```diff
@@ -208,6 +208,22 @@ def load_data(bids_basename):
...
+        picks = raw.copy().pick_types(eog=True, ecg=True).ch_names
+
+        if 'mag' in config.ch_types:
+            picks += raw.copy().pick_types(meg='mag').ch_names
+        if 'grad' in config.ch_types:
+            picks += raw.copy().pick_types(meg='grad').ch_names
+    elif config.get_kind() == 'meg':
+        picks = raw.copy().pick_types(meg=True, eog=True, ecg=True).ch_names
+    else:
+        picks = raw.copy().pick_types(eeg=True, eog=True, ecg=True).ch_names
```
agramfort, line 223: It seems an overkill to copy the raw data to get ch names. Use pick functions and not the instances methods

hoechenberger: At this point, the data isn't preloaded though… So not _too_ much overhead! But ok, will switch to `mne.pick_types`

01-import_and_maxfilter.py
```diff
@@ -208,6 +208,22 @@ def load_data(bids_basename):
...
+        picks = mne.pick_types(raw.info, eog=True, ecg=True)
+
+        if 'mag' in config.ch_types:
+            picks += mne.pick_types(raw.info, meg='mag')
+        if 'grad' in config.ch_types:
+            picks += mne.pick_types(raw.info, meg='grad')
+    elif config.get_kind() == 'meg':
+        picks = mne.pick_types(raw.info, meg=True, eog=True, ecg=True)
+    else:
+        picks = mne.pick_types(raw.info, eeg=True, eog=True, ecg=True)
```
agramfort, line 223: I realize that if you want to analyze grad or mag maybe you still want to use the full MEG sensor array in maxfilter. So I would pick mag or grad after calling maxfilter.

can you check that maxfilter works uniformly if you select grad or mag only on the sample data?

hoechenberger: Excellent point, I never thought about that! I will move the picking after Maxfilter. Do you still want me to run the test you proposed, anyway?

agramfort:
Just pick afterwards

agramfort:
CIs complain but diff looks good

hoechenberger:
Hopefully it works now.

Before ba14f2e:
mprof


After ba14f2e:
mprof3

01-import_and_maxfilter.py
```diff
@@ -222,13 +222,13 @@ def load_data(bids_basename):
...
 
     raw.load_data()
-
     if hasattr(raw, 'fix_mag_coil_types'):
         raw.fix_mag_coil_types()
 
     return raw
 
 
+# @profile
```
agramfort, line 231: to cleanup

config.py
```diff
@@ -1107,3 +1107,22 @@ def plot_auto_scores(auto_scores):
         figs.append(fig)
 
     return figs
+
+
+def get_picks(info):
+    """Retun the names of the channels we wish to analyze.
```
agramfort, line 1113:
```suggestion
    """Return the names of the channels we wish to analyze.
```

agramfort:
thx @hoechenberger !

PR #161, opened 2020-07-16
DOC: Add more info on expected input data (should be anonymized; bads should be marked)

hoechenberger:


agramfort:
thx @hoechenberger

PR #164, opened 2020-07-27
Misc. cleanups

hoechenberger:
- Replace use of os.path.join with BIDSPath where possible
- More consistency of file names
- Better filename sanitization in some places
- And some other minor cleanups


agramfort:
Thx @hoechenberger

PR #168, opened 2020-08-20
Do not anonymize -- should be done in BIDS

hoechenberger:


hoechenberger:
Closing in favor of #160

PR #176, opened 2020-08-27
[MRG] Use bids_path instead of bids_basename

hoechenberger:
Following https://github.com/mne-tools/mne-bids/pull/523


agramfort:
thx @hoechenberger !

PR #183, opened 2020-09-29
MRG, ENH: Conduct dimensionality reduction before ICA

hoechenberger:
We used to **not** reduce dimensionality, and only limit the number of ICs to fit. 

This is an attempt to avoid running ICA on rank-deficient data: we now reduce dimensionality using PCA before the ICA step. We get rid of the option to select how many ICs to fit; the idea being that this can implicitly be controlled using the PCA step. Let's see how this performs and how many ICs it produces. If it turns out we get "too many" ICs (and ICA therefore takes too long / doesn't converge), we should add the previous option back in, in addition to this new one.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -37,7 +37,9 @@ jobs:
               conda update -n base -c defaults conda
               pip install numpy scipy pandas json_tricks matplotlib seaborn nibabel coloredlogs python-picard
               pip install -U scikit-learn
-              pip install --upgrade https://api.github.com/repos/mne-tools/mne-python/zipball/master
+              # pip install --upgrade https://api.github.com/repos/mne-tools/mne-python/zipball/master
+              # XXX remove before merging
+              pip install https://github.com/hoechenberger/mne-python/archive/ica-max_pca_components.zip
               git clone https://github.com/mne-tools/mne-bids.git --depth 1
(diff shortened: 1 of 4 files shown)


hoechenberger:
Oh. `max_pca_components` doesn't accept a float. Bummer.

hoechenberger:
The log message about explained variance will show an incorrect value (100%) until we address this upstream issue: https://github.com/mne-tools/mne-python/issues/8322

hoechenberger:
I've changed things again, following the extensive discussions over at mne-python, and thanks to https://github.com/mne-tools/mne-python/pull/8321

We're now always using `max_pca_components=0.9999` (hardcoded) to avoid problems due to rank deficiency. The user can control how many principal components to feed to ICA via the `ica_n_componens` option, which maps to `ICA`'s `n_components`.

This is good to be merged.

ICA report:
https://731-167957354-gh.circle-artifacts.com/0/reports/ds000248_ica/sub-01_task-audiovisual_proc-ica_report.html

Full report:
https://731-167957354-gh.circle-artifacts.com/0/reports/ds000248_ica/sub-01_task-audiovisual_report.html

agramfort (approved):

agramfort:
Thx @hoechenberger

PR #184, opened 2020-10-06
Plot decoding metrics and save results to TSV

hoechenberger:
Ths is an item of #152.

Currently not doing any group metrics.


agramfort:
why xlsx and not csv or .mat or hdf5? just a naive question.

hoechenberger:
> why xlsx and not csv or .mat or hdf5? just a naive question.

We're already saving the cross-validation scores to a `.mat` file. I thought it would be nice to additionally provide an easily "human-parsable" file with the decoding metrics as well, so you could e.g. use that data directly to for plotting or to pipe it into your favorite statistical software. Currently I'm only writing the mean of the CV scores for each time point to that file.

I picked Excel over CSV because with CSV you will always run into issues on some European language settings, e.g. in Germany, by default a comma is a decimal separator, and the column separator would then be a semicolon. This leads to interoperability problems. Excel doesn't have this problem.

But this is really just an **idea**, maybe we don't need this file at all… idk. I just thought it's so damn cheap to create it, so why not :)

agramfort:
hum it adds a dependency to the software for no strong reason (it's just
nice to have). For me it is slippery slope.

hoechenberger:
> hum it adds a dependency to the software for no strong reason (it's just
> nice to have). For me it is slippery slope.

Yeah but openpyxl is really a standard thing if you're using pandas… and assuming most users will run Anaconda anyway, where it's included… idk! I'm not even sure just how useful it is. To avoid the dependency thing we could move to a plain CSV or TSV. Would be slightly more inconvenient for some users but I guess it would be okay

hoechenberger:
The plots we now generate:
https://739-167957354-gh.circle-artifacts.com/0/reports/ds000246/sub-0001_task-AEF_report.html#16
https://739-167957354-gh.circle-artifacts.com/0/reports/ds000248/sub-01_task-audiovisual_report.html#30

hoechenberger:
@agramfort We're now writing to a TSV instead of XLSX.

99-make_reports.py
```diff
@@ -115,6 +118,33 @@ def plot_auto_scores(subject, session):
...
+    mean_scores = cross_val_scores.mean(axis=0)
+    max_scores = cross_val_scores.max(axis=0)
+    min_scores = cross_val_scores.min(axis=0)
+
+    fig, ax = plt.subplots()
+    ax.axhline(0.5, ls='--', lw=0.5, color='black', label='chance')
+    if times.min() < 0 < times.max():
+        ax.axvline(0, ls='-', lw=0.5, color='black')
+    ax.fill_between(x=times, y1=min_scores, y2=max_scores, color='lightgray',
+                    alpha=0.5, label='range')
```
hoechenberger, line 133:
```suggestion
                    alpha=0.5, label='range [min, max]')
```

09-group_average_sensors.py
```diff
@@ -0,0 +1,208 @@
...
+    if config.decode:
+        average_decoding()
+
+    # ts_args = dict(gfp=True, time_unit='s')
+    # topomap_args = dict(time_unit='s')
+
+    # for idx, evokeds in enumerate(all_evokeds):
+    #     all_evokeds[idx].plot_joint(title=config.conditions[idx],
+    #                                 ts_args=ts_args, topomap_args=topomap_args)  # noqa: E501
+
```
agramfort, line 201: cleanup?

09-group_average_sensors.py
```diff
@@ -0,0 +1,208 @@
+"""
```
agramfort, line 1: you should have use git mv here to keep track of last changes

agramfort (commented):
besides LGTM

+1 for MRG

09-group_average_sensors.py
hoechenberger: I tried to fix this by reverting the commit, using git mv, and then manually merging the changes again. Still GH doesn't seem to know about the history :(

agramfort (approved):
merge when green

thx @hoechenberger

hoechenberger:
Individual participant:
Screenshot 2020-10-07 at 14 35 26

Grand average:
Screenshot 2020-10-07 at 14 35 10

jasmainak:
naive question, why is the decoding not at 50% during baseline?

hoechenberger:
@jasmainak 
> naive question, why is the decoding not at 50% during baseline?

Damn, I hoped nobody would notice, but you're already number two ;)
The reason is that the contrast we're testing in this experiment doesn't really make sense. Consider this really just a test of the plotting functionality, not of the decoding itself.

I would love to add a better example, do you have any suggestions for suitable datasets?

Ideally:
- data should be relatively clean and require very little pre-processing
- the experimental design should be simple
- we need ~5+ participants

hoechenberger:
@jasmainak 
Here is the result for ds000248, but there's unfortunately only a single participant.

jasmainak:
this looks clean. Have you considered ds000117 for dataset?

PR #187, opened 2020-10-13
Adjust ICA for upstream MNE changes

hoechenberger:


hoechenberger:
CI is failing because the ds000248 dataset on OpenNeuro got corrupted 💩 Already opened a ticket.

PR #190, opened 2020-10-31
Allow debug

agramfort:
seems to work !


hoechenberger:
Umm I need to fix CI…

PR #202, opened 2020-11-04
report fixes: use Agg, close figures and take into account task

agramfort:


99-make_reports.py
```diff
@@ -391,25 +398,13 @@ def run_report(subject, session=None):
 
     fname_report = bids_path.copy().update(suffix='report', extension='.html')
     rep.save(fname=fname_report, open_browser=False, overwrite=True)
+    import matplotlib.pyplot as plt  # nested import to help joblib
```
hoechenberger, line 410: We do have the same import at the top level already, is that intentional?

config.py
```diff
@@ -1049,9 +1049,9 @@
                'To turn it on, set use_ica=True.')
         logger.info(msg)
 
-if on_error not in ('continue', 'abort'):
-    msg = (f"on_error must be one of 'continue' or 'abort', but received "
-           f"{on_error}.")
+if on_error not in ('continue', 'abort', 'debug'):
+    msg = (f"on_error must be one of 'continue', 'debug' or 'abort', "
+           f"but received: {on_error}.")
```
hoechenberger, line 1054: Whoops, did I accidentally undo your change here? Good thing you caught it!

99-make_reports.py
```diff
@@ -13,7 +13,11 @@
 import numpy as np
 from scipy.io import loadmat
 
+import matplotlib
+
 import matplotlib.pyplot as plt
+import matplotlib
+matplotlib.use('Agg')
```
hoechenberger, line 20:
```suggestion
```

You already have this in line 29

99-make_reports.py
agramfort: I don't why I need to do this ... otherwise I have a crash with joblib :(

config.py
agramfort: ok no worries

99-make_reports.py
hoechenberger: But can you at least remove the top-level import then?

The crash also is something new, probably related to the Agg backend?

hoechenberger:
Great, thanks @agramfort!

PR #203, opened 2020-11-04
[MRG] Refine docs

hoechenberger:
Still working on the inclusion of all config params, but slowly getting there!


hoechenberger:
The only undocumented settings are now only those listed under "ADVANCED" and "GROUP AVERAGE SENSORS" in config.py

agramfort (approved):

PR #204, opened 2020-11-05
add new params to run.py script

agramfort:


config.py
```diff
@@ -843,7 +843,8 @@
 Set the Study Template logging verbosity.
 """
 
-mne_log_level: Literal['info', 'error'] = 'error'
+# mne_log_level: Literal['info', 'error'] = 'error'
+mne_log_level: Literal['info', 'error'] = 'info'
```
hoechenberger, line 847: I'm okay with info, but could you remove line 846 please?

config.py
```diff
@@ -1077,12 +1100,20 @@ def get_subjects():
...
 
 def get_task():
+    env = os.environ
+    valid_tasks = get_entity_vals(bids_root, entity_key='task')
+
+    if env.get('MNE_BIDS_STUDY_TASK'):
+        task = env['MNE_BIDS_STUDY_TASK']
+        if task not in valid_tasks:
+            raise ValueError(
+                f'Invalid task. It can be {valid_tasks} but got {task}')
```
hoechenberger, line 1110:
```suggestion
                f'Invalid task. It can be: {", ".join(valid_tasks)} but got: {task}')
```

run.py
```diff
@@ -32,24 +32,38 @@
...
+
     module_name = script.replace('.py', '')
     runpy.run_module(mod_name=module_name, run_name='__main__')
 
     logger.info(f'Successfully finished running: {script}')
 
 
-def process(steps, config, root_dir=None):
+def process(steps, config, root_dir=None, task=None, session=None,
+            run=None, subject=None):
```
hoechenberger, line 66: I'm not sure about the order of params here. Wonder if it would make more sense to kind of follow the "BIDS order"? i.e. subject, session, task, run? WDYT?

hoechenberger:
There seems to be a bug in task handling, therefore CI fails

hoechenberger:
New help looks like this (`python run.py --help`):
Screenshot 2020-11-05 at 21 34 36

hoechenberger:
Thanks @agramfort!

PR #210, opened 2020-11-06
Make run.py executable, move scripts to scripts/ dir, remove Makefile

hoechenberger:


agramfort:
I think you need to udpate readme that links to scripts

hoechenberger:
@agramfort This should be good once CI comes back green

agramfort:
beautiful !

PR #212, opened 2020-11-07
More doc improvements

hoechenberger:
- Buttons on the landing page!
- Add "getting started" category to navbar
- Fix Edit button -> now takes you to the GH editor
- Add an announcement bar at top of site, which we can use to inform users about new releases etc


agramfort:
getter better every day https://984-167957354-gh.circle-artifacts.com/0/site/getting_started/install.html !

PR #213, opened 2020-11-15
fixes and use standard location for mf cal and ct files

agramfort:


hoechenberger:
We don't want that because the location of these files is now specified in BIDS, therefore these settings were removed...

BIDSPath.meg_fine_calib or something (I'm in the phone, cannot check now)

scripts/01-import_and_maxfilter.py
```diff
@@ -74,6 +75,37 @@ def init_dataset():
...
+
+
+def get_mf_ctc_fname(subject, session):
+    if config.mf_ctc_fname is None:
+        mf_ctc_fpath = BIDSPath(subject=subject,
+                                session=session,
+                                acquisition='crosstalk',
+                                suffix='meg',
+                                extension='fif',
+                                root=config.bids_root).fpath
```
hoechenberger, line 101: You should use [`BIDSPath.meg_crosstalk_fpath`](https://mne.tools/mne-bids/stable/generated/mne_bids.BIDSPath.html#mne_bids.BIDSPath.meg_crosstalk_fpath)

scripts/01-import_and_maxfilter.py
```diff
@@ -74,6 +75,37 @@ def init_dataset():
...
 
 
+def get_mf_cal_fname(subject, session):
+    if config.mf_cal_fname is None:
+        mf_cal_fpath = BIDSPath(subject=subject,
+                                session=session,
+                                acquisition='calibration',
+                                suffix='meg',
+                                extension='dat',
+                                root=config.bids_root).fpath
```
hoechenberger, line 85: You should use [`BIDSPath. meg_calibration_fpath `](https://mne.tools/mne-bids/stable/generated/mne_bids.BIDSPath.html#mne_bids.BIDSPath.meg_calibration_fpath)

agramfort:
I let you take over this one. Thanks

>

hoechenberger:
@agramfort 
CI passes now.

But how do we got about crosstalk and fine-calibration files in "non-standard locations"? IMHO, we should just assume everything is in BIDS, so we don't need special handling / config options. What was the reason you added support for the non-standard locations? Is that something that just made using the study template more convenient, or was it a necessity for some reason?

hoechenberger:
Thanks @agramfort!

agramfort:
thx @hoechenberger !

PR #219, opened 2020-11-21
MRG: add run_freesurfer.py

agramfort:
@hoechenberger still WIP but let me know what you think


hoechenberger (commented):
To retrieve the subjects dir etc, shouldn't we try to reuse the functions we already have in the study-template scripts?

PR #223, opened 2020-12-11
Remove MEG fine-cal & crosstalk config options

hoechenberger:
We can now find these files thanks to the BIDS specification.

Re-enable ds000248 tests, thanks to @agramfort who managed to update the
dataset.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -56,12 +56,12 @@ jobs:
         #       pip install --user --upgrade --progress-bar off mayavi
         #       pip install --user --upgrade --progress-bar off PySurfer[save_movie]
 
-        # - run:
-        #     name: Install datalad to conveniently pull testing datasets
-        #     command: |
-        #       conda install datalad
-        #       git config --global user.email "user@mne.com"
-        #       git config --global user.name "mne user"
+        - run:
+            name: Install datalad to conveniently pull testing datasets
+            command: |
+              conda install datalad
+              git config --global user.email "user@mne.com"
+              git config --global user.name "mne user"
 
         # Save Python cache
         - save_cache:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -56,12 +56,12 @@ jobs:
         #       pip install --user --upgrade --progress-bar off mayavi
         #       pip install --user --upgrade --progress-bar off PySurfer[save_movie]
 
-        - run:
-            name: Install datalad to conveniently pull testing datasets
-            command: |
-              conda install datalad
-              git config --global user.email "user@mne.com"
-              git config --global user.name "mne user"
+        # - run:
+        #     name: Install datalad to conveniently pull testing datasets
+        #     command: |
+        #       conda install datalad
+        #       git config --global user.email "user@mne.com"
+        #       git config --global user.name "mne user"
 
         # Save Python cache
         - save_cache:
@@ -76,6 +76,7 @@ jobs:
                which python
                git-annex version
                datalad --version
+               openneuro-py --version
                mne_bids --version
                mne --version
                python -c "import mne; mne.sys_info()"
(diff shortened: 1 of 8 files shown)


agramfort:
I think we need to wait for datalad to get the new version or we download with openneuro cli

hoechenberger:
Let's hope the sync works…

hoechenberger:
@sappelhoff `matchingpennies` is now the only dataset we're not retrieving from OpenNeuro. Care to publish it there? :) This would allow us to simplify our test setup, now that we have [`openneuro-py`](https://github.com/hoechenberger/openneuro-py).

hoechenberger:
@jasmainak I've switched all OpenNeuro dataset downloads to [`openneuro-py`](https://github.com/hoechenberger/openneuro-py), meaning we won't need `datalad` for those anymore!

jasmainak:
lovely, and I see you managed to do it without dependency on `awscli` (although there is an unnecessary dependency on `click` :P)

hoechenberger:
> although there is an unnecessary dependency on `click` :P

You guys never want to take advantage of stuff that makes your life easier, eh? 😅😜

hoechenberger:
@agramfort It passes!! Good to merge

agramfort:
thanks @hoechenberger !

sappelhoff:
>  Care to publish it there?

I am afraid I can't do that:
1. I think having the dataset on one service provider is more clear
1. openneuro datasets are licensed under CC0, but eeg_matchingpennies already has an ODBL license
1. the eeg_matchingpennies dataset *is* subject to the GDPR, so uploading to openneuro is not permitted

but that probably means that we cannot use eeg_matchingpennies in the study template anymore, right?

PS: cool work!

hoechenberger:
> 3\. the eeg_matchingpennies dataset _is_ subject to the GDPR, so uploading to openneuro is not permitted

You sure about this? If the data is anonymized, I don't think it should be a problem. When collecting research data, participants consent to anonymized publication, so this should all be covered.

hoechenberger:
> but that probably means that we cannot use eeg_matchingpennies in the study template anymore, right?

We can, it just adds some overhead because it's the only dataset that needs to be fetched via data datalad

hoechenberger:
> > 3. the eeg_matchingpennies dataset _is_ subject to the GDPR, so uploading to openneuro is not permitted
> 
> You sure about this? If the data is anonymized, I don't think it should be a problem. When collecting research data, participants consent to anonymized publication, so this should all be covered.

Also… how come it's okay to publish this on GitHub, then? I must be missing something here?

sappelhoff:
> You sure about this?

unsure and insecure enough to not want to go that direction - and I don't wanna wake sleeping dogs with such a move.

image

> how come it's okay to publish this on GitHub, then? I must be missing something here?

it's mirrored on GitHub via git annex links --> the data is all on OSF, if you clone from GitHub all you are getting is the information on where to get the real data. (datalad/git-annex magic)

> We can, it just adds some overhead because it's the only dataset that needs to be fetched via data datalad

perhaps we'll have to find a good EEG dataset alternative then, I totally see the advantage of not doing workarounds just to accommodate a single dataset.

hoechenberger:
> perhaps we'll have to find a good EEG dataset alternative then, I totally see the advantage of not doing workarounds just to accommodate a single dataset.

We could just dump a Zip somewhere and download that using curl / wget? Would this work with OSF?

sappelhoff:
You *can* select the `eeg_matchingpennies` folder on https://osf.io/cj2dr/ and click download as zip --> but a brief investigation didn't show me a way to construct a generic link for that without clicking.

Also, it'd download the whole dataset, which may be undesirable if all you need is just one subject.

image

hoechenberger:
@sappelhoff Looks like it's time for … osf-py! 

https://developer.osf.io/#

😅😳😭😜

PR #227, opened 2020-12-24
Better integration for FreeSurfer script

hoechenberger:
- recon-all script can now be invoked via: `run.py --steps=freesurfer/recon_all`
- We now vendor the FS BIDS app


hoechenberger:
@agramfort might like this…

agramfort:
Cool! Can you just document this before we merge?

hoechenberger:
Absolutely :)

hoechenberger:
@agramfort any suggestions how we should handle the case when the directory for the FS subject already exists? You added a param `overwrite`, but currently there's no way to pass this when calling the script via the Study Template's `run.py` -- should we add a config option? Or simply fail hard and say, hey, just delete the directory if you want to re-create the surface parcellation?

agramfort:
+1 to fail hard. You can always delete the folder manually

>

PR #228, opened 2021-01-05
Add documentation on FS & BEM surface creation

hoechenberger:

PR #229, opened 2021-01-06
Add documentation on FS & BEM surface creation (#228)

hoechenberger:
* Add documentation on FS & BEM surface creation

* Address @agramfort's comments and small doc improvement


hoechenberger:
Hum, somehow VS Codes GH plugin is playing tricks on me.

PR #233, opened 2021-01-11
Better pyvista camera position, single figure with both hemispheres; more type hitns

hoechenberger:


hoechenberger:
Thanks for your help here via personal chat, @GuillaumeFavelier!

PR #236, opened 2021-01-14
Process Faces dataset

hoechenberger:
Remaining / highlighted issues:

- Original dataset doesn't store fine-calibration and crosstalk files according to BIDS, so not doing Maxwell filtering for now

- I'd like to calculate the contrast of "face" vs "scrambled", however, to do that, I'd need to merge the "familiar" and "unfamiliar" events to "faces". We need to add support for operations like this.

Tests will fail until https://github.com/mne-tools/mne-bids/pull/680 has been merged.


hoechenberger:
Whoops this doesn't address #41! I misread the dataset used there… it wasn't ds000117! Hmm… oh well…

agramfort (approved):
Awesome to be able to add a new dataset with so little code!!!! 🔥

PR #237, opened 2021-01-16
missing reject_tmax from config

agramfort:
see https://mne.tools/mne-biomag-group-demo/auto_scripts/library/config.html#sphx-glr-auto-scripts-library-config-py

where it was used @hoechenberger

### DIFF
--- a/config.py
+++ b/config.py
@@ -373,13 +373,13 @@ End time of the interpolation window in seconds.
 # -------------------
 # done in 02-frequency_filter.py
 
-l_freq: float = 1.
+l_freq: Optional[float] = None
 """
 The low-frequency cut-off in the highpass filtering step.
 Keep it None if no highpass filtering should be applied.
 """
 
-h_freq: float = 40.
+h_freq: Optional[float] = 40.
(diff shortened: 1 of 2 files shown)


hoechenberger:
Interesting, I didn't know about this parameter :)

Why did you change highpass fiktering, though? Is it not a good idea to high-pass by default?

agramfort:
@hoechenberger regarding high-pass I think it's a dangerous default to set it to 1Hz. Maybe 0.3 Hz would be a better default but I think it's better to use simple baselines using pre-stim data by default. Of course we should look at rendered reports to see if data say otherwise and if so maybe we set the l_freq != None in the config of those datasets

hoechenberger:
You're right, 1 Hz highpass is too high. Ok let's try with None and see how well it works!

hoechenberger (approved):

agramfort:
looks good https://1766-167957354-gh.circle-artifacts.com/0/site/examples/ds000248.html

hoechenberger:
Nice!

PR #243, opened 2021-01-19
Only create Epochs for events of interest

hoechenberger:
Working with @dnacombo we realized that it may not always be the best
idea to epoch for _all_ events in the data, as we're currently doing. This
PR changes behavior such that we only epoch for conditions the user
specified in the config.

### DIFF
--- a/scripts/preprocessing/03-make_epochs.py
+++ b/scripts/preprocessing/03-make_epochs.py
@@ -63,6 +63,14 @@ def run_epochs(subject, session=None):
         raw = mne.concatenate_raws(raw_list)
 
     events, event_id = mne.events_from_annotations(raw)
+
+    # Only keep the events of interest.
+    event_id = {cond: event_id[cond] for cond in config.conditions}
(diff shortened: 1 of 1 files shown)


scripts/preprocessing/03-make_epochs.py
```diff
@@ -63,6 +63,14 @@ def run_epochs(subject, session=None):
         raw = mne.concatenate_raws(raw_list)
 
     events, event_id = mne.events_from_annotations(raw)
+
+    # Only keep the events of interest.
+    event_id = {cond: event_id[cond] for cond in config.conditions}
+    keep_events_idx = [idx for idx, event in enumerate(events)
+                       if event[-1] in event_id.values()]
+    events = events[keep_events_idx]
```
agramfort, line 71: this will not work.

you may specify condition faces and the event_id contains faces/first faces/second

hoechenberger:
Thanks @agramfort, @dnacombo, I'll have to rethink this.

PR #244, opened 2021-01-19
Make tables more beautiful

hoechenberger:


hoechenberger:
@agramfort Reports come back nicely, see for example

https://1817-167957354-gh.circle-artifacts.com/0/reports/ds000117/sub-average_ses-meg_task-facerecognition_report.html

Merge if you're happy!!

agramfort:
thx @hoechenberger

PR #245, opened 2021-01-20
Adjustments for ERP-CORE processing

hoechenberger:


config.py
```diff
@@ -213,6 +213,61 @@
     ```
 """
 
+Bipolar_T = Tuple[str, str, str]
```
agramfort, line 216: I don't like parameters which is CamelCase and snake_case.

I would just do:

`eeg_bipolar_channels: Iterable[Tuple[str, str, str]] = []`

hoechenberger: > I don't like parameters which is CamelCase and snake_case.

Heh, TypeScript habit. Will change.

PR #260, opened 2021-03-08
Various changes for metadata & ERP CORE processing

hoechenberger:
This PR adds support for metadata creation, applies baseline correction only after ICA (not before), and slightly improves ICA reports and some log messages.


hoechenberger:
re 27fed7b, see https://github.com/mne-tools/mne-python/issues/9021

‼️**Do not merge without revisiting this** ‼️

PR #267, opened 2021-03-26
MAINT: Refactor CircleCI

larsoner:
Let's see if testing ds003392 works. If it does, I can make the rest of the datasets work the same way. So I'm expecting / hoping `test_ds003392` to work first (and `build_docs` will fail), then I'll try to get `build_docs` working, then probably the rest of the tests.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -50,7 +50,7 @@ jobs:
       - run:
           name: Get conda running
           command: |
-            if [ ! -d "~/miniforge" ]; then
+            if [ -d ~/miniforge ]; then
               wget https://github.com/conda-forge/miniforge/releases/latest/download/Miniforge3-Linux-x86_64.sh -O ~/miniforge.sh;
               chmod +x ~/miniforge.sh;
               ~/miniforge.sh -b -p ~/miniforge;
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -413,12 +413,12 @@ jobs:
             npm install -g --silent gh-pages@3.0.0  # Work around https://github.com/tschaub/gh-pages/issues/354
             git config user.email "ci-build@mne.tools"
             git config user.name "ci-build"
-      - run:
-          name: Deploy docs to gh-pages branch
-          command: gh-pages --message "[skip ci] Update docs" --dist docs/site
       - add_ssh_keys:
           fingerprints:
             - "da:49:10:cd:fd:7d:2f:62:7c:78:be:4a:e1:c4:72:6c"
+      - run:
+          name: Deploy docs to gh-pages branch
+          command: gh-pages --message "[skip ci] Update docs" --dist docs/site
 
 workflows:
   version: 2
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -67,6 +67,27 @@ jobs:
           key: conda-cache
           paths:
             - "~/miniforge"
+      # Look at what we have and fail early if there is some library conflict
+      - run:
+          name: Check installation
+          command: |
+              which python
+              git-annex version
+              datalad --version
+              openneuro-py --version
+              mne_bids --version
+              mne sys_info
+      - persist_to_workspace:
+          root: ~/
+          paths:
+            - project
+            - miniforge
+            - reports
+            - mne_data
+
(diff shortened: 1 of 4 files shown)


larsoner:
So far the dependencies are nicely laid out bi CircleCI, which is nice!

Screenshot from 2021-03-26 11-21-39

larsoner:
Okay this architecture is more or less what I had in mind, let's see if we can get things green and if it actually brings some benefits in terms of timing:

Screenshot from 2021-03-26 12-36-54

larsoner:
@hoechenberger does the doc build actually depend on the tests generating the reports? If so I think this updated dependency graph is hopefully correct:

Screenshot from 2021-03-26 14-02-14

agramfort:
+1 for MRG when CI are happy

thing is really awesome !

larsoner:
Cuts build time from ~1 hour to ~35 minutes (at least sometimes), which should get even shorter once everything is cached on `main`. I think it's possible to squeeze maybe 5 minutes by moving to `pip` from `conda`. Caching the conda took minutes (it's > 2 GB after everything is installed) so it's faster just to install everything, but even still the ~2GB has to be persisted to the workspace (which takes a few minutes). By contrast our MNE-Python CircleCI pip-based build takes ~1 minute to load pip from cache. But this can be done in another PR.

I'm going to push a commit to get rid of a conditional `RUNNING_ON_CIRCLE` that I think shouldn't be necessary. Then @hoechenberger I might need your help with [this error](https://app.circleci.com/pipelines/github/mne-tools/mne-bids-pipeline/1475/workflows/94c98a2b-0ea9-447f-8b30-33de9226c62e/jobs/2220) because I'm not familiar with this doc build architecture:
```
./docs/build-docs.sh
20:12:38 Using custom configuration: tests/configs/config_ds000117.py
20:12:38 Using custom configuration: tests/configs/config_ds000246.py
20:12:38 Using custom configuration: tests/configs/config_ds000248.py
  File "docs/source/examples/gen_examples.py", line 143, in <module>
    demonstrated_funcs_str = gen_demonstrated_funcs_str(config_path)
  File "docs/source/examples/gen_examples.py", line 34, in gen_demonstrated_funcs_str
    config = runpy.run_path(config_module_path)
  File "/home/circleci/miniforge/lib/python3.8/runpy.py", line 265, in run_path
    return _run_module_code(code, init_globals, run_name,
  File "/home/circleci/miniforge/lib/python3.8/runpy.py", line 97, in _run_module_code
    _run_code(code, mod_globals, init_globals,
  File "/home/circleci/miniforge/lib/python3.8/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "config.py", line 1333, in <module>
    raise ValueError(msg)
ValueError: You set mf_reference_run=01, but your dataset only contains the following runs: [None]
```
It suggests that there are some more files that I need to make persist after the `test` runs beyond just the derivative HTMLs/JSONs/TSVs.

.circleci/config.yml
```diff
@@ -1,266 +1,604 @@
...
+          <<: *bashenv
+      - run:
+          <<: *xvfb
+      - attach_workspace:
+          at: ~/
+      # necessary because gen_examples.py runs config script, which otherwise complains:
+      # ValueError: You set mf_reference_run=01, but your dataset only contains the following runs: [None]
+      - restore_cache:
+          keys:
+            - data-cache-ds000248
```
larsoner, line 485: I think this should fix the doc build error. If it does, in principle we could hack in some way *not* to sanity check the `config.py` for this run to get around this at some point, then we don't need to restore the data cache for the doc build, and it can just rely on the HTMLs/reports living in the correct place (which they should based on the `persist_to_workspace` calls made now).

larsoner:
All better (I'm assuming the T1 BEM will finish eventually):

https://2242-167957354-gh.circle-artifacts.com/0/site/index.html

Screenshot from 2021-03-26 17-38-37

hoechenberger:
This is amazing wizardry!!! 🧙‍♂️

hoechenberger:
Thanks a million, @larsoner !

larsoner:
Okay to move to `pip` for the CircleCI build to speed it up?

agramfort:
No objection

PR #270, opened 2021-03-29
MAINT: Move to pip

larsoner:
Let's see if this is faster


hoechenberger:
Nice! Seems you cut down the setup duration from ~8.5 to ~2.5 minutes

hoechenberger:
@larsoner Is this good to merge?

larsoner:
I don't love that the keyserver I used to install a recent enough `datalad` timed out a few times during testing. But we can give it a shot and figure out a workaround if it fails, so +1 for merge from me

hoechenberger:
Thanks @larsoner!

PR #271, opened 2021-03-29
Add ERP CORE examples

hoechenberger:


hoechenberger:
A non-ERP CORE test is failing now, currently trying to get this resolved. Otherwise we're good I believe
@agramfort WDYT, should I add another section just for the grand average reports?
https://2587-167957354-gh.circle-artifacts.com/0/site/examples/ERP_CORE.html

agramfort (approved):

PR #274, opened 2021-04-01
Spread the union of all bad channels across all runs before concatenating

hoechenberger:
Otherwise, concatenation will fail.

Maybe we should add an option upstream in MNE to do this automatically? @agramfort WDYT?

### DIFF
--- a/scripts/preprocessing/03-make_epochs.py
+++ b/scripts/preprocessing/03-make_epochs.py
@@ -26,7 +26,7 @@ logger = logging.getLogger('mne-bids-pipeline')
 @failsafe_run(on_error=on_error)
 def run_epochs(subject, session=None):
     """Extract epochs for one subject."""
-    raw_list = list()
+    raws = list()
     bids_path = BIDSPath(subject=subject,
                          session=session,
                          task=config.get_task(),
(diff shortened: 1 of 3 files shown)


SophieHerbst:
I think we should explicitly discuss the strategy for dealing with bad channels: if we allow to set bad channels per run, one will not expect them to be spread across all runs. 
Wouldn't it make more sense to interpolate the bad channels in the preprocessing, so they do not appear as bad anymore, but after preprocessing, all participants have the same number of channels? 
This is already the case when Maxfilter is used.

hoechenberger:
@SophieHerbst 
> Wouldn't it make more sense to interpolate the bad channels in the preprocessing, so they do not appear as bad anymore, and importantly, after preprocessing, all participants have the same number of channels?

We actually have a separate setting for this when we do grand averaging:
https://github.com/mne-tools/mne-bids-pipeline/blob/6d1a0ffdba9396cbf34db322d95f6f1b190e34c0/config.py#L917-L928

Not sure what the best strategy is, but very happy to discuss!

SophieHerbst:
@hoechenberger to me it seems best to get rid of them once and for all before creating epochs, so we don't have to deal with them every time we average (or extract the data as a matrix, which might be less common, but is something I have done and seen in the past). 
@agramfort, what is your opinion?

hoechenberger:
> to me it seems best to get rid of them once and for all before creating epochs

I like this, we just have to ensure to show in the report clearly WHICH channels were initially bad / later interpolated.

This would also make processing more similar between Maxwell-filtered MEG data and EEG systems. And I like consistency :)

agramfort:
I see different scenario

- one channel is bad for all runs (I don't see a reason to interpolate it)
at least not before grand_average
- one channel is bad for only a fraction of the runs : option 1)  drop all
the time OR 2) interpolate

note that for MEG maxfilter will interpolate systematically

thoughts?

>

hoechenberger:
> note that for MEG maxfilter will interpolate systematically

I'm not sure what you mean by that? My understanding is that for each run, it will interpolate bad channels, no matter what. Is that what you mean?

hoechenberger:
> - one channel is bad for all runs (I don't see a reason to interpolate it)

How would this affect source estimation? If we essentially have a "hole" in the lead field? Or is there not really a difference in the outcome, because all information an interpolated channel contributes is already fully contained in the spatially close channels?

SophieHerbst:
@agramfort 
> - one channel is bad for all runs (I don't see a reason to interpolate it)

The reason would be to obtain the same data matrix for all participants, no matter whether they had bad channels or not, and to not have to deal with this every time we average or do some group analyses. 
I assume that the MNE functions automatically disregard bad sensors, but what happens when you use epochs.get_data() for instance?

hoechenberger:
Thanks @SophieHerbst! @agramfort and I just had a call and we're escalating this to upstream: https://github.com/mne-tools/mne-python/issues/9231

hoechenberger:
Putting this on hold for now!

hoechenberger:
We now have https://github.com/mne-tools/mne-python/pull/9241 which should give us greater flexibility!

PR #275, opened 2021-04-01
Add config option to handle repeated events

hoechenberger:


agramfort (approved):
I can live with that

hoechenberger:
I hope this will only be an intermediate solution until we have better handling in MNE-Python and MNE-BIDS

hoechenberger:
Putting this on hold

hoechenberger:
Wait, that was the wrong one. This is good to merge ;)

PR #281, opened 2021-04-09
Ensure bipolar EOG channels have correct type

hoechenberger:
Also make ERP CORE config more explicit regarding
EOG specification


scripts/preprocessing/01-import_and_maxfilter.py
```diff
@@ -261,6 +261,26 @@ def load_data(bids_path):
                                       ch_name=ch_name, drop_refs=False,
                                       copy=False)
 
+        # If we created a new bipolar channel that the user wishes to
+        # use as an EOG channel, it is probabluy a good idea to set its channel
```
agramfort, line 265: probabluy -> probably

PR #296, opened 2021-04-15
WIP: ICA clean raw data

MerlinDumeur:
This code works for one dataset, on one subject.
I will do the tests soon when I get the time.

I open this PR for feedback.
Some ironing out will need to be done, this is very much still a work in progress.

### DIFF
--- a/config.py
+++ b/config.py
@@ -11,11 +11,12 @@ import traceback
 import sys
 import copy
 import logging
-from typing import Optional, Union, Iterable, List, Tuple, Dict
 if sys.version_info >= (3, 8):
-    from typing import Literal
+    from typing import (Optional, Union, Iterable, List, Tuple, Dict, Literal,
+                        Callable)
 else:
-    from typing_extensions import Literal
+    from typing_extensions import (Optional, Union, Iterable, List, Tuple,
+                                   Dict, Literal, Callable)
 
 import coloredlogs
 import numpy as np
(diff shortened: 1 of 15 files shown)


agramfort:
I would suggest to remove all auto reject related code here and to focus on the rest. We can add auto reject in a next PR

MerlinDumeur:
If this is because the checks fail, I think it would be an easy fix, but I agree it's probably best to keep autoreject separate for now.
It work fine for me though, so it shouldn't be too hard to add later.

config.py
```diff
@@ -569,7 +569,7 @@
 # AUTOMATIC REJECTION OF ARTIFACTS
 # --------------------------------
 
-reject: Optional[dict] = None
+reject: Optional[Union[dict, Literal['auto']]] = None
```
hoechenberger, line 572: This needs to be reverted too

config.py
```diff
@@ -579,17 +579,17 @@
...
 
-???+ note "Note"
-    These numbers tend to vary between subjects.. You might want to consider
-    using the autoreject method by Jas et al. 2018.
-    See https://autoreject.github.io
-
+Pass ``'auto'`` if you want to automate the estimation of the reject
+parameter using AutoReject [Jas et al. 2017] (See https://autoreject.github.io).
+AutoReject is useful as the optimal rejection thresholds tend to vary between
+subjects.
```
hoechenberger, line 585: same

config.py
```diff
@@ -579,17 +579,17 @@
...
 
 ???+ example "Example"
     ```python
     reject = {'grad': 4000e-13, 'mag': 4e-12, 'eog': 150e-6}
     reject = {'grad': 4000e-13, 'mag': 4e-12, 'eeg': 200e-6}
     reject = None
+    reject = 'auto'
```
hoechenberger, line 592: ditto

scripts/preprocessing/04a-run_ica.py
```diff
@@ -105,15 +105,20 @@ def make_epochs_for_ica(raw, subject, session):
                             root=config.deriv_root,
                             check=False)
     epochs = mne.read_epochs(epochs_fname)
-    selection = epochs.selection
+
+    # Not sure why it is necessary to recreate epochs when we already have them
+    if config.no_epoching:
```
MerlinDumeur, line 110: The base code recreates an epoch object, but I don't understand why. Here I just reuse the epochs created last step, not sure if that's okay

hoechenberger: > but I don't understand why.

Because we filter the raw data again with a high-pass filter suitable for ICA. We then create new epochs from the newly-filtered data. You'd have to do the same for the "no epoching" case

MerlinDumeur: I completely missed that, thanks

scripts/preprocessing/05a-apply_ica.py
```diff
@@ -76,41 +93,69 @@ def apply_ica(subject, session):
...
-    epochs_cleaned = ica.apply(epochs.copy())  # Copy b/c works in-place!
-    epochs_cleaned.apply_baseline(config.baseline)
+    if config.no_epoching:
 
-    msg = 'Saving cleaned epochs.'
-    logger.info(gen_log_message(message=msg, step=5, subject=subject,
-                                session=session))
-    epochs_cleaned.save(fname_epo_out, overwrite=True)
+        msg = f'Rejecting ICs: {ica.exclude}'
+        logger.info(gen_log_message(message=msg, step=5, subject=subject,
```
hoechenberger, line 99: This should be logged for both cases, i.e. should be outside the `if` block

MerlinDumeur: yep that's what I intended to do, will fix

run.py
```diff
@@ -184,7 +184,9 @@ def process(config: PathLike,
         script_paths = [*SCRIPT_PATHS['init'], *script_paths]
 
     for script_path in script_paths:
-        step_name = script_path.name.replace('.py', '')[3:]
```
MerlinDumeur, line 187: Changed because it did not work with file names such as `recon_all.py` or `04a-run_ica.py`

MerlinDumeur:
I looked at how mne handles the computation of the source PSD for raw data (`mne.minimum_norm.compute_source_psd`), and it appears that it just creates an Epoch instance with only 1 epoch which is the length of the raw recording, and calls the `mne.minimum_norm.compute_source_psd_epochs` function.

We could maybe take inspiration from this? I could adapt the PR

agramfort:
as soon as you have an epoch you're blocked to ignore/remove transient artifacts?

MerlinDumeur:
yes that's true but that will be always be a problem anyway. If I can't epoch my data because I need long recordings, then there's not much I can about it either way, unless I missed something.

It would be nice to have mne functions on raw data that ignore transient artifacts based on annotations, but I don't think that's available yet? For instance if I want the compute the PSD on raw data, it does not take into account BAD annotations

larsoner:
As of #840 ICA (and SSP) are now also applied to raw data and saved to disk as `proc-clean`

PR #301, opened 2021-04-21
Fix recon_all script

### LINKED ISSUE #295: `ERROR: Could not consume arg: freesurfer` when finishing `run.py freesurfer`
Here is the output from running `run.py freesurfer`:
```
2021-04-15 17:02:00 INFO Now running: init_derivatives_dir
17:02:01 Using custom configuration: /home/merlin/hMT+/config.py
17:02:01 [Step-01] Running: Initializing output directories.
17:02:01 [Step-01] Initializing output directories.
17:02:01 [Step-01] Completed: Initializing output directories.
2021-04-15 17:02:01 INFO Successfully finished running: init_derivatives_dir
2021-04-15 17:02:01 INFO Now running: recon_all
17:02:01 Running FreeSurfer
17:02:01 Running recon-all on subject 01. This will take a LONG time – it's a good idea to let it run over night.
17:02:01 Subject 01 is already present. Please delete the directory if you want to recompute.
ERROR: Could not consume arg: freesurfer
Usage: recon_all.py freesurfer

```
Here I only ask to process subject 01, which was already done.
Even though there was no error within `recon_all.py` (which I checked using a print at the end of `main`), the script exits with an error.
The problem seems to be coming from using both Fire and runpy: https://github.com/mne-tools/mne-bids-pipeline/blob/d49d05351b2da2a4269dff5761a457e0e85964bd/run.py#L98
it seems Fire expects all arguments to be used for both `run.py` and `recon_all.py`, when they are only used for `run.py`.

Note that this is not a regression from #289, as it happens even without that modification.

I'm not sure how fix this, maybe it is simple though as it does not seem to be a problem for other scripts.

hoechenberger:
Turns out that the use of Fire in `recon_all.py`, combined with Fire in `run.py`, created some weird kind of chaining that ended up passing arguments to `recon_all.py` that were supposed to be consumed by `run.py` only.

Since `recon_all.py` has already been somewhat tightly integrated with `run.py` / the BIDS Pipeline, I now simply removed Fire from it, and we get all info we need to operate from the user's Pipeline config and environment variables.

Fixes #295

cc @MerlinDemeur

### DIFF
--- a/scripts/freesurfer/recon_all.py
+++ b/scripts/freesurfer/recon_all.py
@@ -7,8 +7,6 @@ from pathlib import Path
 import logging
 from typing import Union
 
-import fire
-
 from mne.utils import run_subprocess
 from mne.parallel import parallel_func
 
@@ -60,10 +58,10 @@ def run_recon(root_dir, subject, fs_bids_app) -> None:
     run_subprocess(cmd, env=env, verbose=logger.level)
 
 
-def main(*, n_jobs: int = 1) -> None:
+def main() -> None:
     """Run freesurfer recon-all command on BIDS dataset.
 
-    The command allows to run the freesurfer recon-all
+    The script allows to run the freesurfer recon-all
     command on all subjects of your BIDS dataset. It can
     run in parallel with the --n_jobs parameter.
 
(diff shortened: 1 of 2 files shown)


scripts/freesurfer/recon_all.py
```diff
@@ -109,4 +107,4 @@ def main(*, n_jobs: int = 1) -> None:
 
 
 if __name__ == '__main__':
-    fire.Fire(main)
+    main()
```
hoechenberger, line 110: Actually I think this can be removed now – the "proper" way to run this step is through `run.py`

agramfort (approved):

hoechenberger:
@MerlinDumeur Could you please check if it works for you too? :)

scripts/freesurfer/recon_all.py
hoechenberger: Actually I was mistaken here, `run.py` currently runs all processing steps as scripts, so we do need this bit

MerlinDumeur:
Works fine for me!
thanks

Issue #314, opened 2021-04-28
Cannot use both SSP and ICA


guiomar:
Hi!

When marking `use_uca=true` I get this error: 

> Cannot use both SSP and ICA.

The `use_ssp` field is not declared in the config file (as the default is false)

Could I be doing something wrong? Would you mind trying to reproduce?

agramfort:
can you replicate the issue with some public data so we can easily
replicate?
if so please share also your config.py file.

otherwise share a dropbox folder with me please

guiomar:
Yes, I used ds000248, wit this config file:

> 
> bids_root = '../6089740e02bc8861a590b94b/bids/.'
> deriv_root = 'out_dir'
> subjects = ['01']
> runs = ['01']
> find_flat_channels_meg = False
> find_noisy_channels_meg = False
> use_maxwell_filter = False
> fix_stim_artifact = False
> h_freq = 40
> decim = 10
> on_rename_missing_events = 'raise'
> event_repeated = 'error'
> conditions = ['Auditory', 'Visual', 'Auditory/Left', 'Auditory/Right'] 
> contrasts = [('Visual', 'Auditory'),('Auditory/Right', 'Auditory/Left')]
> use_ica = True
> ica_algorithm = 'picard'
> ch_types = ['meg']

guiomar:
oh! I realized the use_ssp if True by default! SORRY

guiomar:
I really apologize for all these absurd questions. I'm getting used now to python and mne tool (which I'm finding awesome). Thanks for your patience!

agramfort:
no problem

agramfort:
don't worry :)

hoechenberger:
@agramfort I’m starting to think we should set both, ICA and SSP, to False by default. WDYT?

agramfort:
Yes agreed

guiomar:
Yes, I also think it might be more intuitive, since all the other booleans are false by default

PR #315, opened 2021-04-29
No SSP by default; remove "use_{ica,ssp}" settings in favor of new setting "spatial_filter"

hoechenberger:
### Before merging …

@guiomar @agramfort WDYT?

x-ref: #314

### DIFF
--- a/config.py
+++ b/config.py
@@ -1319,13 +1319,15 @@ if (use_maxwell_filter and
         len(set(ch_types).intersection(('meg', 'grad', 'mag'))) == 0):
     raise ValueError('Cannot use maxwell filter without MEG channels.')
 
-if use_ica and ica_algorithm not in ('picard', 'fastica', 'extended_infomax'):
+if (spatial_filter == 'ica' and
+        ica_algorithm not in ('picard', 'fastica', 'extended_infomax')):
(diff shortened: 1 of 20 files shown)


config.py
```diff
@@ -762,22 +762,21 @@
...
 
-# ICA
-# ~~~
+Use `'ssp'` for SSP, `'ica'` for ICA, and `None` if you do not wish to apply
+a spatial filter for artifact removal.
 
-use_ica: bool = False
-"""
-Whether independent component analysis should be used or not.
+The Pipeline will try to automatically discover EOG and ECG artifacts. ForSSP,
```
agramfort, line 774:
```suggestion
The Pipeline will try to automatically discover EOG and ECG artifacts. For SSP,
```

hoechenberger, line 774:
```suggestion
The Pipeline will try to automatically discover EOG and ECG artifacts. For SSP,
```

guiomar:
Thanks! I was going to suggest to make it optional too and have either: ssp, ica or none. Since if you mark both as true it gives you error, so it's more convenient from people to be prevented with this optional choice.
I was only wandering if there would be cases were people do ssp and then some ica to remove some additional component that couldn't be cleaned with the ssp. But I guess this is probably not going to be that common to be included in the general pipeline

PR #316, opened 2021-04-29
Add getter functions for {bids,deriv}_root

hoechenberger:
### Before merging …

This ensures that the documentation for both parameters shows up correctly in the rendered output. In `main`, that's not the case, because we re-defined both variables in `config.py` after their first introduction with a corresponding docstring. The re-definition, however, nixes the docstring. The getter functions solve this issue.

No changelog entry needed as only a documentation change to users.

### DIFF
--- a/config.py
+++ b/config.py
@@ -1289,28 +1289,6 @@ if "MNE_BIDS_STUDY_CONFIG" in os.environ:
             exec("%s = custom_cfg.%s" % (val, val))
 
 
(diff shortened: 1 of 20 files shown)


agramfort (approved):

hoechenberger:
`bids_root` and `deriv_root` are now properly documented:
https://4222-167957354-gh.circle-artifacts.com/0/site/settings/general.html#config.bids_root

PR #322, opened 2021-05-07
Use absolute paths to in more places

hoechenberger:
### Before merging …

### DIFF
--- a/config.py
+++ b/config.py
@@ -1384,8 +1384,11 @@ if bem_mri_images not in ('FLASH', 'T1', 'auto'):
 def get_bids_root() -> pathlib.Path:
     # BIDS_ROOT environment variable takes precedence over any configuration file
     # values.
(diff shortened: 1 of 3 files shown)


config.py
```diff
@@ -1395,14 +1398,18 @@ def get_bids_root() -> pathlib.Path:
...
 
 
 def get_deriv_root() -> pathlib.Path:
     if deriv_root is None:
         return get_bids_root() / 'derivatives' / PIPELINE_NAME
     else:
-        return pathlib.Path(deriv_root).expanduser()
+        return (pathlib.Path(deriv_root)
+                .expanduser()
+                .resolve())
```
hoechenberger, line 1412: no `strict` here because it's okay if this directory doesn't exist, as we'll then just create it

agramfort (approved):

PR #327, opened 2021-05-10
WIP, ENH: bypass script 01

### LINKED ISSUE #321: ENH/API: script 01 obsolete for EEG
I am wondering if we can skip script 1; in my use case it simply copies/duplicates the EEG files (which I already duplicated by converting to bids ...). 

cc @hoechenberger @agramfort @sappelhoff

dengemann:
Closes #321.

### Before merging …


agramfort (commented):
let's see if it breaks any of the CIs.

dengemann:
@hoechenberger @agramfort any ideas about the failing tests?

hoechenberger:
Looking into this now

scripts/preprocessing/01-import_and_maxfilter.py
```diff
@@ -466,10 +466,12 @@ def main():
     msg = 'Running Step 1: Data import'
```
hoechenberger, line 466:
```suggestion
    msg = 'Running Step 1: Maxwell filter
```

scripts/preprocessing/01-import_and_maxfilter.py
```diff
@@ -466,10 +466,12 @@ def main():
...
-    parallel(run_func(subject, session) for subject, session in
-             itertools.product(config.get_subjects(), config.get_sessions()))
+    if config.use_maxwell_filter:
+        parallel, run_func, _ = parallel_func(run_maxwell_filter,
+                                              n_jobs=config.N_JOBS)
+        parallel(run_func(subject, session) for subject, session in
+                 itertools.product(config.get_subjects(),
+                                   config.get_sessions()))
 
     msg = 'Completed Step 1: Data import'
```
hoechenberger, line 476:
```suggestion
    msg = 'Completed Step 1: Maxwell filter
```

scripts/preprocessing/02-frequency_filter.py
```diff
@@ -72,7 +72,15 @@ def run_filter(subject, run=None, session=None):
     logger.info(gen_log_message(message=msg, step=2, subject=subject,
                                 session=session, run=run,))
 
-    raw = mne.io.read_raw_fif(raw_fname_in)
+    read_raw = mne.io.read_raw_fif
```
hoechenberger, line 75: Hm, we do this nowhere else in the pipeline, do you feel strongly about keeping this?

hoechenberger (commented):
@dengemann I believe I found the issue, will push an attempted fix shortly.

hoechenberger:
Fixed the issue with the "missing" events, but now there's something else showing up. Digging into this…

hoechenberger:
Well. This blew up.

dengemann:
@hoechenberger I think we're moving away from the purpose of the pull request which was to skip any script that runs before the 02 frequency filtering.

hoechenberger:
We're skipping unless the user wants to add or remove channels or rename events

Maybe we should have a call, as this thing is much more complex than at first anticipated. We're doing many more things in the import step than just copying the data.

dengemann:
> Maybe we should have a call, as this thing is much more complex than at first anticipated. We're doing many more things in the import step than just copying the data.

Yes, but even these things are not mandatory for all datasets and situations, no?
Sure let's have a call.

agramfort:
discusses with @hoechenberger 

the plan is to add an `import_raw` function in config.py and use this function in the first script we need.

hoechenberger:
Will make a PR later today!

hoechenberger:
Ok I think I've decluttered most of the functionality and refactored it into numerous small functions. Now it's time to re-assemble things and eventually get rid of script 00 (import)

PR #331, opened 2021-05-18
no proj ON in unclean epochs

agramfort:
cc @hoechenberger

### DIFF
--- a/scripts/preprocessing/03-make_epochs.py
+++ b/scripts/preprocessing/03-make_epochs.py
@@ -94,7 +94,7 @@ def run_epochs(subject, session=None):
                                 session=session))
     epochs = mne.Epochs(raw, events=events, event_id=event_id,
                         tmin=config.epochs_tmin, tmax=config.epochs_tmax,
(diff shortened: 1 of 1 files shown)


hoechenberger:
Do we need to explicitly apply the proj at some later time point during processing?

agramfort:
it's gets applied by read_epochs by default... don't ask me why...

>

hoechenberger:
We have `proj=True` when creating `Epochs` in script 4 too, I believe this need to be changed as well..?

> it's gets applied by read_epochs by default

Does this mean that when reading the epochs in step 5 (where we apply ICA) we should pass `proj=False` to `read_epochs()`?

agramfort:
I think it's ok to apply proj before doing SSP or ICA.

I want to have epochs without proj to be able to spot bad channels more
easily

>

hoechenberger:
Ok, sounds fair to me. I've restarted the one failing CI

hoechenberger (approved):

hoechenberger:
Thanks @agramfort!

PR #334, opened 2021-05-20
Remove by default `conditions` values in config.py

guiomar:
Maybe we could remove the by default `conditions` names, without knowing the real event names in the data it's a bit risky to set them at random?

### Before merging …

### DIFF
--- a/config.py
+++ b/config.py
@@ -666,7 +666,7 @@ occurrence of matching event types. The columns indicating the event types
 will be named with a ``last_`` instead of a ``first_`` prefix.
 """
 
-conditions: Union[Iterable[str], Dict[str, str]] = ['left', 'right']
+conditions: Union[Iterable[str], Dict[str, str]] = []
 """
 The time-locked events based on which to create evoked responses.
 This can either be name of the experimental condition as specified in the


agramfort (commented):
good point !

these are bad default.

+1 for MRG if CIs are green.

thx

hoechenberger:
Thanks @guiomar! What would the intended default behavior be? Using all events? Should we maybe use `conditions = None` to represent this case?

hoechenberger:
@larsoner the `setup_env` job is failing quite frequently when trying to add the signing key; any idea if we could, well, improve this behavior somehow? I could imagine simply retrying a few times instead of failing hard immediately?

agramfort:
I think it's hard to guess the conditions.

could we make some parameters required ?

hoechenberger:
> I think it's hard to guess the conditions.
> 
> could we make some parameters required ?

I'm not entirely sure what you mean. So we agree that the default is not great, right? But do we want to force users to specify something here? (I'd be +1 for this, but I'm not sure this is what @guiomar had in mind)

agramfort:
we could make it None and have the script crash with a message if None
saying that it must be changed?

hoechenberger:
> we could make it None and have the script crash with a message if None
> saying that it must be changed?

Yes, sounds good to me!

hoechenberger:
@guiomar Want to give this a shot? Otherwise I can take over

guiomar:
Yes! I'm trying to figure out which config parameters are strictly necessary for each of the pipeline steps (preprocessing, sensors, sources). We could leave them None and raise an error if missing or left empty. Which I think it's how is done now in general, but maybe not specifically for each of the steps separately.
We can talk about it in our meeting now :) Thanks!

hoechenberger:
@guiomar and I just had a call. We will make the default `None` and raise if it's not changed

guiomar:
Thanks @hoechenberger !! Can I remove this PR an make a new one?

hoechenberger:
> Can I remove this PR an make a new one?

Yes whatever you prefer!

PR #342, opened 2021-05-22
Clear ds003392 cache

hoechenberger:
We need to clear the cache, otherwise the download will continue to fail

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -194,7 +194,7 @@ jobs:
           at: ~/
       - restore_cache:
           keys:
-            - data-cache-ds003392
+            - data-cache-ds003392-new
       - run:
           <<: *bashenv
       - run:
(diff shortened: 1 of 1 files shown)


hoechenberger:
The issue is related to the fact that the dataset has been updated upstream, yet our cache contained the old version.

openneuro-py detected: oh, remote file `CHANGES` is larger than the local one, so let's resume this interrupted download! Which effectively lead to a corrupted file, because `CHANGES` upstream was **prepended**, no appended, yet "resuming" a download appends to the local file.

The solution is to make openneuro-py check if the upstream revision of a dataset actually matches the local revision before trying to resume a download. x-ref https://github.com/hoechenberger/openneuro-py/issues/45

hoechenberger:
This also highlights long-standing problems with the OpenNeuro API:

- we don't get file sizes for all files
- we don't get hashes

Otherwise we could just cache the hashes to avoid such issues.

agramfort (approved):

agramfort:
Thx 🙏

PR #353, opened 2021-05-28
different runs for different subjects

### LINKED ISSUE #352: runs = "all" assumes that all participants have the same runs
The runs = "all" parameter makes the assumption that the runs are identical for each subject. 

(I think the function `get_entity_vals(get_bids_root(), entity_key='run')` in the function `get_runs()` in the config.py file reunites all the run names that exist for all the subjects)

for example, if subject 403 has only 8 runs when subject 512 has 9 runs, we then get errors that look like this when dealing with subject 403:

> 2021-05-27 15:31:01 INFO Successfully finished running: init_derivatives_dir
> 2021-05-27 15:31:01 INFO Now running: import_and_maxfilter
> 15:31:02 [Step-01] Running Step 1: Data import
> 15:31:02 [Step-01][sub-403] Loading reference run: 01.
> 15:31:04 [Step-01][sub-403] Applying Maxwell filter to experimental data.
> 15:31:10 [Step-01][sub-403] Processing empty-room recording …
> 15:31:11 [Step-01][sub-403] Applying Maxwell filter to empty-room recording
> 15:31:14 [Step-01][sub-403] Applying Maxwell filter to experimental data.
> 15:31:24 [Step-01][sub-403] Applying Maxwell filter to experimental data.
> 15:31:33 [Step-01][sub-403] Applying Maxwell filter to experimental data.
> 15:31:44 [Step-01][sub-403] Applying Maxwell filter to experimental data.
> 15:31:53 [Step-01][sub-403] Applying Maxwell filter to experimental data.
> 15:32:02 [Step-01][sub-403] Applying Maxwell filter to experimental data.
> 15:32:13 [Step-01][sub-403] Applying Maxwell filter to experimental data.
> 15:32:21  A critical error occurred.

It would be good to add a warning immediately in the config.py if the participants do not have the same number of runs and do not have the same list of run names. What do you think ?

crsegerie:
Closes  #352.

During the process, I've moved up the get_subjects() function without changing it.

### Before merging …

### DIFF
--- a/config.py
+++ b/config.py
@@ -1552,10 +1552,11 @@ def get_runs(subject: str, verbose: bool = False) -> Union[List[str], List[None]
 # XXX This check should actually go into the CHECKS section, but it depends
 # XXX on get_runs(), which is defined after that section.
 inter_runs = get_intersect_run()
-if ('MKDOCS' not in os.environ and 
-    mf_reference_run is not None and
-    mf_reference_run not in inter_runs
-    ):
+mf_ref_error = (
+    ('MKDOCS' not in os.environ) and
+    (mf_reference_run is not None) and
+    (mf_reference_run not in inter_runs))
+if mf_ref_error:
     msg = (f'You set mf_reference_run={mf_reference_run}, but your dataset '
            f'only contains the following runs: {inter_runs}')
     raise ValueError(msg)
--- a/config.py
+++ b/config.py
@@ -1501,7 +1501,11 @@ def get_runs(subject: str, verbose: bool = False) -> list:
         # Notify if different subjects do not share the same runs
 
         # set() to remove duplicates and check for values count
-        same_runs = len(list(set(list(subj_runs.values())))) == 1
+        same_runs = True
+        for runs_sub_i in subj_runs.values():
+            if set(runs_sub_i) != set(list(subj_runs.values())[0]):
+                same_runs = False
+
         if not same_runs:
             msg = ("Extracted all the runs. "
                    "Beware, not all subjects share the same "
--- a/config.py
+++ b/config.py
@@ -1512,7 +1512,8 @@ def get_runs_all_subjects() -> dict:
     -------
     a dict of runs present in the bids_path
     for each subject asked in the configuration file
-    (and not for each subject present in the bids_path)."""
+    (and not for each subject present in the bids_path).
+    """
     # We cannot use get_subjects() because if there is just one subject
     valid_subs = get_entity_vals(get_bids_root(), entity_key='subject')
 
--- a/config.py
+++ b/config.py
@@ -1495,6 +1495,7 @@ def get_runs(subject: str, verbose: bool = False) -> list:
         Returns a list of the runs of this subject.
     verbose
         Notify if different subjects do not share the same runs.
+
     Returns
     -------
     list of string
--- a/config.py
+++ b/config.py
@@ -1412,58 +1412,6 @@ def get_deriv_root() -> pathlib.Path:
                 .resolve())
 
 
-def get_sessions():
-    sessions_ = copy.deepcopy(sessions)  # Avoid clash with global variable.
-
-    env = os.environ
-    if env.get('MNE_BIDS_STUDY_SESSION'):
-        sessions_ = env['MNE_BIDS_STUDY_SESSION']
-    elif sessions_ == 'all':
-        sessions_ = get_entity_vals(bids_root, entity_key='session')
-
-    if not sessions_:
-        return [None]
-    else:
-        return sessions_
-
-
-def get_runs() -> list:
-    runs_ = copy.deepcopy(runs)  # Avoid clash with global variable.
-    valid_runs = get_entity_vals(get_bids_root(), entity_key='run')
-
-    env_run = os.environ.get('MNE_BIDS_STUDY_RUN')
-    if env_run and env_run not in valid_runs:
-        raise ValueError(
-            f'Invalid run. It can be {valid_runs} but '
-            f'got {env_run}')
-    elif env_run:
-        runs_ = [env_run]
-    elif runs_ == 'all':
-        runs_ = valid_runs
-
-    if not runs_:
-        return [None]
-    else:
-        return runs_
-
-
-# XXX This check should actually go into the CHECKS section, but it depends
-# XXX on get_runs(), which is defined after that section.
-if mf_reference_run is not None and mf_reference_run not in get_runs():
-    msg = (f'You set mf_reference_run={mf_reference_run}, but your dataset '
-           f'only contains the following runs: {get_runs()}')
-    raise ValueError(msg)
-
-
-def get_mf_reference_run() -> str:
-    # Retrieve to run identifier (number, name) of the reference run
-    if mf_reference_run is None:
-        # Use the first run
-        return get_runs()[0]
-    else:
-        return mf_reference_run
-
-
 def get_subjects() -> List[str]:
     global subjects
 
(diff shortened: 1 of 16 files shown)


config.py
```diff
@@ -1427,9 +1453,67 @@ def get_sessions():
...
+
+def get_runs(subject: str, verbose: bool = False) -> list:
+    """Returns a list of runs.
+
+    Parameters
+    ----------
+    subject
+        Returns a list of the runs of this subject.
+    verbose
+        Notify if different subjects do not share the same runs.
```
agramfort, line 1497: Returns section missing

agramfort (commented):
is it already functional on your end @crsegerie ?

config.py
```diff
@@ -1427,9 +1453,72 @@ def get_sessions():
         return sessions_
 
 
-def get_runs() -> list:
+def get_runs_all_subjects() -> dict:
+    '''Returns a dict of runs present in the bids_path
+    for each subject asked in the configuration file
+    (and not for each subject present in the bids_path).'''
+    # We cannot use get_subjects() because if there is just one subject
+    # we need to creat the full list of ignore_subjects
```
agramfort, line 1461:
```suggestion
    # we need to create the full list of ignore_subjects
```

but move this comment below as here it's not yet understandable what you mean

config.py
```diff
@@ -1427,9 +1453,72 @@ def get_sessions():
...
+    return subj_runs
+
+
+def get_intersect_run() -> list:
+    '''Returns the intersection of all the runs of all subjects.'''
+    subj_runs = get_runs_all_subjects()
+    return list(set.intersection(*map(set, subj_runs.values())))
+
+
+def get_runs(subject: str, verbose: bool = False) -> list:
```
hoechenberger, line 1489:
```suggestion
def get_runs(subject: str, verbose: bool = False) -> Union[List[str], List[None]]:
```

config.py
```diff
@@ -1427,9 +1453,72 @@ def get_sessions():
...
+    ----------
+    subject
+        Returns a list of the runs of this subject.
+    verbose
+        Notify if different subjects do not share the same runs.
+
+    Returns
+    -------
+    list of string
+        list of runs of the subject.
```
hoechenberger, line 1502:
```suggestion
    The list of runs of the subject. If no BIDS `run` entity could be found,
    returns `[None]`.
```

config.py
```diff
@@ -1427,9 +1453,72 @@ def get_sessions():
...
+
+        same_runs = True
+        for runs_sub_i in subj_runs.values():
+            if set(runs_sub_i) != set(list(subj_runs.values())[0]):
+                same_runs = False
+
+        if not same_runs:
+            msg = ("Extracted all the runs. "
+                   "Beware, not all subjects share the same "
+                   "list of run-names.")
```
hoechenberger, line 1520:
```suggestion
            msg = ('Extracted all the runs. '
                   'Beware, not all subjects share the same '
                   'set of runs.')
```

config.py
```diff
@@ -1427,9 +1453,72 @@ def get_sessions():
...
-    valid_runs = get_entity_vals(get_bids_root(), entity_key='run')
+
+    subj_runs = get_runs_all_subjects()
+    valid_runs = subj_runs[subject]
+
+    if len(get_subjects()) > 1:
+        # Notify if different subjects do not share the same runs
+
+        same_runs = True
+        for runs_sub_i in subj_runs.values():
```
agramfort, line 1513: maybe something like

`if len(set(subj_runs.values())) > 2:`

ie enough to know if some subjects have different number of runs?

docs/source/changes.md
```diff
@@ -41,6 +41,8 @@ authors:
   ({{ gh(322) }} by {{ authors.hoechenberger }})
 - Warn if using ICA and no EOG- or ECG-related ICs were detected.
   ({{ gh(351) }} by {{ authors.crsegerie }})
+- Added the possibility to have different runs for different subjects.
+  ({{ gh(?) }} by {{ authors.crsegerie }}))
```
hoechenberger, line 45: don't forget to update this one :)

scripts/preprocessing/02-frequency_filter.py
```diff
@@ -141,9 +141,17 @@ def main():
     logger.info(gen_log_message(step=2, message=msg))
 
     parallel, run_func, _ = parallel_func(run_filter, n_jobs=config.N_JOBS)
+
+    # Enabling different runs for different subjects
+    sub_run_ses = []
+    for subject in config.get_subjects():
+        sub_run_ses = sub_run_ses + list(itertools.product(
```
agramfort, line 148:
```suggestion
        sub_run_ses += list(itertools.product(
```

hoechenberger (commented):
Left some small suggestions

Is this already working as intended for you locally? :)

crsegerie:
> is it already functional on your end @crsegerie ?
Not quite.

There is still an error on my own : 
> 02:08:52 [Step-03][sub-403, run-07] Loading filtered raw data from /home/charb/Desktop/parietal/time_in_wm/derivatives/mne-bids-pipeline/sub-403/meg/sub-403_task-wm_run-07_proc-filt_raw.fif
> 02:08:53 [Step-03][sub-403, run-08] Loading filtered raw data from /home/charb/Desktop/parietal/time_in_wm/derivatives/mne-bids-pipeline/sub-403/meg/sub-403_task-wm_run-08_proc-filt_raw.fif
> Killed
The killed surprises me, because I do not see it in CI

And at the beginning of each step, the warning appears:
> 01:58:28 Extracted all the runs. Beware, not all subjects share the same list of run-names.
I should change it to only warn for the first step

config.py
crsegerie: That was what I've first tried, but then I got the error : unhashable type : list

hoechenberger:
> > Killed

This most likely indicates that you're running out of memory … We'd need to have to see how to optimize things here. This problem seems unrelated to this PR though

hoechenberger:
@agramfort Do you think instead of concatenating all filtered runs before creating epochs, we could create epochs for each run individually and then concat those? This would avoid the out-of-memory problem @crsegerie sees locally

hoechenberger:
> @agramfort Do you think instead of concatenating all filtered runs before creating epochs, we could create epochs for each run individually and then concat those? 

Ah, right, there is [`mne.concatenate_epochs()`](https://mne.tools/dev/generated/mne.concatenate_epochs.html)

I can take a stab at this sometime later today (separate PR)

config.py
```diff
@@ -1427,9 +1453,72 @@ def get_sessions():
...
+    ----------
+    subject
+        Returns a list of the runs of this subject.
+    verbose
+        Notify if different subjects do not share the same runs.
+
+    Returns
+    -------
+    The list of runs of the subject. If no BIDS `run` entity could be found,
+    returns `[None]`.
```
agramfort, line 1502:
```suggestion
    runs : list
        The list of runs of the subject. If no BIDS `run` entity could be found,
        returns `[]`.
```

I think it should return an empty list if no runs are found

config.py
agramfort: then maybe:

`if len(set(map(tuple, subj_runs.values()))) > 2:`

?

config.py
```diff
@@ -1444,52 +1533,37 @@ def get_runs() -> list:
...
     # Retrieve to run identifier (number, name) of the reference run
     if mf_reference_run is None:
         # Use the first run
-        return get_runs()[0]
+        if inter_runs:
+            return inter_runs[0]
+        else:
+            ValueError("The intersection of runs by subjects is empty"
+                       "Check the list of runs:"
+                       f"{get_runs_all_subjects()}")
```
agramfort, line 1562:
```suggestion
            ValueError("The intersection of runs by subjects is empty. "
                       "Check the list of runs: "
                       f"{get_runs_all_subjects()}")
```

config.py
hoechenberger: Yes this would be preferable but it would require refactoring other places of the pipeline too. I'd say we leave it like this and try to change this sometime later in a followup PR

crsegerie:
There is  a conflict I'm not sure how to deal with : 

image

I do not understand the MKDOC part ?

hoechenberger:
> I do not understand the MKDOC part ?

I added this `MKDOCS` check to make it work with our documentation build, which uses [MkDocs](https://www.mkdocs.org). It's a bit hacky but seems to work: When building the docs, we set this environment variable so the config "knows": "oh ok, so this is a doc build, not an actual run of the pipeline."

config.py
```diff
@@ -1551,15 +1551,15 @@ def get_runs(subject: str, verbose: bool = False) -> Union[List[str], List[None]
...
-inter_runs = get_intersect_run()
-mf_ref_error = (
-    ('MKDOCS' not in os.environ) and
-    (mf_reference_run is not None) and
-    (mf_reference_run not in inter_runs))
-if mf_ref_error:
-    msg = (f'You set mf_reference_run={mf_reference_run}, but your dataset '
-           f'only contains the following runs: {inter_runs}')
-    raise ValueError(msg)
+if ('MKDOCS' not in os.environ):
```
hoechenberger, line 1554:
```suggestion
if 'MKDOCS' not in os.environ:
```

config.py
```diff
@@ -1551,7 +1551,7 @@ def get_runs(subject: str, verbose: bool = False) -> Union[List[str], List[None]
 
 
 # XXX This check should actually go into the CHECKS section, but it depends
-# XXX on get_runs(), which is defined after that section.
+# XXX on get_runs(subject=subject), which is defined after that section.
```
hoechenberger, line 1554: Can you please undo this change? In the comment here we left out the parameters on purpose – after all, we just wanted to point out a function name

config.py
```diff
@@ -1479,9 +1505,72 @@ def get_sessions():
...
+        valid_runs_subj = get_entity_vals(
+            get_bids_root(), entity_key='run',
+            ignore_subjects=ignore_subjects)
+        subj_runs[subj] = valid_runs_subj
+
+    return subj_runs
+
+
+def get_intersect_run() -> list:
+    '''Returns the intersection of all the runs of all subjects.'''
```
hoechenberger, line 1536: Can you use double-quotes " around the docstring please?

config.py
```diff
@@ -1479,9 +1505,72 @@ def get_sessions():
         return sessions_
 
 
-def get_runs() -> list:
+def get_runs_all_subjects() -> dict:
+    '''Returns a dict of runs present in the bids_path
+    for each subject asked in the configuration file
+    (and not for each subject present in the bids_path).'''
```
hoechenberger, line 1511: Can you use double-quotes " around the docstring please and make it such that you have a short one-sentence summary in the first line? The docstring must look sth like this:

```
"""Do this and that.

More detailed explanation here!
"""
```

You also don't need to mention the return types there – ideally, add a `Returns` section for this purpose!

hoechenberger (commented):
Just some nitpicks re docstrings, otherwise LGTM

config.py
```diff
@@ -1479,9 +1505,76 @@ def get_sessions():
...
-def get_runs() -> list:
+def get_runs_all_subjects() -> dict:
+    """Gives the mapping between subjects and their runs.
+
+    Returns
+    -------
+    a dict of runs present in the bids_path
+    for each subject asked in the configuration file
+    (and not for each subject present in the bids_path)."""
+    # We cannot use get_subjects() because if there is just one subject
```
hoechenberger, line 1516:
```suggestion
    (and not for each subject present in the bids_path).
    """
    # We cannot use get_subjects() because if there is just one subject
```

hoechenberger (approved):

hoechenberger:
Great job, @crsegerie!!!

agramfort:
🎉

PR #355, opened 2021-05-29
Reduce memory usage when creating epochs

hoechenberger:
### Before merging …


hoechenberger:
@crsegerie Using this branch, do you still see the issue with the killed process?

hoechenberger:
It just crossed my mind, we need to do something similar during ICA prep too

hoechenberger:
WIP:
Needs handling for EOG and ECG epoching

hoechenberger:
@agramfort CI fails because `joblib` doesn't preserve `sys.path` when spawning a new process. Is that by design, or something that should be reported as a bug?

I could work around this but it would be a bit ugly.

run.py
```diff
@@ -98,7 +98,11 @@ def _run_script(script_path, config, root_dir, subject, session, task, run):
     if subject:
         env['MNE_BIDS_STUDY_SUBJECT'] = subject
 
+    # Insert the script's parent directory into the Python search path, so
+    # the script can import other scripts from the same directory
+    sys.path.insert(0, str(script_path.parent))
     runpy.run_path(script_path, run_name='__main__')
+    sys.path.remove(str(script_path.parent))
```
hoechenberger, line 105:
```suggestion
    sys.path.insert(0, str(script_path.parent.absolute()))
    runpy.run_path(script_path, run_name='__main__')
    sys.path.remove(str(script_path.parent.absolute()))
```

agramfort (commented):
diff looks good

did you check that the rendered reports now look almost identical as before?

crsegerie:
Weird, when I run your code I get an error:
image



I tried to debug a little and I got:
image

It's because at line 569 we assign `config = None`

hoechenberger:
@crsegerie This is an issue in MNE-BIDS not being able to locate some files and then believing it might be BTi data (which comes in a **directory**, it's unrelated to the Pipeline *per se*
Is this when it's trying to start processing the second participant for you?

crsegerie:
@hoechenberger Yes you are right!

hoechenberger:
> @hoechenberger Yes you are right!

@crsegerie Care to make an issue report for MNE-BIDS? (Not sure if we ever made one!) This issue is super annoying and has been around for ages … we really should fix it. You don't need to put together a working example, just state what happened :)

hoechenberger:
> did you check that the rendered reports now look almost identical as before?

No, will do once CI passes!

agramfort (approved):
awesome job @hoechenberger !

PR #361, opened 2021-06-01
Baseline interval check

### LINKED ISSUE #356: Missing Baseline interval Check
I think there is a missing check which should have followed the reading of the config file:

image

After having set the baseline to (-1, 0), I should have been warned immediately that this won't be compatible with the epochs data.

crsegerie:
Closes #356.

### Before merging …

I used `isinstance(baseline[0], numbers.Number)`, which is a practical way of handling the cases with t=0 or t=None.

### DIFF
--- a/config.py
+++ b/config.py
@@ -1378,12 +1378,12 @@ if bem_mri_images not in ('FLASH', 'T1', 'auto'):
 
 if baseline[0] and baseline[0] < epochs_tmin:
     msg = ('baseline should be contained in [epochs_tmin, epochs_tmax]. '
-           'But {baseline} in not contained in {[epochs_tmin, epochs_tmax]}')
+           f'But {baseline} in not contained in {[epochs_tmin, epochs_tmax]}')
     raise ValueError(msg)
 
 if baseline[1] and baseline[1] > epochs_tmax:
     msg = ('baseline should be contained in [epochs_tmin, epochs_tmax]. '
-           'But {baseline} in not contained in {[epochs_tmin, epochs_tmax]}')
+           f'But {baseline} in not contained in {[epochs_tmin, epochs_tmax]}')
     raise ValueError(msg)
 
 
--- a/config.py
+++ b/config.py
@@ -1421,7 +1421,7 @@ def check_baseline(
             (baseline[1] is not None) and
             (baseline[0] >= baseline[1])):
         msg = (f'The end of the baseline period must occur after its start, '
-               f'but you set baseline={baseline}'
+               f'but you set baseline={baseline}')
         raise ValueError(msg)
 
 
--- a/config.py
+++ b/config.py
@@ -1376,6 +1376,16 @@ if bem_mri_images not in ('FLASH', 'T1', 'auto'):
            f'are: "FLASH", "T1", and "auto".')
     raise ValueError(msg)
 
+if baseline[0] and baseline[0] < epochs_tmin:
+    msg = ('baseline should be contained in [epochs_tmin, epochs_tmax]. '
+           'But {baseline} in not contained in {[epochs_tmin, epochs_tmax]}')
+    raise ValueError(msg)
+
+if baseline[1] and baseline[1] > epochs_tmax:
+    msg = ('baseline should be contained in [epochs_tmin, epochs_tmax]. '
+           'But {baseline} in not contained in {[epochs_tmin, epochs_tmax]}')
+    raise ValueError(msg)
+
 
 ###############################################################################
 # Helper functions
--- a/config.py
+++ b/config.py
@@ -1387,17 +1387,21 @@ if bem_mri_images not in ('FLASH', 'T1', 'auto'):
     raise ValueError(msg)
 
 
-def check_baseline(baseline: Tuple[Optional[float], Optional[float]],
-                   epochs_tmin: float, epochs_tmax: float) -> None:
+def check_baseline(
+    *,
+    baseline: Tuple[Optional[float], Optional[float]],
+    epochs_tmin: float,
+    epochs_tmax: float
+) -> None:
     """Raises error if baseline not compatible with [epochs_tmin, epochs_tmax].
 
     Parameters
     ----------
     baseline
         Tuple indicating the beginning and end of the baseline interval.
-    epochs_tmin:
+    epochs_tmin
         Beginning of Epochs.
-    epochs_tmin:
+    epochs_tmax
         End of Epochs.
 
     Raises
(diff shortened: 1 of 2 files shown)


config.py
```diff
@@ -1377,6 +1378,23 @@ def get_t1_from_meeg(bids_path):
     raise ValueError(msg)
 
 
+if isinstance(baseline[0], numbers.Number) and baseline[0] < epochs_tmin:
```
agramfort, line 1381: you should define a check_baseline function and deal with the fact that baseline can be None.

crsegerie: I don't understand. I think that even if it's None, my check handles it: 

image

Or am I missing some cases?

Otherwise, yes you are right it would be better with a function.

hoechenberger: @agramfort 
> you should define a check_baseline function

I don't think that's necessary – the approach @crsegerie took IMHO fits perfectly fine into how we've been doing things in `config.py` so far. We only need to check whether the baseline makes sense **once**, upon loading the config.

However I would suggest switching from `isinstance(baseline[1], numbers.Number)` to `... is not None`

agramfort: indeed my bad. So I would just encapsulate this test in a check_baseline function. Doing so it's explicit what variable is used by the check and you don't pollute the namespace with for example the msg variable

hoechenberger: Sounds like a good idea! Let's teach good practice to @crsegerie, even if that means to tell him: don't do it the way we did it before 😅😁🙈

agramfort:
make it work, make it nice, make it fast.

it's never too late to clean "after it works"

>

config.py
```diff
@@ -1386,6 +1387,48 @@ def get_t1_from_meeg(bids_path):
     raise ValueError(msg)
 
 
+def check_baseline(baseline: Tuple[Optional[float], Optional[float]],
+                   epochs_tmin: float, epochs_tmax: float) -> None:
```
hoechenberger, line 1391:
```suggestion
def check_baseline(
    *,
    baseline: Tuple[Optional[float], Optional[float]],
    epochs_tmin: float,
    epochs_tmax: float
) -> None:
```

config.py
```diff
@@ -1386,6 +1387,48 @@ def get_t1_from_meeg(bids_path):
...
+    """Raises error if baseline not compatible with [epochs_tmin, epochs_tmax].
+
+    Parameters
+    ----------
+    baseline
+        Tuple indicating the beginning and end of the baseline interval.
+    epochs_tmin:
```
hoechenberger, line 1398:
```suggestion
    epochs_tmin
```

config.py
```diff
@@ -1386,6 +1387,48 @@ def get_t1_from_meeg(bids_path):
...
+        Beginning of Epochs.
+    epochs_tmin:
```
hoechenberger, line 1400:
```suggestion
    epochs_tmax
```

docs/source/changes.md
```diff
@@ -47,6 +47,8 @@ authors:
   ({{ gh(322) }} by {{ authors.hoechenberger }})
 - Warn if using ICA and no EOG- or ECG-related ICs were detected.
   ({{ gh(351) }} by {{ authors.crsegerie }})
+- Check that the baseline interval is compatible with [epochs_tmin, epochs_tmax] in the user's configuration file.
```
hoechenberger, line 50:
```suggestion
- Check that the baseline interval falls into `[epochs_tmin, epochs_tmax]`.
```

config.py
```diff
@@ -1386,6 +1387,48 @@ def get_t1_from_meeg(bids_path):
...
+        msg = ('baseline should be contained in [epochs_tmin, epochs_tmax]. '
+               f'But {baseline} in not contained in '
+               f'{[epochs_tmin, epochs_tmax]}.')
+        raise ValueError(msg)
+
+    if baseline[1] is not None and baseline[1] > epochs_tmax:
+        msg = ('baseline should be contained in [epochs_tmin, epochs_tmax]. '
+               f'But {baseline} in not contained in '
+               f'{[epochs_tmin, epochs_tmax]}.')
+        raise ValueError(msg)
```
hoechenberger, line 1419:
```suggestion
    if ((baseline[0] is not None and baseline[0] < epochs_tmin) or
            (baseline[1] is not None and baseline[1] > epochs_tmax)):
        msg = (f'baseline {baseline} outside of epochs interval '
               f'{[epochs_tmin, epochs_tmax]}.')

        raise ValueError(msg)
```

config.py
```diff
@@ -1386,6 +1387,48 @@ def get_t1_from_meeg(bids_path):
...
+
+    if ((baseline[0] is not None) and
+        (baseline[1] is not None) and
+            (baseline[0] >= baseline[1])):
+        msg = f'{baseline} is not a correct time-interval.'
+        raise ValueError(msg)
```
hoechenberger, line 1425:
```suggestion
    if ((baseline[0] is not None) and
            (baseline[1] is not None) and
            (baseline[0] >= baseline[1])):
        msg = (f'The end of the baseline period must occur after its start, '
               f'but you set baseline={baseline}'
        raise ValueError(msg)
```

hoechenberger (commented):
I left a bunch of simplifications and style improvements

config.py
crsegerie: why the *,  in the parameters ?

config.py
crsegerie: habits die hard

config.py
drammock: `*` in this case forces all arguments to be keyword arguments (no positional args allowed)

hoechenberger:
Thanks @crsegerie! 🥳

config.py
```diff
@@ -1386,6 +1387,48 @@ def get_t1_from_meeg(bids_path):
...
+    epochs_tmax
+        End of Epochs.
+
+    Raises
+    ------
+    ValueError
+        if baseline not contained in [epochs_tmin, epochs_tmax].
+        if baseline is not a correct time-interval.
+    """
+    if ((baseline[0] is not None and baseline[0] < epochs_tmin) or
```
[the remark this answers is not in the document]
agramfort: see https://github.com/mne-tools/mne-bids-pipeline/pull/363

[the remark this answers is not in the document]
crsegerie: Yes you are right, I overlooked this. Time to set up the type checker in pylint!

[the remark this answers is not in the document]
hoechenberger: That's not so easy, as we're not typing everything. But you can just enable basic type checking in Pylance, this is what I'm using.

PR #363, opened 2021-06-02
check baseline to support None

agramfort:


hoechenberger:
Good catch @agramfort! I overlooked this one

PR #365, opened 2021-06-04
DNM: Add support for fNIRS data

rob-luke:
Hi, would you like to add support for fNIRS to mne-bids-pipeline?

This is an issue as PR. And allowed me to understand this package and determine what would be needed for it to run well with fNIRS data. It has the following issues to be addressed before a review is appropriate:

- [ ] This PR relies on allowing snirf support in `mne.Report`. See https://github.com/mne-tools/mne-python/pull/9443
- [ ] This PR uses functions from MNE-NIRS. Is this undesirable? If so, then the convenience functions (e.g. `get_short_channels`) could be easily reimplemented, they are trivial. And we could either remove the signal enhancement functions or move them to MNE (they aren't 100% stable yet, so probably just best to remove them from here).
- [ ] Wait for the fNIRS BIDS spec to be accepted. See progress here: https://github.com/bids-standard/bids-specification/pull/802

### DIFF
--- a/config.py
+++ b/config.py
@@ -1341,7 +1341,7 @@ elif 'eeg' in ch_types and len(ch_types) > 1:  # EEG + some other channel types
     msg = ('EEG data can only be analyzed separately from other channel '
            'types. Please adjust `ch_types` in your configuration.')
     raise ValueError(msg)
-elif any([ch_type not in ('meg', 'mag', 'grad') for ch_type in ch_types]):
+elif any([ch_type not in ('meg', 'mag', 'grad', 'nirs') for ch_type in ch_types]):
     msg = ('Invalid channel type passed. Please adjust `ch_types` in your '
            'configuration.')
     raise ValueError(msg)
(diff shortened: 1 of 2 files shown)


rob-luke:
Using the pipeline from this PR I get nice reports and the following decoding performance on my [tapping data](https://github.com/rob-luke/BIDS-NIRS-Tapping) for tapping with left hand vs tapping with right hand...

image

and the following performance on my [Audio Visual](https://www.sciencedirect.com/science/article/pii/S0378595521000903) data for the visual condition vs auditory condition

image

agramfort (commented):
So far so good @rob-luke 👍

larsoner:
@rob-luke we should come back to this -- now that I'm more familiar with mne-bids-pipeline I think it wouldn't be too bad to get this working. Do you want to try a new PR with the latest file structure? If you don't have time I could take a stab at it

rob-luke:
> we should come back to this

Agreed!

My plan of attack here is:
1. Get fNIRS BIDS officially merged (should happen in coming days)
2. Update MNE-BIDS to meet merged BIDS spec for fNIRS and add recommended fields
3. Tackle this PR in MNE-BIDS-Pipeline

> Do you want to try a new PR with the latest file structure? If you don't have time I could take a stab at it

If you want to take a stab at this, then go wild :) I will be doing the items above in that order unless there is a reason not too.

rob-luke:
FYI. fNIRS is now merged into BIDS main. So I will move on to adding complete support in MNE-BIDS.

larsoner:
Go go go @rob-luke 🚀

PR #370, opened 2021-06-07
add exclude_runs

### LINKED ISSUE #366: Adding a param exclude_runs?
I am working on a study with 20 subjects with 8 runs each. But among these 20*8 runs, only 3 are faulty. So I would like to add a parameter `exclude_runs` mapping the participants to their faulty runs. For example: `exclude_runs={"215": ["03"], ...}`. I started to discuss with @hoechenberger, and he suggested me to create a method to specify the valid runs to consider instead. Here is the transcript of the conversation

Charbel 
Shall I begin to code the param `exclude_runs`?

Richard 
I thought what we wanted instead was specifying the includes for each participant individually … no?

Charbel
`include_runs` is already `runs`. And `exclude_runs` would be more symmetric because we already have a param `exclude_subjects`. wdyt ?

Richard 
I generally like the idea but I’m wondering if “bad” runs aren’t a problem that should be fixed in BIDS beforehand?
With subjects it’s different – you may want to exclude only a small number of participants from a big study without having to provide a super long “subjects” list. Whereas for runs, we rarely have more than 8, 9, maybe 10, so it’s easy to specify all that you want explicitly

Charbel

> “bad” runs aren’t a problem that should be fixed in BIDS beforehand?

Since the exclusion of whole subjects is already done in the params of the bids pipeline, I don't think it would be a considerable conceptual stretch to add the exclude_runs param?
But yes, having a long list of runs isn't a problem, and would be very easy to code.

crsegerie:
Closes #366

### Before merging …

### DIFF
--- a/config.py
+++ b/config.py
@@ -89,7 +89,7 @@ runs: Union[Iterable, Literal['all']] = 'all'
 The runs to process.
 """
 
-exclude_runs: Optional[dict[str, list[str]]] = None
+exclude_runs: Optional[Dict[str, List[str]]] = None
 """
 Specify runs to exclude from analysis.
 
--- a/config.py
+++ b/config.py
@@ -91,7 +91,7 @@ The runs to process.
 
 exclude_runs: Optional[Dict[str, List[str]]] = None
 """
-Specify runs to exclude from analysis.
+Specify runs to exclude from analysis, for each participant invidiually.
 
 ???+ example "Example"
     ```python
--- a/config.py
+++ b/config.py
@@ -1570,8 +1570,8 @@ def get_runs_all_subjects() -> dict:
             get_bids_root(), entity_key='run',
             ignore_subjects=ignore_subjects)
         if exclude_runs and subj in exclude_runs:
-            valid_runs_subj = [
-                r for r in valid_runs_subj if r not in exclude_runs[subj]]
+            valid_runs_subj = [r for r in valid_runs_subj
+                               if r not in exclude_runs[subj]]
         subj_runs[subj] = valid_runs_subj
 
     return subj_runs
--- a/config.py
+++ b/config.py
@@ -89,6 +89,22 @@ runs: Union[Iterable, Literal['all']] = 'all'
 The runs to process.
 """
 
+exclude_runs: Optional[dict[str, list[str]]] = None
+"""
+Specify runs to exclude from analysis.
+
+???+ example "Example"
+    ```python
+    exclude_runs = None  # Include all runs.
+    exclude_runs = {'01': ['02']}  # Exclude run 02 of subject 01.
+    ```
+
+???+ info "Good Practice / Advice"
+    Keep track of the criteria leading you to exclude
+    a run (e.g. too many movements, missing blocks, aborted experiment,
+    did not understand the instructions, etc, ...).
+"""
+
 crop_runs: Optional[Tuple[float, float]] = None
 """
 Crop the raw data of each run to the specified time interval ``[tmin, tmax]``,
@@ -1553,6 +1569,9 @@ def get_runs_all_subjects() -> dict:
         valid_runs_subj = get_entity_vals(
             get_bids_root(), entity_key='run',
             ignore_subjects=ignore_subjects)
+        if exclude_runs and subj in exclude_runs:
+            valid_runs_subj = [
+                r for r in valid_runs_subj if r not in exclude_runs[subj]]
         subj_runs[subj] = valid_runs_subj
 
     return subj_runs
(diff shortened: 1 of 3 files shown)


config.py
```diff
@@ -89,6 +89,22 @@
...
+???+ example "Example"
+    ```python
+    exclude_runs = None  # Include all runs.
+    exclude_runs = {'01': ['02']}  # Exclude run 02 of subject 01.
+    ```
+
+???+ info "Good Practice / Advice"
+    Keep track of the criteria leading you to exclude
+    a run (e.g. too many movements, missing blocks, aborted experiment,
+    did not understand the instructions, etc, ...).
```
hoechenberger, line 105:
```suggestion
    did not understand the instructions, etc.).
```

config.py
```diff
@@ -89,6 +89,22 @@
 The runs to process.
 """
 
+exclude_runs: Optional[Dict[str, List[str]]] = None
```
hoechenberger, line 92: I think we should support both, a dict of lists (as you do now) **and** a list of strings (as we do with `exclude_subjects`) to exclude the same run across all subjects. WDYT?

config.py
```diff
@@ -1553,6 +1569,9 @@ def get_runs_all_subjects() -> dict:
         valid_runs_subj = get_entity_vals(
             get_bids_root(), entity_key='run',
             ignore_subjects=ignore_subjects)
+        if exclude_runs and subj in exclude_runs:
+            valid_runs_subj = [
+                r for r in valid_runs_subj if r not in exclude_runs[subj]]
```
hoechenberger, line 1574: Maybe easier to read:

```suggestion
            valid_runs_subj = [r for r in valid_runs_subj
                               if r not in exclude_runs[subj]]
```

docs/source/changes.md
```diff
@@ -5,6 +5,7 @@ authors:
   hoechenberger: "[Richard Höchenberger](https://github.com/hoechenberger)"
   guiomar: "[Julia Guiomar Niso Galán](https://github.com/guiomar)"
   robluke: "[Robert Luke](https://github.com/rob-luke)"
+  crsegerie: "[Charbel-Raphaël Segerie](https://github.com/crsegerie)"
```
hoechenberger, line 8: hum, why were you missing here all of a sudden? 😱🧐

docs/source/changes.md
```diff
@@ -21,6 +22,8 @@ authors:
   ({{ gh(355) }} by {{ authors.hoechenberger }})
 - Add time-frequency plot to report.
   ({{ gh(367) }} by {{ authors.robluke }})
+- Add the possibility to `exclude_runs`.
```
hoechenberger, line 25:
```suggestion
- Add possibility to exclude runs from the analysis via the new
  [`exclude_runs`][config.exclude_runs] setting.
```

hoechenberger (commented):
Good start, I left a few comments!

config.py
hoechenberger: cc @agramfort (see my previous comment)

crsegerie: I thought about it, but I think that the usage of this exclude_runs should be parsimonious. And if I want to exclude a run for all subjects I can just delete it from the list of runs in the runs list.

agramfort: I think we can add this later if requested

agramfort (approved):
LGTM

config.py
```diff
@@ -89,6 +89,22 @@
 The runs to process.
 """
 
+exclude_runs: Optional[Dict[str, List[str]]] = None
+"""
+Specify runs to exclude from analysis.
```
hoechenberger, line 94:
```suggestion
Specify runs to exclude from analysis, for each participant invidiually.
```

config.py
hoechenberger: Ok fair enough

hoechenberger (approved):

crsegerie (commented):
individually

hoechenberger (changes requested):
Oh we forgot to add the new param to the docs. @crsegerie Can you have a look how we do this with the other params? Search for: `::: config.`

config.py
```diff
@@ -89,6 +89,22 @@
 The runs to process.
 """
 
+exclude_runs: Optional[Dict[str, List[str]]] = None
+"""
+Specify runs to exclude from analysis, for each participant invidiually.
```
hoechenberger:
```suggestion
Specify runs to exclude from analysis, for each participant individually.
```

hoechenberger (approved):
LGTM! Let's wait for the doc build to come back before merging

hoechenberger:
Doc build looks good 

https://7635-167957354-gh.circle-artifacts.com/0/site/settings/general.html

hoechenberger:
Thanks @crsegerie!!

PR #375, opened 2021-06-10
Cast command line arguments to strings

hoechenberger:
Cast command line arguments to strings

Avoid the bug we saw yesterday with @agramfort

### Before merging …


docs/source/changes.md
```diff
@@ -92,3 +92,5 @@ authors:
   ({{ gh(336) }} by {{ authors.agramfort }}, {{ authors.hoechenberger }}, and {{ authors.crsegerie }})
 - `epochs_tmin` and `epochs_tmax` were named incorrectly in some test config files.
   ({{ gh(340) }} by {{ authors.crsegerie }})
+- Passing subject, session, task, and run names that consist only of numbers to `run.py` doesn't throw an exception anymore, even if the names weren't double-quotes.
```
hoechenberger, line 95:
```suggestion
- Passing subject, session, task, and run names that consist only of numbers to `run.py` doesn't throw an exception anymore, even if the values weren't double-quoted.
```

docs/source/changes.md
```diff
@@ -92,3 +92,5 @@ authors:
...
+  ({{ gh(374) }} by {{ authors.hoechenberger }})
```
hoechenberger, line 96:
```suggestion
  ({{ gh(375) }} by {{ authors.hoechenberger }})
```

agramfort:
Does it work if subject starts with 0?

hoechenberger:
> Does it work if subject starts with 0?

I don't think so, and there's no obvious way to fix it… unless if maybe we ask users to specify the subject ID as `sub-xxx`, this would remove the  ambiguity… But then we'd still have the same issue for session, run, and potentially task :(

agramfort:
then I would say that if you detect that fire converted run or subject to
an int you give
an informative message on how to format the parameter on the command line to
keep it a string.

my 2c

>

hoechenberger:
@agramfort I was actually mistaken, if there are trailing zeros, it works out of the box! And with the changes here, we now also handle numbers that come without trainling zeros. So I think it's a good thing to have! Let's wait for CI to come back

agramfort (approved):

agramfort:
thx @hoechenberger

PR #378, opened 2021-06-11
MRG, ENH: (Re)create high-resolution scalp surface

hoechenberger:
Adds new config option `recreate_scalp_surfaces`

I got error messages for some datasets when generating the report, where it
would complain about not being able to fine the surfaces. This problem is
gone with this PR.

### Before merging …


agramfort:
let's try to get this one https://github.com/mne-tools/mne-python/pull/9463 in first.

ok for you? it will be cleaner.

hoechenberger:
Perfect!

agramfort:
you can now use https://github.com/mne-tools/mne-python/pull/9463

hoechenberger:
CI will fail until https://github.com/mne-tools/mne-python/pull/9465 has been merged

agramfort (approved):
Merge when green

hoechenberger:
@larsoner In this PR, I'm adding creation of scalp surfaces. This requires `mkheadsurf`, which apparently is not included in the minimal set of commands we're using in CI. Is there an easy way to get this binary to run without installing the entire FreeSurfer package?

larsoner:
We could add it to `minimal_cmds`, but I'm pretty sure the reason we don't include it is that it takes a long time to run. I'd advocate for making a fake `mkheadsurf` script just for these doc builds that just `wget`s a static file from some remote storage location and puts it in the right place, assuming that the FreeSurfer command would have produced the correct file.

If you really want, we can add it to `minimal_cmds`, basically you can 1) do what our CIs do to get the files and set paths, 2) manually copy the freesurfer binary over to the `bin/` directory, and 3) try running it, and if it fails 4) pull in whatever other libraries/binaries it needs to run properly.

hoechenberger:
@larsoner I've assembled all additional scripts & binaries required to run `mkheadsurf`. Here's the complete list:

```
1.1K Jan 18  2017 getpwdcmd
9.6K Jan 18  2017 mkheadsurf
 69B Jan 18  2017 mri_binarize
5.4M Jan 18  2017 mri_seghead
5.7M Jan 18  2017 mri_tessellate
```
Compressed this is about 4 MB

Mind if I just add these to `minimal_cmds.zip` for Linux? I've already tested running `mne make_scalp_surfaces` with these binaries on CircleCI and it's working.

larsoner:
Those sizes seem fine, so that covers one of two considerations I mentioned. How much time does it take to run? Is testing the FS command's functionality (and our interfacing with it) worth this extra time on every build compared to just faking it?

hoechenberger:
@larsoner We only run this command for the test cases where we actually produce source estimates. I believe it's important to run & inspect the whole thing (it produces output that goes into the report) to help us ensure everything is working the way we want it to

I did some timing tests on Circle:

```
$ time mne make_scalp_surfaces --subjects-dir=/home/circleci/mne_data/ds000248/derivatives/freesurfer/subjects --subject=sub-01 --force --overwrite

real	2m20.973s
user	2m8.234s
sys	0m1.027s
```
If we don't decimate (i.e. only produce the "dense" surface):
```
time mne make_scalp_surfaces --subjects-dir=/home/circleci/mne_data/ds000248/derivatives/freesurfer/subjects --subject=sub-01 --force --overwrite --no-decimate

real	0m41.302s
user	0m36.710s
sys	0m0.308s
```

WDYT?

larsoner:
40s is actually not too terrible. If it were up to me I'd still make a fake `mri_seghead` script that just copied in a remote file to save time on every PR since 1) I think it's safe to trust that FreeSurfer won't break `mri_seghead` anytime soon, and if they do, we'll see it elsewhere, and 2) the build time is already a bit long, and this will make that problem worse. But I can live with you adding those binaries. If you go that route, though, please add them for Linux and macOS as they should ideally be kept in sync

hoechenberger:
> If you go that route, though, please add them for Linux and macOS as they should ideally be kept in sync

I might be mistaken but my impression when I last checked was that for macOS, we only ship `mne_*` binaries, no FS stuff.

larsoner:
Could be, if that's the case then don't worry about it

docs/source/changes.md
```diff
@@ -25,6 +25,11 @@ authors:
 - Add possibility to exclude runs from the analysis via the new
   [`exclude_runs`][config.exclude_runs] setting.
   ({{ gh(370) }} by {{ authors.crsegerie }})
+- We now create high-resolution scalp surfaces for visualization of the
+  coregistration in the reports. Existing surfaces can be force-regenerated by
+  setting the new [`recreate_scalp_surfaces`][config.recreate_scalp_surfaces]
```
hoechenberger, line 30:
```suggestion
  setting the new [`recreate_scalp_surface`][config.recreate_scalp_surface]
```

tests/configs/config_ds000248.py
```diff
@@ -24,6 +24,7 @@
 
 bem_mri_images = 'FLASH'
 recreate_bem = True
+recreate_scalp_surfaces = True
```
hoechenberger, line 27:
```suggestion
recreate_scalp_surface = True
```

hoechenberger:
It's working! I've update the set of minimal commands on OSF and limited surface creation for the report to the dense surface only. This is the result:

https://8206-167957354-gh.circle-artifacts.com/0/site/examples/ds000248/sub-01_task-audiovisual_report.html#17


Good to merge from my end!

agramfort (approved):
Merge if happy. Maybe revert the changes to CI script ?

agramfort:
👍

PR #377, opened 2021-06-10
WIP : use dataclass for config

agramfort:
### Before merging …

- [ ] Changelog has been updated (`docs/source/changes.md`)

PR #383, opened 2021-06-14
Add per-script configuration objects

agramfort:
Closes #377

### Before merging …

### DIFF
--- a/scripts/init/00-init_derivatives_dir.py
+++ b/scripts/init/00-init_derivatives_dir.py
@@ -2,6 +2,7 @@ import logging
 import itertools
 from typing import Optional
 
+from mne.utils import BunchConst
 from mne.parallel import parallel_func
 from mne_bids.config import BIDS_VERSION
 from mne_bids.utils import _write_json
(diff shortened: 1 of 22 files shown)


hoechenberger:
Report step is missing, otherwise it seems we're green.

scripts/init/00-init_derivatives_dir.py
```diff
@@ -12,63 +13,79 @@
...
 
 
+def get_config(
+    subject: Optional[str] = None,
+    session: Optional[str] = None
+) -> BunchConst:
+    cfg = BunchConst(
+        datatype=config.get_datatype(),
+        subjects=config.get_subjects(),
+        sessions=config.get_sessions(),
```
agramfort, line 64: needed?

scripts/preprocessing/04a-run_ica.py
```diff
@@ -300,14 +310,10 @@ def run_ica(subject, session=None):
         epochs = make_epochs(
             raw=raw,
             event_id=event_id,
-            tmin=config.epochs_tmin,
-            tmax=config.epochs_tmax,
-            metadata_tmin=config.epochs_metadata_tmin,
-            metadata_tmax=config.epochs_metadata_tmax,
-            metadata_keep_first=config.epochs_metadata_keep_first,
-            metadata_keep_last=config.epochs_metadata_keep_last,
```
agramfort, line 308: you removed this on purpose?

scripts/sensor/03-time_frequency.py
```diff
@@ -53,21 +50,26 @@ def run_time_frequency(subject, session=None):
...
-        if 'eeg' in config.ch_types and config.eeg_reference == 'average':
+        if 'eeg' in cfg.ch_types and cfg.eeg_reference == 'average':
             epochs.set_eeg_reference('average')
         else:
             epochs.apply_proj()
-        epochs.pick(config.analyze_channels)
+        epochs.pick(cfg.analyze_channels)
+
+    freqs = np.arange(config.time_frequency_freq_min,
+                      config.time_frequency_freq_max)
```
agramfort, line 63:
```suggestion
    freqs = np.arange(cfg.time_frequency_freq_min,
                      cfg.time_frequency_freq_max)
```

scripts/sensor/03-time_frequency.py
```diff
@@ -77,15 +79,37 @@ def run_time_frequency(subject, session=None):
...
+        datatype=config.get_datatype(),
+        acq=config.acq,
+        rec=config.rec,
+        space=config.space,
+        deriv_root=config.get_deriv_root(),
+        time_frequency_conditions=config.time_frequency_conditions,
+        analyze_channels=config.analyze_channels,
+        spatial_filter=config.spatial_filter,
+        ch_types=config.ch_types,
+        eeg_reference=config.eeg_reference,
```
agramfort: you miss time_frequency_freq_min and time_frequency_freq_max

scripts/sensor/04-group_average.py
```diff
@@ -175,19 +177,47 @@ def average_decoding(session):
         del contrast_score_stats, fname_out
 
 
+def get_config(
+    subject: Optional[str] = None,
+    session: Optional[str] = None
+) -> BunchConst:
+    cfg = BunchConst(
```
agramfort, line 184: here you don't need to pass subject I think as it's a group script but you need subjects and certainly session

scripts/init/00-init_derivatives_dir.py
hoechenberger: Forgot to update this script; fixed now

scripts/preprocessing/04a-run_ica.py
hoechenberger: Yes, we only need metadata when creating the non-ICA epochs (step 3)

scripts/sensor/03-time_frequency.py
hoechenberger: Fixed

scripts/sensor/03-time_frequency.py
hoechenberger: Fixed

scripts/sensor/04-group_average.py
hoechenberger: I've  now added `subjects = config.get_subjects()` to the `Bunch`; `session` is not necessary here due to the (weird?) way we're operating in `main()`. Should probably be refactored…

hoechenberger:
@agramfort

We're green.

The change we discussed regarding `get_runs()` also accepting a session is a bit more complicated than I'd initially thought (has a bunch of side-effects), so let's do this in a followup PR – the diff here is already way too huge for any sane mind to handle during this head wave! 🥵

agramfort:
thx @hoechenberger

PR #386, opened 2021-06-17
ENH: Add automated break annotation

hoechenberger:
### Before merging …

### DIFF
--- a/config.py
+++ b/config.py
@@ -382,6 +382,106 @@ itself, nor to the source analysis stage.
 """
 
 ###############################################################################
+# BREAK DETECTION
+# ---------------
+
+find_breaks: bool = True
+"""
+During an experimental run, the recording might be interrupted by breaks of
+various durations, e.g. to allow the participant to stretch, blink, and swallow
+freely. During these periods, large-scale artifacts are often picked up by the
+recording system. These artifacts can impair certain stages of processing, e.g.
+the peak-detection algorithms we use to find EOG and ECG activity. In some
+cases, even the bad channel detection algorithms might not function optimally.
+It is therefore advisable to mark such break periods for exclusion at early
+processing stages.
+
+If `True`, try to mark breaks by finding segments of the data where no
+exprimental events have occurred. This will then add annotations with the
+description `BAD_break` to the continuous data, causing these segments to be
+ignored in all following processing steps.
+
+???+ example "Example"
+    Automatically find break periods, and annotate them as `BAD_break`.
+    ```python
+    find_breaks = True
+    ```
+
+    Disable break detection.
+    ```python
+    find_breaks = False
+    ```
+"""
+
+min_break_duration: float = 15.
+"""
+The minimal duration (in seconds) of a data segment without any experimental
+events for it to be considered a "break". Note that the minimal duration of the
+generated `BAD_break` annotation will typically be smaller than this, as by
+default, the annotation will not extend across the entire break.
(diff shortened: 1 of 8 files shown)


config.py
```diff
@@ -381,6 +381,105 @@
     ```
 """
 
+###############################################################################
+# BREAK DETECTION
+# ---------------
+
+find_breaks: bool = True
+"""
+During an experimental run, recording might be interrupted by breaks of various
```
hoechenberger, line 390:
```suggestion
During an experimental run, recordings might be interrupted by breaks of various
```

config.py
```diff
@@ -381,6 +381,105 @@
...
+durations, e.g. to allow the participant to stretch, blink, and swallow freely.
+During these periods, large-scale artifacts are often picked up by the
+recording system. These artifacts can impair certain stages of processing, e.g.
+the peak-detection algorithms we use to find EOG and ECG activity. In some
+cases, even the bad channel detection algorithms might not function optimally.
+It is therefore advisable to mark such break periods for exclusion.
```
hoechenberger, line 396:
```suggestion
It is therefore advisable to mark such break periods for exclusion early during processing.
```

config.py
```diff
@@ -381,6 +381,105 @@
...
+description `BAD_break` to the continuous data, causing these segments to be
+ignored in all following processing steps.
+
+???+ example "Example"
+    Automatically find break periods, and annotate them as `BAD_break`.
+    ```python
+    find_breaks = True
+    ```
+
+    Disable break detection
```
hoechenberger, line 409:
```suggestion
    Disable break detection.
```

config.py
```diff
@@ -381,6 +381,105 @@
...
+    ```python
+    find_breaks = False
+    ```
+"""
+
+min_break_duration: float = 15.
+"""
+The minimal duration of a data segment without any experimental events for it
+to be considered a "break", in seconds. Note that the minimal duration of the
```
hoechenberger, line 418:
```suggestion
The minimal duration (in seconds) of a data segment without any experimental events for it
to be considered a "break". Note that the minimal duration of the
```

config.py
```diff
@@ -381,6 +381,105 @@
...
+    ```
+"""
+
+t_break_annot_start_after_previous_event: float = 5.
+"""
+Once a break of at least [`min_break_duration`][configmin_break_duration]
+seconds has been discovered, we generate a `BAD_break` annotation that does not
+necessarily span the entire break period. Instead, you will typically want to
+start it some time after the last event before the break period, as to not
+unnecessarily discard brain activity immediately following that last event.
```
hoechenberger, line 439:
```suggestion
unnecessarily discard brain activity immediately following last event.
```

config.py
```diff
@@ -381,6 +381,105 @@
...
+
+This parameter controls how much time in seconds should pass after the last
```
hoechenberger, line 441:
```suggestion
This parameter controls how much time (in seconds) should pass after the last
```

hoechenberger:
@agramfort Any ideas for shorter parameter names? 🤔

agramfort (approved):
LGTM !

hoechenberger:
@agramfort Style check fails because lines are getting too long due to `t_break_annot_start_after_previous_event` and `t_break_annot_stop_before_next_event`. But honestly I cannot think of anything more concise that would still be descriptive.

agramfort:
use # noqa ?

>

PR #387, opened 2021-06-17
Allow using fsaverage as a template MRI, e.g. for EEG studies without MR scans

agramfort:
cc @apmellot @dengemann

### DIFF
--- a/tests/configs/config_ds000248_no_mri.py
+++ b/tests/configs/config_ds000248_no_mri.py
@@ -15,4 +15,4 @@ ch_types = ['meg']
 use_maxwell_filter = False
 process_er = False
 
-use_mri_template = True
+use_template_mri = True
--- a/tests/configs/config_ds000248_no_mri.py
+++ b/tests/configs/config_ds000248_no_mri.py
@@ -1,5 +1,5 @@
 """
-MNE Sample Data: Using the `fsaverage` template MRI.
+MNE Sample Data: Using the `fsaverage` template MRI
 """
 
 study_name = 'ds000248'
--- a/tests/configs/config_ds000248_no_mri.py
+++ b/tests/configs/config_ds000248_no_mri.py
@@ -1,10 +1,9 @@
 """
-MNE Sample Data
+MNE Sample Data: Using the `fsaverage` template MRI.
 """
 
 study_name = 'ds000248'
-bids_root = '~/mne_data/ds000248'
-deriv_root = '~/mne_data/ds000248/derivatives/mne-bids-pipeline-no-mri'
+bids_root = '~/mne_data/ds000248_no_mri'
 
 subjects = ['01']
 rename_events = {'Smiley': 'Emoji',
--- a/tests/configs/config_ds000248_no_mri.py
+++ b/tests/configs/config_ds000248_no_mri.py
@@ -1,0 +1,19 @@
+"""
+MNE Sample Data
+"""
+
+study_name = 'ds000248'
+bids_root = '~/mne_data/ds000248'
+deriv_root = '~/mne_data/ds000248/derivatives/mne-bids-pipeline-no-mri'
+
+subjects = ['01']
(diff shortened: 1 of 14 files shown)


hoechenberger (changes requested):
Looks quite good already! Nice!
Unfortunately CI is broken b/c of OpenNeuro problems.
@agramfort If the `ds000248_no_mri` test passes locally for you, we may be able to move ahead anyway and merge. But before that, there are a few things missing:

- [ ] changelog entry
- [x] entry in `datasets.py`
- [x] `run_tests.py` needs to be updated
- [x] Circle config needs to be updated
- [x] The new config option should be added to `docs/source/settings/source/bem.md`

hoechenberger:
@agramfort I think I'd like to rename the param from `use_mri_template` to `use_template_mri`, WDYT?

tests/configs/config_ds000248_no_mri.py
```diff
@@ -0,0 +1,19 @@
+"""
+MNE Sample Data
+"""
+
+study_name = 'ds000248'
+bids_root = '~/mne_data/ds000248'
+deriv_root = '~/mne_data/ds000248/derivatives/mne-bids-pipeline-no-mri'
```
agramfort, line 7: should I remove this line now @hoechenberger?

tests/datasets.py
```diff
@@ -109,6 +109,14 @@ class DATASET_OPTIONS_T(TypedDict):
             'derivatives/freesurfer/subjects/sub-01/mri/aparc.a2009s+aseg.mgz'
         ],
     },
+    'ds000248_no_mri': {
+        'git': '',
+        'openneuro': 'ds000248',
+        'osf': '',
+        'web': '',
+        'include': ['derivatives/freesurfer/subjects/sub-01'],
+        'exclude': ['derivatives/freesurfer'],
```
agramfort, line 118: can you also exclude the "anat" folder?

agramfort (commented):
for @hoechenberger

agramfort:
fine with use_template_mri

>

tests/datasets.py
hoechenberger: The current combination of include and exclude that I've added here actually doesn't make sense … We only need `derivatives/freesurfer`, right?? Will update

tests/datasets.py
```diff
@@ -114,8 +114,8 @@ class DATASET_OPTIONS_T(TypedDict):
         'openneuro': 'ds000248',
         'osf': '',
         'web': '',
-        'include': ['derivatives/freesurfer/subjects/sub-01'],
-        'exclude': ['derivatives/freesurfer'],
+        'include': ['derivatives/freesurfer'],
```
agramfort, line 117: We don’t need any freesurfer. Fsaverage should be fetched automatically by the pipeline

hoechenberger: Oh, so … what do we even need then? 😅 just the MEG?

hoechenberger: Changed it to downloading only the MEG for sub-01 now, and omitting the anat

hoechenberger:
@agramfort It appears the fsaverage download doesn't work as intended (or not at all?), hence the test fails. I can look into this later today, unless you immediately have an idea as to where things might be going wrong :)

agramfort:
I am on it

agramfort:
should be fixed @hoechenberger

hoechenberger (approved):

hoechenberger:
Thanks @agramfort, let's see if the artifact production works now as expected, and then proceed to merging.

hoechenberger:
It doesn't show up in the examples gallery, but I'm going ahead and will be merging now, and will fix the gallery later (I think it's really just the TOC entry that is missing!)

hoechenberger:
Ok I think I resolved it. Let's wait for CI to go green

hoechenberger:
🥳 Thanks @agramfort!

PR #392, opened 2021-06-21
ENH: eog/ecg n comp

dengemann:
@agramfort @hoechenberger here is the first portion


agramfort (approved):

hoechenberger:
We should add a test case. Will do.

hoechenberger:
I figured that the parameters `average_ecg_projs` and `average_eog_projs` and their respective descriptions:
```
 Whether to average [ECG,EOG] proejction vectors or not.
```
were a bit misleading, because according to the MNE docs, the `average` parameter determines whether the SSP vectors are to be calculated on epochs (`average=False`) or evoked data (`average=True`).

I, therefore, have renamed the parameters and adjusted their docstrings:

```python
ecg_proj_from: Literal['epochs', 'evoked'] = 'evoked'
"""
Whether to calculate the ECG proejction vectors based on the evoked ECG signal
 (i.e., the averaged ECG epochs), or on the ECG epochs.
"""
```

Okay with you, @agramfort and @dengemann?

`ds000248` test passes locally for me.

config.py
```diff
@@ -903,6 +903,73 @@
...
+"""
+
+n_proj_ecg: Dict[str, float] = dict(n_mag=1, n_grad=1, n_eeg=1)
+"""
+Number of SSP vectors to create for ECG artifacts for each channel type.
+"""
+
+ecg_proj_from: Literal['epochs', 'evoked'] = 'evoked'
+"""
+Whether to calculate the ECG proejction vectors based on the evoked ECG signal
```
agramfort, line 922:
```suggestion
Whether to calculate the ECG projection vectors based on the evoked ECG signal
```

config.py
```diff
@@ -903,6 +903,73 @@
...
+(i.e., the averaged ECG epochs), or on the ECG epochs.
+"""
+
+eog_proj_from: Literal['epochs', 'evoked'] = 'evoked'
+"""
+Whether to calculate the EOG proejction vectors based on the evoked EOG signal
```
agramfort, line 928:
```suggestion
Whether to calculate the EOG projection vectors based on the evoked EOG signal
```

config.py
```diff
@@ -1616,6 +1683,15 @@ def check_baseline(
 check_baseline(baseline=baseline, epochs_tmin=epochs_tmin,
                epochs_tmax=epochs_tmax)
 
+# SSP projectors baswed on evoked or epochs
+if ecg_proj_from not in ('evoked', 'epochs'):
+    raise ValueError(f'ecg_proj_from must be one of: evoked, epochs; but got: '
```
agramfort, line 1688:
```suggestion
    raise ValueError(f'ecg_proj_from must be one of: "evoked", "epochs"; but got: '
```

config.py
```diff
@@ -1616,6 +1683,15 @@ def check_baseline(
...
+                     f'{ecg_proj_from} instead.')
+
+if eog_proj_from not in ('evoked', 'epochs'):
+    raise ValueError(f'eog_proj_from must be one of: evoked, epochs; but got: '
```
agramfort, line 1692:
```suggestion
    raise ValueError(f'eog_proj_from must be one of: "evoked", "epochs"; but got: '
```

tests/configs/config_ds000248_rest.py
```diff
@@ -0,0 +1,34 @@
+"""
+MNE Sample Data
+"""
```
agramfort, line 3: this file should not be in this PR

hoechenberger: Thanks, removed

tests/configs/config_ds000248.py
```diff
@@ -23,6 +23,10 @@
 noise_cov = 'emptyroom'
 
 spatial_filter = 'ssp'
+n_proj_eog = dict(n_mag=1, n_grad=1, n_eeg=1)
+n_proj_ecg = dict(n_mag=1, n_grad=1, n_eeg=0)
+ecg_proj_from = 'evoked'
+eog_proj_from = 'epochs'
```
dengemann, line 29: I find this very unintuitive @hoechenberger @agramfort -- if we average or not does not depend on on whether it's epochs or evoked does it? I would find a simple average parameter more intuitive, matching the MNE API.

dengemann:
> I figured that the parameters `average_ecg_projs` and `average_eog_projs` and their respective descriptions:
> 
> ```
>  Whether to average [ECG,EOG] proejction vectors or not.
> ```
> 
> were a bit misleading, because according to the MNE docs, the `average` parameter determines whether the SSP vectors are to be calculated on epochs (`average=False`) or evoked data (`average=True`).
> 
> I, therefore, have renamed the parameters and adjusted their docstrings:
> 
> ```python
> ecg_proj_from: Literal['epochs', 'evoked'] = 'evoked'
> """
> Whether to calculate the ECG proejction vectors based on the evoked ECG signal
>  (i.e., the averaged ECG epochs), or on the ECG epochs.
> """
> ```
> 
> Okay with you, @agramfort and @dengemann?
> 
> `ds000248` test passes locally for me.

I think I'm -1 on the naming, it is confusing at multiple levels. I would not call link averaging in this context to epochs VS evoked.

tests/configs/config_ds000248.py
agramfort: @dengemann you would prefer ecg_proj_epochs_average= bool ?

Personally I prefer typing True or False but no strong feeling. A book is also closer to MNE API.

hoechenberger:
@dengemann 
I've addressed this in cabb799, better now?

tests/configs/config_ds000248.py
dengemann: yes I would. For the same reasons, hence, my proposal.

agramfort:
honestly I would prefer 

eog_proj_from_average = True | False
ecg_proj_from_average = True | False

hoechenberger:
Thanks @dengemann!

PR #396, opened 2021-06-22
ENH: add autoreject logic to SSP

dengemann:
@hoechenberger @agramfort :)

### Before merging …

### DIFF
--- a/config.py
+++ b/config.py
@@ -1999,13 +1999,17 @@ def get_reject(
     return _get_reject(reject=reject, ch_types=ch_types, epochs=epochs)
 
 
-def get_ssp_reject(ssp_type, epochs) -> Dict[str, float]:
+def get_ssp_reject(
+    *,
+    ssp_type: Literal['ecg', 'eog'],
+    epochs: mne.BaseEpochs
+) -> Dict[str, float]:
     if ssp_type == 'ecg':
         reject = ssp_reject_ecg
     elif ssp_type == 'eog':
         reject = ssp_reject_eog
     else:
-        raise ValueError("Only 'eog'and 'ecg' are supported.")
+        raise ValueError("Only 'eog' and 'ecg' are supported.")
     return _get_reject(reject=reject, ch_types=ch_types, epochs=epochs)
 
 
--- a/config.py
+++ b/config.py
@@ -952,8 +952,12 @@ otherwise, SSP won't be able to "see" these artifacts.
     ```
 """
 
-ssp_reject_eog: Optional[Union[Dict[str, float],
-                               Literal['autoreject_global']]] = None
+ssp_reject_eog: Optional[
+    Union[
+        Dict[str, float],
+        Literal['autoreject_global']
+    ]
+] = None
 """
 Peak-to-peak amplitude limits of the EOG epochs to exclude from SSP fitting.
 This allows you to remove strong transient artifacts, which could negatively
--- a/config.py
+++ b/config.py
@@ -929,7 +929,8 @@ Whether to calculate the EOG projection vectors based on the the averaged or
 on individual EOG epochs.
 """
 
-ssp_reject_ecg: Optional[Dict[str, float]] = None
+ssp_reject_ecg: Union[Optional[Dict[str, float]],
+                      Optional[Literal['autoreject_global']]] = None
 """
 Peak-to-peak amplitude limits of the ECG epochs to exclude from SSP fitting.
 This allows you to remove strong transient artifacts, which could negatively
@@ -947,7 +948,8 @@ otherwise, SSP won't be able to "see" these artifacts.
     ```
 """
 
-ssp_reject_eog: Optional[Dict[str, float]] = None
+ssp_reject_eog: Optional[Union[Dict[str, float],
+                               Literal['autoreject_global']]] = None
 """
 Peak-to-peak amplitude limits of the EOG epochs to exclude from SSP fitting.
 This allows you to remove strong transient artifacts, which could negatively
--- a/config.py
+++ b/config.py
@@ -1999,6 +1999,16 @@ def get_reject(
     return _get_reject(reject=reject, ch_types=ch_types, epochs=epochs)
 
 
+def get_ssp_reject(ssp_type, epochs) -> Dict[str, float]:
+    if ssp_type == 'ecg':
+        reject = ssp_reject_ecg
+    elif ssp_type == 'eog':
+        reject = ssp_reject_eog
+    else:
+        raise ValueError("Only 'eog'and 'ecg' are supported.")
+    return _get_reject(reject=reject, ch_types=ch_types, epochs=epochs)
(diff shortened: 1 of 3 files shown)


config.py
```diff
@@ -929,7 +929,8 @@
 on individual EOG epochs.
 """
 
-ssp_reject_ecg: Optional[Dict[str, float]] = None
+ssp_reject_ecg: Union[Optional[Dict[str, float]],
+                      Optional[Literal['autoreject_global']]] = None
```
hoechenberger, line 933:
```suggestion
ssp_reject_ecg: Optional[
    Union[
        Dict[str, float],
        Literal['autoreject_global']
    ]
] = None
```

config.py
```diff
@@ -947,7 +948,8 @@
     ```
 """
 
-ssp_reject_eog: Optional[Dict[str, float]] = None
+ssp_reject_eog: Optional[Union[Dict[str, float],
+                               Literal['autoreject_global']]] = None
```
hoechenberger, line 952:
```suggestion
ssp_reject_eog: Optional[
    Union[
        Dict[str, float],
        Literal['autoreject_global']
    ]
] = None
```

scripts/preprocessing/04b-run_ssp.py
```diff
@@ -52,14 +53,17 @@ def run_ssp(cfg, subject, session=None):
...
     raw = mne.io.read_raw_fif(raw_fname_in)
-    # XXX : n_xxx should be options in config
     msg = 'Computing SSPs for ECG'
     logger.debug(gen_log_message(message=msg, step=4, subject=subject,
                                  session=session))
 
+    reject_ecg_ = cfg._get_reject(
+        reject=cfg.ssp_reject_ecg,
+        epochs=(create_ecg_epochs(raw)
+                if cfg.ssp_reject_ecg == 'autoreject_global' else None))
```
hoechenberger, line 63: I'd suggest to add a new public function `get_ssp_reject()` to `config.py`, and calling that outside of the processing function when creating `cfg`:

```python
cfg = BunchConst(...,
    ssp_reject_ecg=config.get_reject_ssp(which='ecg')
)
```

hoechenberger (changes requested):

scripts/preprocessing/04b-run_ssp.py
dengemann: Not sure I understand this reasoning

hoechenberger: We recently switched to the script-local cfg objects that should contain all the settings required for that particular script (and only those)

Anything "constant" should be added to that object before running the processing functions, which will receive `cfg` as a parameter

cc @agramfort

hoechenberger: Reason is we've started to try and make our functions as "pure" as possible to avoid a total mess of side-effects

dengemann: but you have that bunch constant already there and it is used, what is the point?

dengemann: What you suggest is less readable than the current line defining the bunch constant; on the other hand we need the logic in the lines you commented on to pass the ecg/eog epochs computed from the raw, nothing that can be handled at config level.

config.py
```diff
@@ -1989,6 +1999,16 @@ def get_reject(
     return _get_reject(reject=reject, ch_types=ch_types, epochs=epochs)
 
 
+def get_ssp_reject(ssp_type, epochs) -> Dict[str, float]:
```
hoechenberger, line 2002:
```suggestion
def get_ssp_reject(
    *,
    ssp_type: Literal['ecg', 'eog'],
    epochs: mne.BaseEpochs
) -> Dict[str, float]:
```

config.py
```diff
@@ -1989,6 +1999,16 @@ def get_reject(
...
+    if ssp_type == 'ecg':
+        reject = ssp_reject_ecg
+    elif ssp_type == 'eog':
+        reject = ssp_reject_eog
+    else:
+        raise ValueError("Only 'eog'and 'ecg' are supported.")
```
hoechenberger, line 2008:
```suggestion
        raise ValueError("Only 'eog' and 'ecg' are supported.")
```

scripts/preprocessing/04b-run_ssp.py
```diff
@@ -52,14 +53,17 @@ def run_ssp(cfg, subject, session=None):
         raw_fname_in.update(split='01')
 
     raw = mne.io.read_raw_fif(raw_fname_in)
-    # XXX : n_xxx should be options in config
     msg = 'Computing SSPs for ECG'
     logger.debug(gen_log_message(message=msg, step=4, subject=subject,
                                  session=session))
 
+    reject_ecg_ = cfg.get_ssp_reject(
```
hoechenberger, line 60:
```suggestion
    reject_ecg_ = config.get_ssp_reject(
```
Why the underscore?

scripts/preprocessing/04b-run_ssp.py
```diff
@@ -75,9 +79,13 @@ def run_ssp(cfg, subject, session=None):
     else:
         ch_names = None
 
+    reject_eog_ = cfg.get_ssp_reject(
```
hoechenberger, line 82:
```suggestion
    reject_eog_ = config.get_ssp_reject(
```
Why the underscore?

scripts/preprocessing/04b-run_ssp.py
agramfort: to me the underscore suggest is a temp / local variable just used below. It's just a visual hint

agramfort (approved):
just a changelog entry then for me 

thx @dengemann !

scripts/preprocessing/04b-run_ssp.py
```diff
@@ -52,14 +53,17 @@ def run_ssp(cfg, subject, session=None):
         raw_fname_in.update(split='01')
 
     raw = mne.io.read_raw_fif(raw_fname_in)
-    # XXX : n_xxx should be options in config
     msg = 'Computing SSPs for ECG'
     logger.debug(gen_log_message(message=msg, step=4, subject=subject,
                                  session=session))
 
+    reject_ecg_ = config.get_ssp_reject(
```
hoechenberger, line 60: Why the underscore?

scripts/preprocessing/04b-run_ssp.py
```diff
@@ -75,9 +79,13 @@ def run_ssp(cfg, subject, session=None):
     else:
         ch_names = None
 
+    reject_eog_ = config.get_ssp_reject(
```
hoechenberger, line 82: Why the underscore?

hoechenberger (approved):
2 nitpicks, otherwise LGTM

hoechenberger:
Thanks @dengemann

PR #402, opened 2021-06-23
Improve logging

hoechenberger:


scripts/preprocessing/02-frequency_filter.py
```diff
@@ -177,12 +177,12 @@ def filter_data(
 
         raw_er.load_data()
         filter(
-            raw=raw_er, subject=subject, session=session, run=run,
+            raw=raw_er, subject='emptyroom', session=session, run=run,
```
hoechenberger, line 180: `subject` here is only used for logging

scripts/preprocessing/02-frequency_filter.py
```diff
@@ -177,12 +177,12 @@ def filter_data(
...
             h_freq=cfg.h_freq, l_freq=cfg.l_freq,
             h_trans_bandwidth=cfg.h_trans_bandwidth,
             l_trans_bandwidth=cfg.l_trans_bandwidth
         )
-        resample(raw=raw_er, subject=subject, session=session, run=run,
+        resample(raw=raw_er, subject='emptyroom', session=session, run=run,
```
hoechenberger, line 185: `subject` here is only used for logging

PR #403, opened 2021-06-23
When retrieving BIDS entity values, limit query to the data type we're actually processing

hoechenberger:
Fix for @dengemann

### Before merging …

### DIFF
--- a/config.py
+++ b/config.py
@@ -24,8 +24,8 @@ import numpy as np
 import pandas as pd
 import json_tricks
 import mne
-from mne_bids import BIDSPath, read_raw_bids
-from mne_bids.path import get_entity_vals
+import mne_bids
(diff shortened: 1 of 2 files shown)


config.py
```diff
@@ -1757,10 +1757,13 @@ def get_deriv_root() -> pathlib.Path:
 def get_subjects() -> List[str]:
     global subjects
 
-    env = os.environ
-
-    valid_subjects = get_entity_vals(get_bids_root(), entity_key='subject')
+    all_datatypes = mne_bids.get_datatypes(root=get_bids_root())
+    ignore_datatypes = set(all_datatypes) - set(get_datatype())
```
dengemann, line 1761: yesss! exactly what I was thinking

PR #414, opened 2021-07-07
Decoding in time-frequency space using CSP + Permutation statistics

crsegerie:
This PR adds a new script to the pipeline : `03b_time_frequency_csp.py`,  designed to analyze in the time-frequency domain a contrast between two conditions.

There are two main steps in this script:
1. **Decoding for each time-frequency bin**: For each time-frequency bin, we use a csp classifier in order to distinguish between the two conditions. We compute the roc-auc score for each time-frequency bin.
2. **Permutation statistics at the group level**: We try to answer the following question: is the difference between the two conditions statistically significant? We use the classic permutations cluster tests on the time-frequency roc-auc map.

During the script, we create a new type of reports: `*-proc-csp+permutation+test_report.html`, that contain the following figures :

- Frequency roc-auc decoding:
image

- Time-frequency decoding
image

- Permutation statistics
image

CI Test: WIP
The code was tested in the working memory time study with 20 subjects. [The code also passed the CI tests, although the majority of the CI test is not designed to test this type of code, specifically designed for group studies. WIP]

WIP:
- [x] Change log entry
- [x] Run time optimization
- [x] Reports

Coding style: In the code, I use two utils class : Tf, and Pth, each one dealing with time-frequency useful information, and paths management. This is a bit of a disruption because there are no such class in the rest of the pipeline, but they have enabled to DRY up the code significantly.


config.py
```diff
@@ -1201,6 +1201,26 @@
...
+"""
+Statistic level used in the time frequency script.
+"""
+
+alpha_t_test: float = 0.1
+"""
+Threshold level used to determine clusters.
+"""
+
+n_permutations: int = 10000
```
agramfort, line 1219:
```suggestion
n_permutations: int = 10_000
```

config.py
```diff
@@ -1201,6 +1201,26 @@
 confidence interval of the mean decoding score.
 """
 
+reg: float = 0.1
```
agramfort, line 1204:
```suggestion
csp_reg: float = 0.1
```

config.py
```diff
@@ -1201,6 +1201,26 @@
...
+"""
+Regularization used in the covariance estimator when calculating CSPs.
+"""
+
+alpha: float = 0.05
```
agramfort, line 1209:
```suggestion
cluster_stats_alpha: float = 0.05
```

config.py
```diff
@@ -1201,6 +1201,26 @@
...
+"""
+Statistic level used in the time frequency script.
+"""
+
+alpha_t_test: float = 0.1
```
agramfort, line 1214:
```suggestion
cluster_stats_alpha_t_test: float = 0.1
```

config.py
```diff
@@ -1250,6 +1270,19 @@
     ```
 """
 
+n_freqs: int = 5
+"""
+The number of frequencies used for decoding using time frequency widows.
```
agramfort, line 1275:
```suggestion
The number of frequencies used for decoding using time frequency windows.
```
is this really needed?

scripts/sensor/03b-time_frequency_csp.py
```diff
@@ -0,0 +1,866 @@
...
+https://mne.tools/stable/auto_tutorials/machine-learning/50_decoding.html#common-spatial-pattern
+Warning: This step, especially the double loop on the time-frequency bins
+is very computationally expensive.
+
+- 2. Permutation statistics
+We try to answer the following question: is the difference between
+the two conditions statistically significant? We use the classic permutations
+cluster tests on the time-frequency roc-auc map.
+More details are available here:
+https://mne.tools/stable/auto_tutorials/stats-sensor-space/10_background_stats.html#sphx-glr-auto-tutorials-stats-sensor-space-10-background-stats-py
```
agramfort, line 22:
```suggestion
https://mne.tools/stable/auto_tutorials/stats-sensor-space/10_background_stats.html
```

scripts/sensor/03b-time_frequency_csp.py
```diff
@@ -0,0 +1,866 @@
...
+) -> Tuple[BaseEpochs, np.ndarray]:
+    """Band-pass and clean the epochs and prepare labels.
+
+    Returns:
+    --------
+    epochs_filter, y
+    """
+    # Prepare epoch_filter
+    epochs_filter = epochs.copy()
+    epochs_filter.decimate(decim=cfg.decim)
```
agramfort, line 266: decim should already have been applied to clean epochs I think

scripts/sensor/03b-time_frequency_csp.py
```diff
@@ -0,0 +1,866 @@
...
+    epochs_filter.filter(fmin, fmax, n_jobs=1, fir_design='firwin',
+                         skip_by_annotation='edge')
+    epochs_filter.drop_bad()
```
agramfort, line 269: it should already be clean so no need of drop_bad

scripts/sensor/03b-time_frequency_csp.py
```diff
@@ -0,0 +1,866 @@
...
+    """Band-pass and clean the epochs and prepare labels.
+
+    Returns:
+    --------
+    epochs_filter, y
+    """
+    # Prepare epoch_filter
+    epochs_filter = epochs.copy()
+    epochs_filter.decimate(decim=cfg.decim)
+    epochs_filter.filter(fmin, fmax, n_jobs=1, fir_design='firwin',
```
agramfort, line 267: why specififying the fir_design?

scripts/sensor/03b-time_frequency_csp.py
```diff
@@ -0,0 +1,866 @@
...
+    report.add_figs_to_section(
+        fig,
+        section="Time - frequency decoding",
+        captions=f'sub-{subject}')
+
+    msg = f"Decoding for subject {subject} finished successfully."
+    logger.info(gen_log_message(message=msg, subject=subject, step=3))
+
+
+def load_and_average(
```
agramfort, line 525: why do you need this?

scripts/sensor/03b-time_frequency_csp.py
```diff
@@ -0,0 +1,866 @@
...
+
+    # Calculating the 95% confidence intervals
+    # TODO: The confidence intervals indicated on the individual
+    # plot are not quite valid because
+    # they use only the std of the cross-validation scores.
+    # We should also divide by the square root.
+    all_freq_scores_std = load_and_average(
+        pth.freq_scores, subjects=subjects, average=False)
+    std_ = np.nanstd(all_freq_scores_std, axis=0, ddof=1)
+    conf_int = 1.96*std_ / np.sqrt(all_freq_scores_std.shape[0])
```
agramfort, line 702: where did you get this from?

agramfort: You should do the same thing that we do to average time decoding scores.

agramfort (commented):
a first batch of comments

config.py
crsegerie: Yes I think it's needed. The other possibility would be to specify n_freq_windows.
(n_freq_windows = n_freqs - 1)

scripts/sensor/03b-time_frequency_csp.py
crsegerie: Yes exactly, the decim parameter was already applied, but it's possible that to speed up the calculation we want to decim even further. But we should do this without the cfg.decim parameter and do this automatically.

Or... Do you think that by not decimating, we could facilitate the estimation of the covariance ?

scripts/sensor/03b-time_frequency_csp.py
crsegerie: You are right, but it's weird, If I remember correctly, drop_bad has been very important for the numerical stability

scripts/sensor/03b-time_frequency_csp.py
agramfort: don't do this and profile the code to see where time is spent.

config.py
agramfort: I don't get this parametrization. I would say you need a list of tuples with (freq_min, freq_max) for every fred band or allow an array and decode between each value. clear?

scripts/sensor/03b-time_frequency_csp.py
agramfort: I don't understand what you mean by "numerical stability". What is unstable? it should not happen.

scripts/sensor/03b-time_frequency_csp.py
crsegerie: I agree it's not necessary to dump and then load the numpy arrays containing the tf_scores and frequency_scores, but since the function one_subject_decoding takes a long time to compute, and since the group analysis function is very cheap computationally, it's easy to just comment the lines where I launch the one_subject_decoding in order to just recalculate the group statistics by changing the parameters like the tstat value to rapidly iterate with the plotting. Also, At the beginning, it was not clear if we would split the script into 2 scripts between the csp estimation and the permutation test.

hoechenberger:
Thanks @crsegerie 🚀 I will review tomorrow morning 👌

config.py
```diff
@@ -1201,6 +1201,26 @@
 confidence interval of the mean decoding score.
 """
 
+csp_reg: float = 0.1
+"""
+Regularization used in the covariance estimator when calculating CSPs.
```
hoechenberger, line 1206:
```suggestion
Regularization used in the covariance estimator when calculating CSPs. Must be
between 0 and 1.
```

config.py
```diff
@@ -1201,6 +1201,26 @@
...
+"""
+Statistic level used in the time frequency script.
+"""
+
+cluster_stats_alpha_t_test: float = 0.1
+"""
+Threshold level used to determine clusters.
+"""
+
+n_permutations: int = 10_000
```
hoechenberger, line 1219: Ah cool, I'd forgotten about this way to make large numbers more readable! 👌

scripts/sensor/03b-time_frequency_csp.py
crsegerie: image

So the majority of the is spent in the CSP in the covariance estimator.

config.py
```diff
@@ -1201,6 +1201,26 @@
...
+
+cluster_stats_alpha: float = 0.05
+"""
+Statistic level used in the time frequency script.
+"""
+
+cluster_stats_alpha_t_test: float = 0.1
+"""
+Threshold level used to determine clusters.
+"""
```
hoechenberger, line 1217: I would name this param differently, e.g.,

```
cluster_t_dist_alpha_thres: float = 0.1
"""
The percentile of the t-distribution to use as the threshold when creating the
clusters. The t-distribution is created with `number of subjects - 1` degrees
of freedom. For example, if you set this value to `0.05` and your analysis
includes 10 subjects, the value of the 95th percentile of the t-distribution
`t(10)` will be used for thresholding the clusters.
"""
```
@agramfort WDYT?

config.py
```diff
@@ -1250,6 +1270,19 @@
     ```
 """
 
+n_freqs: int = 5
+"""
+The number of frequencies used for decoding using time frequency windows.
+Example: n_freqs = 5 generates 4 frequency bins.
```
hoechenberger, line 1276: I think I, as a user, would prefer to specify the number of frequency bins?

config.py
```diff
@@ -1250,6 +1270,19 @@
...
+"""
+
+n_cycles: int = 1
+"""
+Control the duration of the time windows in the CSP decoding script.
+n_cycle = 1 is just above the Nyquist requirements.
+Increase it in case of numerical instabilities.
+"""
```
hoechenberger, line 1284: As per discussion yesterday, this should probably go away.

docs/source/settings/sensor/statistics.md
```diff
@@ -3,3 +3,7 @@
 ::: config.decoding_metric
 ::: config.decoding_n_splits
 ::: config.n_boot
+::: config.csp_reg,
+::: config.cluster_stats_alpha,
+::: config.cluster_stats_alpha_t_test,
```
hoechenberger, line 8: No commas here please :)

docs/source/settings/sensor/time_freq.md
```diff
@@ -1 +1,3 @@
 ::: config.time_frequency_conditions
+::: config.n_freqs,
+::: n_cycles=config.n_cycles,
```
hoechenberger, line 3: no commas here please :)

hoechenberger:
First small round of reviews, more to come :)

config.py
crsegerie: yep, this is Alexandre's idea!

config.py
crsegerie: yes this was a definitely a bad api... I've deleted that part this morning

config.py
crsegerie: Solved in the last commit

config.py
crsegerie: in all cases, your description is way better than mine

scripts/sensor/03b-time_frequency_csp.py
crsegerie: Yes you were right. The important part is the next line, when we drop Stim channels

scripts/sensor/03b-time_frequency_csp.py
crsegerie: I don't understand from which part of the code I should take inspiration from?

agramfort: here https://github.com/mne-tools/mne-bids-pipeline/blob/main/scripts/sensor/04-group_average.py#L158

crsegerie: integrated with "bootstrap confidence intervals"

scripts/sensor/03b-time_frequency_csp.py
```diff
@@ -0,0 +1,988 @@
...
+    A boolean numpy array containing the labels.
+    """
+    tf_cond = cfg.time_frequency_conditions
+    epochs_cond_0 = epochs[tf_cond[0]]
+    event_id_condition_0 = set(epochs_cond_0.events[:, 2])
+    epochs_cond_1 = epochs[tf_cond[1]]
+    event_id_condition_1 = set(epochs_cond_1.events[:, 2])
+
+    y = epochs.events[:, 2].copy()
+    for i in range(len(y)):
```
hoechenberger, line 208: Style: use `enumerate()` in situations like these

crsegerie: But here I think we cannot spare the usage of brackets. Even with enumerate, we still would have to use y[i], no ?

crsegerie: Because we wave to modify the value, not just read it

crsegerie:
Finally, Green !

crsegerie:
Green, finally !

crsegerie:
Tests details: 

| __name__ | Nsubjects | Nruns | Ncontrasts | ch_types | CSP | CSP Warning |
| --- | --- | --- | --- | --- | --- | --- |
| config_ds000117 | 1 | 2 | 3 | ['meg'] |  | Small warning because of time ordering + Warning because decimation |
| config_ds000246 | 1 | 1 | 1 | ['meg'] |  | MAG dimension 247 reduced manually to 100. |
| config_ds000247 | 1 | Nan | Nan | ['meg'] | Contrast = 0 so CSP not run |  |
| config_ds000248 | 1 | Nan | 2 | ['meg'] |  |  |
| config_ds000248_coreg_surfaces | 1 | Nan | Nan | ['meg'] | Contrast = 0 so CSP not run |  |
| config_ds000248_FLASH_BEM | 1 | Nan | Nan | ['meg'] | Contrast = 0 so CSP not run |  |
| config_ds000248_ica | 1 | Nan | Nan | ['meg'] | Contrast = 0 so CSP not run |  |
| config_ds000248_no_mri | 1 | Nan | 1 | ['meg'] |  |  |
| config_ds000248_T1_BEM | 1 | Nan | Nan | ['meg'] | Contrast = 0 so CSP not run |  |
| config_ds001810 | 5 | Nan | 1 | ['eeg'] |  | (Small warning because of decimation, otherwise perfect) |
| config_ds001971 | 1 | 1 | 1 | ['eeg'] |  |  |
| config_ds003104 | Nan | Nan | Nan | ['meg'] | Contrast = 0 so CSP not run |  |
| config_ds003392 | 1 | Nan | 1 | ['meg'] |  |  |
| config_eeg_matchingpennies | 1 | Nan | 1 | ['eeg'] |  |  |

I made a small script to extract into a csv the configuration settings to check that the new script has been tested in all major configurations.

Runtime: Overall, even though the new script would take a long time to compute naively, with proper decimation and dimension reduction, the script runs in less than ~30 seconds for all tests.

crsegerie:
It's not perfect yet, but I think I can hand the baton over to you.
I hope this script won't contain too many anti-patterns, and that it won't be too hard to merge... but before the end of my internship, I have to move on to other things.

I'm glad I've been able to work on it. Thank you @agramfort for the various optimization tips, thank you @hoechenberger for the peer coding sessions, thank you @SophieHerbst for the study idea!

PR #419, opened 2021-07-10
Enforce specification of subjects_dir if deriv_root is set

### LINKED ISSUE #404: FreeSurfer step doesn't respect deriv_root
Instead, it will write outputs to `bids_root/derivatives/freesurfer` no matter what.

Reported by @dengemann

hoechenberger:
Didn't work properly before: recon-all output was always written to `SUBJECTS_DIR/`.

Fixes #404

### Before merging …

### DIFF
--- a/config.py
+++ b/config.py
@@ -2049,8 +2049,9 @@ def get_ica_reject() -> Dict[str, float]:
 
 def get_fs_subjects_dir():
     if not subjects_dir and deriv_root is not None:
-        # We do this check here to avoid an error message when a user doesn't
-        # intend to run the source analysis steps anyway.
+        # We do this check here (and not in our regular checks section) to
+        # avoid an error message when a user doesn't intend to run the source
+        # analysis steps anyway.
         raise ValueError(
             'When specifying a "deriv_root", you must also supply a '
             '"subjects_dir".'
--- a/config.py
+++ b/config.py
@@ -59,6 +59,10 @@ deriv_root: Optional[PathLike] = None
 The root of the derivatives directory in which the pipeline will store
 the processing results. If ``None``, this will be
 ``derivatives/mne-bids-pipeline`` inside the BIDS root.
+
+Note: Note
+    If specified and you wish to run the source analysis steps, you must
+    set [`subjects_dir`][config.subjects_dir] as well.
 """
 
 subjects_dir: Optional[PathLike] = None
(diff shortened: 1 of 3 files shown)


agramfort:
Not sure about this. In ds117 they ship a free surfer recon but it’s incomplete. Rather than deleted their files I’ve changed the subjects_dir. With this fix I would be stuck

hoechenberger:
I'm not sure I fully understand. Maybe let's have a call tonight if you have time (I'm currently out to have coffee :))

hoechenberger:
Btw motivation was that @dengemann had sets deriv_root that deviated from the default (bids_root/derivatives), yet our FS script would produce the  recon-all output in bids_root/derivatives/freesurfer, which was clearly the wrong place

hoechenberger:
@agramfort I think I got what you meant and have started a new attempt. WDYT?

config.py
```diff
@@ -59,13 +59,26 @@
...
-derivativesfor all subjects. Specifically, the ``subjects_dir`` is the
-$SUBJECTS_DIR used by the Freesurfer software.
+Path to the directory that contains the FreeSurfer parcellations of all
+subjects. Specifically, this defines the ``SUBJECTS_DIR`` that is used by
+FreeSurfer. When running the ``freesurfer`` processing step to create the
+parcellations from anatomical scans in the BIDS dataset, the output will be
+stored in this directory.
+
+If ``None``, this will default
+to [`bids_root`][config.bids_root_root]`/freesurfer/subjects`.
```
hoechenberger, line 77:
```suggestion
to [`bids_root`][config.bids_root_root]`/derivatives/freesurfer/subjects`.
```

hoechenberger:
Ok should be good now.

config.py
```diff
@@ -59,13 +59,26 @@
...
+    If specified and you wish to run the source analysis steps, you must
+    set [`subjects_dir`][config.subjects_dir] as well.
 """
 
 subjects_dir: Optional[PathLike] = None
 """
-Path to the directory that contains the MRI data files and their
-derivativesfor all subjects. Specifically, the ``subjects_dir`` is the
-$SUBJECTS_DIR used by the Freesurfer software.
+Path to the directory that contains the FreeSurfer parcellations of all
```
agramfort, line 70: It’s the « reconstructions » not the parcellations. To fix here and below.

agramfort (commented):
Besides LGTM !

config.py
hoechenberger: Right, thanks!

PR #421, opened 2021-07-11
some fixes when trying to process the MIND dataset

agramfort:
### Before merging …

### DIFF
--- a/config.py
+++ b/config.py
@@ -1985,11 +1985,20 @@ def _get_reject(
         # Automated threshold calculation requested
         import autoreject
 
+        ch_types_autoreject = list(ch_types)
+        if 'meg' in ch_types_autoreject:
+            ch_types_autoreject.remove('meg')
+            if 'mag' in epochs:
+                ch_types_autoreject.append('mag')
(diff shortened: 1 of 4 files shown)


config.py
```diff
@@ -1985,11 +1985,20 @@ def _get_reject(
         # Automated threshold calculation requested
         import autoreject
 
+        ch_types_autoreject = list(ch_types)
+        if 'meg' in ch_types_autoreject:
+            ch_types_autoreject.remove('meg')
+            if 'mag' in epochs:
+                ch_types_autoreject.append('mag')
+            if 'grad' in epochs:
+                ch_types_autoreject.append('grad')
```
agramfort, line 1994: @apmellot this should fix your problem on camcan

hoechenberger: Could you elaborate why this is needed? Doesn't `meg` do the right thing?

agramfort: autoreject does not support "meg"

hoechenberger:
Thanks @agramfort!

apmellot:
@agramfort Thank you !

PR #424, opened 2021-07-12
Bug occuring when l_freq and h_freq = None

apmellot:
When there is no frequency filter applied to the raw file  (so h_freq=None and l_freq=None) no raw "filt" file is created and it creates a bug at the beginning of "make_epochs" when this file is called.


agramfort:
it fixes your pb @apmellot ?

apmellot:
Yes I tested it on a few subjects @agramfort

PR #427, opened 2021-07-13
fix forward for CTF data

### LINKED ISSUE #426: CTF source localization bug
During Step-10 I get an error.  Bids pipeline is trying to find a fif file in the original bids meg directory, but the original dataset is a CTF .ds.  The BIDSPath is also trying to find the -ave.fif file, so the designation of 'extension' during the BIDSPath initiation is ambiguous.

ERROR:
21:36:00 [Step-10][sub-ON26309, ses-01] Estimating head ↔ MRI transform
21:36:01  A critical error occurred.
....
FileNotFoundError: fname does not exist: /fast/BIDS_HV_V1mod/sub-ON26309/ses-01/meg/sub-ON26309_ses-01_task-airpuff_run-01_meg.fif

______________
The file it is looking for is :  /fast/BIDS_HV_V1mod/sub-ON26309/ses-01/meg/sub-ON26309_ses-01_task-airpuff_run-01_meg.**ds**

_______________

The issue appears to arise in - Line 107 of 02-make-forward.py.  The bids_path is expecting a .fif file.  I tested changing this to .ds in the bids_path designation, but it does not resolve the issue - since the source projection is being done on the average file, which has been saved out as a .fif file in the derivatives folder.

```
def run_forward(cfg, subject, session=None):
    bids_path = BIDSPath(subject=subject,
                         session=session,
                         task=cfg.task,
                         acquisition=cfg.acq,
                         run=None,
                         recording=cfg.rec,
                         space=cfg.space,
                         extension='.fif',   #<<<<<<<<<<<
                         datatype=cfg.datatype,
                         root=cfg.deriv_root,
                         check=False)
```
I am assuming that the way to solve this would be to call/instantiate BIDSPath twice, once for the acquisition dataset and once for the derivatives dataset.  

extra info:
mne-bids v 0.8  (commit - 3a4543076912cebbc89a5f0b9433 )
mne-bids-pipeline: (commit - 861aa05c332ea9e215fd1250df4)
OS: ubuntu linux

agramfort:
closes #426

### DIFF
--- a/docs/source/changes.md
+++ b/docs/source/changes.md
@@ -189,3 +189,5 @@ authors:
 - Parameters retrieved from the BIDS dataset were not limited to the modality
   ("data type") we meant to process, sometimes leading to crashes.
   ({{ gh(391) }} by {{ authors.hoechenberger }})
+- Fix forward computation for CTF data.
+  ({{ gh(427) }} by {{ authors.agramfort }})
--- a/scripts/source/02-make_forward.py
+++ b/scripts/source/02-make_forward.py
@@ -61,7 +61,8 @@ def _prepare_forward(cfg, bids_path, fname_trans):
 
     trans = get_head_mri_trans(
         bids_path.copy().update(run=cfg.runs[0],
-                                root=cfg.bids_root),
+                                root=cfg.bids_root,
+                                extension=None),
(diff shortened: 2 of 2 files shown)


agramfort:
thx @jstout211 for the bug report

PR #432, opened 2021-07-15
Fix error occuring when there are not enough ecg or eog epochs for autoreject

apmellot:
In order to work autoreject needs at least 5 epochs. Here, if there are not enough ecg or eog epochs the rejection threshold is set to None.
I tested it on a few subjects and it seems to work :ok_hand:

### DIFF
--- a/config.py
+++ b/config.py
@@ -934,6 +934,17 @@ order to remove the artifacts. The ICA procedure can be configured in various
 ways using the configuration options you can find below.
 """
 
+min_ecg_epochs: int = 5
+"""
+Minimal number of ECG epochs needed to compute SSP or ICA rejection with autoreject.
+"""
+
+min_eog_epochs: int = 5
+"""
+Minimal number of EOG epochs needed to compute SSP or ICA rejection with autoreject.
+"""
+
+
 # Rejection based on SSP
 # ~~~~~~~~~~~~~~~~~~~~~~
 
--- a/config.py
+++ b/config.py
@@ -936,12 +936,12 @@ ways using the configuration options you can find below.
 
 min_ecg_epochs: int = 5
 """
-Minimal number of ECG epochs needed to compute SSP or ICA rejection with autoreject.
+Minimal number of ECG epochs needed to compute SSP or ICA rejection.
 """
 
 min_eog_epochs: int = 5
 """
-Minimal number of EOG epochs needed to compute SSP or ICA rejection with autoreject.
+Minimal number of EOG epochs needed to compute SSP or ICA rejection.
 """
 
 
--- a/scripts/preprocessing/04b-run_ssp.py
+++ b/scripts/preprocessing/04b-run_ssp.py
@@ -78,7 +78,8 @@ def run_ssp(*, cfg, subject, session=None):
                                             **cfg.n_proj_ecg)
 
     if not ecg_projs:
-        msg = 'No ECG events could be found. No ECG projectors computed.'
+        msg = ('Not enough ECG events could be found. No ECG projectors are '
+               'computed.')
         logger.info(**gen_log_kwargs(message=msg, subject=subject,
                                      session=session))
 
@@ -112,7 +113,8 @@ def run_ssp(*, cfg, subject, session=None):
                                             **cfg.n_proj_eog)
 
     if not eog_projs:
-        msg = 'Not enough EOG events could be found. No EOG projectors computed.'
+        msg = ('Not enough EOG events could be found. No EOG projectors are '
+               'computed.')
         logger.info(**gen_log_kwargs(message=msg, subject=subject,
                                      session=session))
 
--- a/scripts/preprocessing/04b-run_ssp.py
+++ b/scripts/preprocessing/04b-run_ssp.py
@@ -57,10 +57,13 @@ def run_ssp(*, cfg, subject, session=None):
     logger.debug(**gen_log_kwargs(message=msg, subject=subject,
                                   session=session))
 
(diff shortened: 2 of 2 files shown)


agramfort:
I will let @dengemann review

scripts/preprocessing/04b-run_ssp.py
```diff
@@ -57,10 +57,14 @@ def run_ssp(*, cfg, subject, session=None):
...
-        epochs=(create_ecg_epochs(raw)
-                if cfg.ssp_reject_ecg == 'autoreject_global' else None))
+    reject_ecg_ = None
+    ecg_epochs = None
+    if cfg.ssp_reject_ecg == 'autoreject_global':
+        ecg_epochs = create_ecg_epochs(raw)
+        if len(ecg_epochs) >= 5:
+            reject_ecg_ = config.get_ssp_reject(
+                ssp_type='ecg',
+                epochs=ecg_epochs)
```
agramfort, line 67: was it done like this in old @dengemann 's code?
First I would expose the param 5 in the config.py as a study parameter.
Second I would have thought that in this case you rather do not compute
any projection. ecg_projs is therefore here just an empty list.

apmellot: In @dengemann's code it is : 
```
    ecg_epochs = mne.preprocessing.create_ecg_epochs(raw)
    if len(ecg_epochs) >= 5:
        reject_ecg = get_rejection_threshold(ecg_epochs, decim=8)
    else:
        reject_ecg = None
```

agramfort: this means compute an SSP (which can be very bad if not enough data). What I suggest to not compute any SSP.

dengemann: yes that is what has to be done (and the old logic did it).

scripts/preprocessing/04b-run_ssp.py
```diff
@@ -79,10 +83,15 @@ def run_ssp(*, cfg, subject, session=None):
...
 
-    reject_eog_ = config.get_ssp_reject(
-        ssp_type='eog',
-        epochs=(create_eog_epochs(raw)
-                if cfg.ssp_reject_eog == 'autoreject_global' else None))
+    reject_eog_ = None
+    eog_epochs = None
+    if cfg.ssp_reject_eog == 'autoreject_global':
+        eog_epochs = create_eog_epochs(raw)
+        if len(eog_epochs) >= 5:
```
dengemann, line 90: How about we introduce an option here? `MIN_SSP_N_EOG` or something similar? Perhaps we want allow larger numbers.

scripts/preprocessing/04b-run_ssp.py
```diff
@@ -79,10 +83,15 @@ def run_ssp(*, cfg, subject, session=None):
...
+            reject_eog_ = config.get_ssp_reject(
+                ssp_type='eog',
+                epochs=eog_epochs)
+
     eog_projs, _ = compute_proj_eog(raw, ch_name=ch_names,
                                     average=cfg.eog_proj_from_average,
                                     reject=reject_eog_,
```
dengemann, line 97: I think this should now move inside a conditional and only get computed if enough events are preset.

scripts/preprocessing/04b-run_ssp.py
hoechenberger: We don't need to restrict this to SSP -> how about `min_eog_epochs : None | float`, and it can be used for SSP and ICA alike?

hoechenberger:
@apmellot Can you please pull in the main branch? We made some optimizations there to stop our CI runs from clogging up CircleCI :)

hoechenberger:
Merci ! :)

Tests are now running for `ds000247` only per default.

If you want to run on all datasets (previous behavior, add `[circle full]` to your commit message, e.g.: `Add cool new feature [circle full]`.

config.py
```diff
@@ -934,6 +934,17 @@
...
 
+min_ecg_epochs: int = 5
+"""
+Minimal number of ECG epochs needed to compute SSP or ICA rejection with autoreject.
+"""
+
+min_eog_epochs: int = 5
+"""
+Minimal number of EOG epochs needed to compute SSP or ICA rejection with autoreject.
+"""
```
agramfort, line 945: I would not restrict this to autoreject. What do you think @dengemann ?

dengemann:
Yeah that makes sense… also when not using autoreject computing SSPs requires data
________________________________

apmellot:
There is something I don't understand in your comments @agramfort and @dengemann : in which other situation than with autoreject a minimal number of epochs is needed ? When I look at the code I only see two ways of doing the ssp reject : with autoreject or with the reject parameter directly in the config file. Epochs are only computed in the first scenario that is why I only put a minimal number of epochs for autoreject.

agramfort:
Do an experiment with mne Sample dataset. How many events do you need to estimate an ssp projection ? If you have too little data it will be very noisy and it’s maybe better to simply not apply any projection

agramfort:
@apmellot can you see why CI fails?

also do you confirm it runs smoothly for you as is?

apmellot:
Yes @agramfort , I changed a message in run_ssp and now the line is too long.
I tested it on subjects with enough and not enough eog epochs, it works on both.

agramfort (approved):
@dengemann please merge this if ok for you.

dengemann:
Thank you @apmellot @hoechenberger  & @agramfort !

PR #429, opened 2021-07-13
WIP : Storing processing logs and failures in excel file

agramfort:
@hoechenberger I would love your input on this one

PR #435, opened 2021-07-15
MAINT: Only build with circle full

### LINKED ISSUE #434: CI: CircleCI builds causing backups
In MNE the `main` build is slow (~2h) but on PRs it's fast (~15 min) because we triage what needs to be built based on changes. It would be good if there were some way to do the same thing for MNE-BIDS-Pipeline, as the PRs currently open are causing backups for all `mne-tools` projects on CircleCI.

One easy and useful option would be to only build one project on a given PR unless the `[circle full]` tag is passed, but always do the full build on `main` builds. I'm happy to open a PR to do this if people agree it would be helpful.

larsoner:
Closes #434

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -15,54 +15,7 @@ _bashenv: &bashenv
   name: Set BASH_ENV
   command: |
     set -e
-
-    # skip job if not in a [circle full] or on main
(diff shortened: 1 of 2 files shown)


larsoner:
I think this is close to working, but we need the regression from #429 to be fixed before we know

hoechenberger:
Thanks, @larsoner!

PR #436, opened 2021-07-16
Fix incorrect step display in log output

hoechenberger:
The issue we're seeing (processing steps are not displayed correctly / are outdated in the log) is due to the fact that we're using an environment variable, runpy, and parallelization, combined with Python's caching of imports.

After fiddling with this for quite a while and reading LOTS of Stack Overflow posts, I conclude that we cannot cleanly achieve our goal by setting an environment variable in run.py. We are not seeing similar issues with all the other variables we're using because they don't change over the course of the pipeline run. However, the MNE_BIDS_STUDY_SCRIPT_PATH is supposed to be updated with each new script that we call. This is why the issue surfaces there.

<strike>My (interim?) solution is to write the path of the currently-executed script to disk, read it back from there in config.py, and set the environment variable accordingly – in config.py, too, and not in run.py as we currently do on `main`.

This comes with the limitation that we can only have one pipeline instance run at a time, but I guess this is okay for now. There are certainly ways to resolve this and support multiple pipelines running in parallel, but I think we need to discuss this.

WDYT?</strike>

### Before merging …

### DIFF
--- a/config.py
+++ b/config.py
@@ -1526,18 +1526,6 @@ of an error.
 PIPELINE_NAME = 'mne-bids-pipeline'
 VERSION = '0.1.dev0'
 CODE_URL = 'https://github.com/mne-tools/mne-bids-pipeline'
-MNE_CONFIG_DIR = pathlib.Path(mne.get_config_path()).parent
-PIPELINE_SETTINGS_FILE = MNE_CONFIG_DIR / 'mne-bids-pipeline.json'
-
-
-def _get_current_script() -> str:
-    with PIPELINE_SETTINGS_FILE.open('r', encoding='utf-8') as f:
-        settings = json_tricks.load(f)
-
-    return settings['current_script']
-
-
-os.environ['MNE_BIDS_STUDY_SCRIPT_PATH'] = _get_current_script()
 
 ###############################################################################
 # Logger
--- a/config.py
+++ b/config.py
@@ -1526,6 +1526,18 @@ of an error.
 PIPELINE_NAME = 'mne-bids-pipeline'
 VERSION = '0.1.dev0'
 CODE_URL = 'https://github.com/mne-tools/mne-bids-pipeline'
+MNE_CONFIG_DIR = pathlib.Path(mne.get_config_path()).parent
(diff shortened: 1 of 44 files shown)


run.py
```diff
@@ -234,12 +238,25 @@ def process(config: PathLike,
...
         step_name = f'{script_path.parent.name}/{script_path.name}'
         logger.info(f'🚀 Now running script: {step_name} 👇')
         _run_script(script_path, config, root_dir, subject, session, task, run,
                     n_jobs)
         logger.info(f'🎉 Done running script: {step_name} 👏')
 
+    # Clear the settings
+    with PIPELINE_SETTINGS_FILE.open('w', encoding='utf-8') as f:
+        settings = {
+            'current_script': ''
```
hoechenberger, line 256:
```suggestion
            'current_script': None
```

agramfort:
I think it’s fragile too… I consider we stop trying to do black magic and we pass __file__ to an option script_path variable to the logging function

hoechenberger:
> I consider we stop trying to do black magic and we pass **file** to an option script_path variable to the logging function

This won't be easy for all functions residing inside `config.py`. Imagine we call `config.foo()` from `01-myscript.py`. All logging happening inside `config.foo()` would need a way to know which script called it.

We could obviously add a new param to all `config.py`-based functions to pass `__file__` when we call them, but I'm not sure if I like that? I mean, we could decide all functions need to accept a `script_path` param. WDYT?

hoechenberger:
> I think it’s fragile too…

We could use a lock file and I have ideas for supporting parallel runs. Maybe we should have a brief call?

hoechenberger:
@agramfort 
New approach based on @tomMoral's suggestion, WDYT?

config.py
```diff
@@ -1527,6 +1528,8 @@ def get_t1_from_meeg(bids_path):
 VERSION = '0.1.dev0'
 CODE_URL = 'https://github.com/mne-tools/mne-bids-pipeline'
 
+os.environ['MNE_BIDS_STUDY_SCRIPT_PATH'] = str(__file__)
```
hoechenberger, line 1531: Needs to be set for `config.py` because it's not passing through `failsafe_run()`, which does this export for all other scripts

agramfort (approved):
Nice

PR #439, opened 2021-07-16
WIP: Optimize ds000248 downloading & caching

hoechenberger:
### Before merging …


larsoner:
If many dataset-specific changes are expected, you could add support for `[circle ds000248]` or so, where the commit message is searched (after looking for `[circle full]` and/or `[circle skip]`) for an instance of `[circle X Y Z]` and where `X` `Y` and `Z` are treated as a substrings to match to various dataset names.

agramfort:
One idea is to download each dataset only once but put the derivatives in different folders for each version

hoechenberger:
> One idea is to download each dataset only once but put the derivatives in different folders for each version 

Yep, that's what I'm trying to do here... hang on ;)

agramfort (dismissed):
I like the approach 👍

scripts/report/01-make_reports.py
```diff
@@ -695,6 +695,25 @@ def get_config(
...
+    # simply assign a dummy value in such cases.
+    # `get_fs_subject()` calls `get_fs_subjects_dir()`, so take care of this
+    # too.
+    try:
+        fs_subjects_dir = config.get_fs_subjects_dir()
+    except ValueError: 
+        fs_subjects_dir = None
+        fs_subject = None
+    else:
+        fs_subject = config.get_fs_subject(subject=subject),
```
hoechenberger, line 715: I'm pretty certain @agramfort won't like this. Alternative approaches welcome.
(I was thinking about adding an `on_error` switch to `get_fs_subjects_dir()`, but then thought it'd be a bit over-engineered)

hoechenberger:
> I like the approach 👍

Now that actually comes as a surprise 😅 Nice 😁

scripts/report/01-make_reports.py
```diff
@@ -695,6 +695,25 @@ def get_config(
...
+    # simply assign a dummy value in such cases.
+    # `get_fs_subject()` calls `get_fs_subjects_dir()`, so take care of this
+    # too.
+    try:
+        fs_subjects_dir = config.get_fs_subjects_dir()
+    except ValueError:
+        fs_subjects_dir = None
+        fs_subject = None
+    else:
+        fs_subject = config.get_fs_subject(subject=subject),
```
hoechenberger, line 715:
```suggestion
        fs_subject = config.get_fs_subject(subject=subject)
```

tests/configs/config_ERP_CORE.py
```diff
@@ -28,6 +28,7 @@
 
 study_name = 'ERP-CORE'
 bids_root = '~/mne_data/ERP_CORE'
+deriv_root = '~/mne_data/mne-bids-pipeline/tests/ERP_CORE'
```
agramfort, line 31: I would have even considered:
```suggestion
deriv_root = '~/mne_data/derivatives/mne-bids-pipeline/tests/ERP_CORE'
```
so we just move one step up the derivatives folder. But I won't fight for it.

agramfort (commented):
let me know when it's done for you @hoechenberger

tests/configs/config_ERP_CORE.py
hoechenberger: Yes I like this. Then we can drop the `test` subdir too:

```
'~/mne_data/derivatives/mne-bids-pipeline/ERP_CORE'
```

agramfort:
+1

hoechenberger:
@agramfort this is good to merge from my end

hoechenberger:
Merging on behalf of @agramfort

PR #441, opened 2021-07-19
Add one line of output if we skip a step

hoechenberger:
We now print "Skipping …" if a step is being skipped, so users can be certain there was no fluke.

### Before merging …

### DIFF
--- a/run.py
+++ b/run.py
@@ -238,7 +238,7 @@ def process(config: PathLike,
         logger.info(f'🚀 Now running script: {step_name} 👇')
         _run_script(script_path, config, root_dir, subject, session, task, run,
                     n_jobs)
-        logger.info(f'🎉 Done running script: {step_name} 👏')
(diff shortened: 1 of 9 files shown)


run.py
```diff
@@ -238,7 +238,7 @@ def process(config: PathLike,
         logger.info(f'🚀 Now running script: {step_name} 👇')
         _run_script(script_path, config, root_dir, subject, session, task, run,
                     n_jobs)
-        logger.info(f'🎉 Done running script: {step_name} 👏')
+        logger.info(f'🎉 Done running script: {step_name} 👆')
```
hoechenberger, line 241: That way, log messages belonging to one step can be found as in
```
🚀 Now running script: xxx 👇
foo
bar
🎉 Done running script: foo 👆
```

agramfort (approved):

PR #452, opened 2021-10-04
WIP ENH support source localization of non-evoked data

apmellot:
### Before merging …

We already managed add support for non-evoked data in make_forward but it is not finished for make_inverse : it writes the inverse solution and apply it to evokeds. We could add a parameter to choose on which files to apply the inverse solution.
@hoechenberger @agramfort @dengemann

### DIFF
--- a/config.py
+++ b/config.py
@@ -1470,6 +1470,18 @@ covariance can ONLY be estimated from the pre-stimulus period.
     ```
 """
 
+source_info_update: Optional[Dict[str, str]] = None
+"""
+Match the file name from where to read the info in forward and inverse solution
+computation by updating 'bids_path'.
+
+???+ example "Example"
+    ```python
+    source_info_update = {'processing': 'clean',
+                          'suffix': 'epo'}
+    ```
+"""
+
 ###############################################################################
 # ADVANCED
 # --------
--- a/scripts/source/02-make_forward.py
+++ b/scripts/source/02-make_forward.py
@@ -118,7 +118,10 @@ def run_forward(*, cfg, subject, session=None):
                          root=cfg.deriv_root,
                          check=False)
 
-    fname_evoked = bids_path.copy().update(suffix='ave')
+    if cfg.source_info_update == None:
+        fname_info = bids_path.copy().update(suffix='ave')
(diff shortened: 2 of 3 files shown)


config.py
```diff
@@ -1470,6 +1470,18 @@ def get_t1_from_meeg(bids_path):
     ```
 """
 
+source_info_update: Optional[Dict[str, str]] = None
```
agramfort, line 1473: if you make it default to `dict(suffix='ave')` then if you avoid the if and you can document that 
is this case you take the info from the evoked by default

dengemann: that makes sense.

config.py
```diff
@@ -1482,6 +1482,18 @@ def get_t1_from_meeg(bids_path):
     ```
 """
 
+inverse_targets: Optional[List[str]] = []
```
hoechenberger, line 1485: Shouldn't this default to [`evoked`]?

config.py
```diff
@@ -1470,6 +1470,30 @@ def get_t1_from_meeg(bids_path):
...
+    source_info_update = {'processing': 'clean',
+                          'suffix': 'epo'}
+    ```
+"""
+
+inverse_targets: Optional[List[str]] = []
+"""
+List of the different types of data on which to apply the inverse operator.
+
+If [] only the inverse solution coputation is done.
```
hoechenberger, line 1489: I would make it `None` in this case … `Optional[...]` in fact implies that `None` is an accepted value

config.py
apmellot: Indeed :)

config.py
```diff
@@ -1470,6 +1470,30 @@ def get_t1_from_meeg(bids_path):
...
+
+inverse_targets: Optional[List[str]] = ['evoked']
+"""
+List of the different types of data on which to apply the inverse operator.
+
+If None only the inverse solution computation is done.
+
+???+ example "Example"
+    ```python
+    inverse_targets = ['raw', 'evoked', 'cov', 'epochs']
```
agramfort, line 1493: so far anything different from evoked with be silently ignored.

PR #468, opened 2021-11-09
Adjustments for MNE-Python main dev branch

hoechenberger:
### Before merging …

### DIFF
--- a/config.py
+++ b/config.py
@@ -2917,12 +2917,17 @@ def import_er_data(
     return raw_er
 
 
-def get_reference_run_info(
+class ReferenceRunParams(TypedDict):
+    montage: mne.channels.DigMontage
+    dev_head_t: mne.Transform
+
+
+def get_reference_run_params(
     *,
     subject: str,
     session: Optional[str] = None,
     run: str
-) -> mne.Info:
+) -> ReferenceRunParams:
 
     msg = f'Loading info for run: {run}.'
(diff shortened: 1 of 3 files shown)


hoechenberger:
CI failing because of https://github.com/OpenNeuroOrg/openneuro/issues/2472

hoechenberger:
The issue upstream at OpenNeuro has been resolved, restarting CI

config.py
```diff
@@ -2906,9 +2906,12 @@ def import_er_data(
     _drop_channels_func(cfg, raw=raw_er, subject='emptyroom', session=session)
 
     # Set same set of bads as in the experimental run, but only for MEG
-    # channels (because we won't have any others in empty-room recordings)
+    # channels (we might not have non-MEG channels in empty-room recordings).
     raw_er.info['bads'] = [ch for ch in bads if ch.startswith('MEG')]
 
+    # Only keep MEG channels.
+    raw_er.pick_types(meg=True)
```
agramfort, line 2913: for CTF I assume you also need ref_meg=True

to be tested though

larsoner: The default for `pick_types` should be `'auto'` which means

> If ‘auto’, reference channels are included if compensations are present and meg is not False. 

So for CTF I think this will end up acting like True anyway

agramfort:
hopefully :)

>

PR #471, opened 2021-11-17
Fix bug in make_epochs()

hoechenberger:
We didn't make use of provided `event_id` when generating events from annotations. This could lead to inconsistent event code <> event name mappings across
runs or participants.

### Before merging …

### DIFF
--- a/run.py
+++ b/run.py
@@ -4,11 +4,12 @@ import os
 import sys
 import pathlib
 import logging
-from typing import Union, Optional, Tuple
+from typing import Union, Optional, Tuple, List, Dict
 if sys.version_info >= (3, 8):
     from typing import Literal
 else:
     from typing_extensions import Literal
+from types import ModuleType
 
 import fire
 import coloredlogs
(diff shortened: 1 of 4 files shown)


hoechenberger:
CI failing due to upstream bug:
https://github.com/mne-tools/mne-bids/pull/916

run.py
```diff
@@ -32,8 +33,17 @@
...
+    *,
+    config: PathLike,
+    root_dir: Optional[PathLike] = None,
+    subject: Optional[str] = None,
+    session: Optional[str] = None,
+    task: Optional[str] = None,
+    run: Optional[str] = None,
+    interactive: Optional[str] = None,
+    n_jobs: Optional[str] = None
+) -> Dict[str, Tuple[ModuleType]]:
```
ktavabi, line 46: What kind of declaration is this? @hoechenberger

hoechenberger: These are called "type hints" and I think were introduced with Python 3.6 initially.

You need Python 3.7 or newer for these here to work.

PR #472, opened 2021-11-17
Add support for parallelization via Dask backend in joblib

agramfort:
Screenshot 2021-11-17 at 22 44 41

seems to work !


run.py
```diff
@@ -248,11 +248,12 @@ def process(
         # n_workers = multiprocessing.cpu_count()  # FIXME should use N_JOBS
         n_workers = 4
         logger.info(f'👾 Initializing Dask client with {n_workers} workers …')
+        dask_temp_dir = "./.dask-worker-space"
```
hoechenberger: Careful with this, as this will spill memory to disk in the current working directory, which could leads to problems

config.py
```diff
@@ -2207,6 +2212,7 @@ def get_n_jobs() -> int:
 def setup_dask_client():
     global dask_client
     global dask_temp_dir
+    global dask_worker_memory_limit
     global open_dask_dashboard
```
hoechenberger, line 2216: These 3 shouldn't be necessary, no?

run.py
agramfort: this is gone now

config.py
agramfort: indeed

agramfort:
I think this one is ready for more testing / reviews

hoechenberger:
Thanks @agramfort, I'll put it to the test tonight / this weekend!

tests/configs/config_ds000247.py
```diff
@@ -23,3 +23,5 @@
 
 epochs_tmin = 0
 baseline = None
+
+on_error = "debug"
```
hoechenberger, line 27:
```suggestion
```

hoechenberger (commented):
Looks good so far, I'll run some tests and update the docs

hoechenberger:
CI error for ds000248 is due to an OpenNeuro problem, reported at https://github.com/OpenNeuroOrg/openneuro/issues/2521

hoechenberger (approved):

agramfort:
all green https://22114-167957354-gh.circle-artifacts.com/0/site/index.html 🚀 

merging !

hoechenberger:
😍😍😍

PR #477, opened 2021-11-26
Drastically reduce memory usage during epoching

hoechenberger:
Peak memory consumption during epoching reduced by more than 50%!
And we're also faster now!

# Benchmarks

### `main`
Screen Shot 2021-11-26 at 10 24 14

### this branch

Screen Shot 2021-11-26 at 10 24 32

### `main`

Screen Shot 2021-11-26 at 10 24 53

### this branch

Screen Shot 2021-11-26 at 10 25 20

# How this was achieved

- We preload epochs so we can remove the respective raw from memory
- Instead of creating a list of all epochs for all runs before concatenation and then concatenating them all at once, we now concatenate epochs immediately after they were created. This further reduced peak memory usage significantly (although I don't fully understand why!)
- During ICA epochs generation, we now drop all epochs that are not in `config.conditions`. In `main`, we create epochs from ALL events, often creating uselessly huge numbers of epochs, increasing memory footprint and massively slowing down ICA decomposition.
- We now only keep a max of 300 EOG and ECG epochs (hardcoded limit)

# Further ideas

We may be able to further reduce peak memory consumption at the cost of I/O and processing speed.

- For example, currently we're creating EOG and ECG epochs in the same loop as the "task" epochs. If we split this into separate loops, peak memory consumption could be lower.
- Currently, EOG and ECG epochs are reduced to 300 elements only AFTER the bloated epochs objects have been created. Instead, one could first only extract EOG and ECG **events**, subset those to get 300, and only then create the epochs.

### Before merging …

### DIFF
--- a/scripts/preprocessing/_04a_run_ica.py
+++ b/scripts/preprocessing/_04a_run_ica.py
@@ -116,8 +116,8 @@ def make_ecg_epochs(
         ecg_dur_per_run = total_ecg_dur / n_runs
         t_mid = (raw.times[-1] - raw.times[0]) / 2
         raw = raw.crop(
-            tmin=t_mid - 1/2 * ecg_dur_per_run,
-            tmax=t_mid + 1/2 * ecg_dur_per_run
+            tmin=max(t_mid - 1/2 * ecg_dur_per_run, 0),
+            tmax=min(t_mid + 1/2 * ecg_dur_per_run, raw.times[-1])
         ).load_data()
 
         ecg_epochs = create_ecg_epochs(raw,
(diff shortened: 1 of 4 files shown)


hoechenberger:
I just disabled EOG and ECG epochs creation during ICA and this is what the mprof looks like then:

Screen Shot 2021-11-26 at 10 25 20

So this is definitely a must-investigate.

hoechenberger:
Current status during the ICA script, after implementing the changes we discussed during our Zoom call @agramfort 

There's still room for improvement but I guess for now it's okay to move on!

Screen Shot 2021-11-26 at 16 26 59

scripts/preprocessing/_04a_run_ica.py
```diff
@@ -114,7 +114,7 @@ def make_ecg_epochs(
         # (across all runs)
         total_ecg_dur = 5 * 60
         ecg_dur_per_run = total_ecg_dur / n_runs
-        t_mid = (raw.times[-1] - raw.times[0]) / 2
+        t_mid = (raw.times[-1] + raw.times[0]) / 2
```
hoechenberger, line 117: Ah, good catch! A true mathematician 😁

hoechenberger:
Merci @agramfort!

PR #479, opened 2021-11-26
Add CSP

hoechenberger:
This is a WIP rework of #414

Closes #414

### Before merging …

- [ ] Include new config options in documentation


config.py
```diff
@@ -1421,8 +1421,106 @@ class ArbitraryContrast(TypedDict):
...
+"""
+
+###############################################################################
+# TIME-FREQUENCY CSP
+# ------------------
+
+decoding_csp: bool = False
+"""
+Whether to run decoding via Common Spatial Patterns (CSP) analysis on the
+time-frequency representation of the data.
```
agramfort, line 1436: you don't feed the TFR data as input.
```suggestion
Whether to run decoding via Common Spatial Patterns (CSP) analysis on the
data. CSP takes as input data covariances that are estimated on different
time and frequency ranges. This allows to obtain decoding scores defined over
time and frequency.
```

config.py
```diff
@@ -1421,8 +1421,106 @@ class ArbitraryContrast(TypedDict):
...
+"""
+
+decoding_csp_times: Optional[ArrayLike] = np.linspace(
+    max(0, epochs_tmin),
+    epochs_tmax,
+    num=6
+)
+"""
+The edges of the time bins to use for CSP decoding.
+Must contain at least two elements. By default, 4 equally-spaced bins are
```
agramfort, line 1446:
```suggestion
Must contain at least two elements. By default, 5 equally-spaced bins are
```

config.py
```diff
@@ -1421,8 +1421,106 @@ class ArbitraryContrast(TypedDict):
...
+    Create 2 time bins of different durations (0–0.4, 0.4–0.6 sec):
+    ```python
+    decoding_csp_times = [0, 0.4, 0.6]
+    ```
+"""
+
+decoding_csp_freqs: Dict[str, ArrayLike] = {
+    'custom': [
+        time_frequency_freq_min,
+        (time_frequency_freq_max - time_frequency_freq_min) / 2 + time_frequency_freq_min,  # noqa: E501
```
agramfort, line 1468:
```suggestion
        (time_frequency_freq_max + time_frequency_freq_min) / 2,  # noqa: E501
```
but just use `np.linspace(time_frequency_freq_min, time_frequency_freq_max, 3)`

config.py
```diff
@@ -1421,8 +1421,106 @@ class ArbitraryContrast(TypedDict):
...
+        time_frequency_freq_max
+    ]
+}
+"""
+The edges of the frequency bins to use for CSP decoding.
+
+This parameter must be a dictionary with they keys specifying the "name" to use
```
agramfort, line 1475:
```suggestion
This parameter must be a dictionary with keys specifying the "name" to use
```

config.py
```diff
@@ -1421,8 +1421,106 @@ class ArbitraryContrast(TypedDict):
...
+scalar values, specifying the edges of the respective frequency bin(s).
+
+Defaults to two frequency bins, one from
+[`time_frequency_freq_min`][config.time_frequency_freq_min] to the mindpoint
+between this value and
+[`time_frequency_freq_max`][config.time_frequency_freq_max]; and the other from
+that midpoint to `time_frequency_freq_max`.
+
+???+ example "Example"
+    Create two frquency bins, one for 4–8 Hz, and another for 8–14 Hz:
```
agramfort, line 1486:
```suggestion
    Create two frequency bins, one for 4–8 Hz, and another for 8–14 Hz:
```

config.py
```diff
@@ -1421,8 +1421,106 @@ class ArbitraryContrast(TypedDict):
...
+        'upper_range': [10, 14]
+    ]
+    ```
+
+    Create partially overlapping frequency bins, one for 4–10 Hz, and another
+    for 8–14 Hz:
+    ```python
+    decoding_csp_freqs = {
+        'lower_range': [4, 10],
+        'upper_range': [8, 14]
```
agramfort, line 1518: how does this affect 2d plots?

config.py
```diff
@@ -2201,6 +2299,39 @@ def check_baseline(
             )
 
 
+if decoding_n_splits < 2:
+    raise ValueError('decoding_n_splits should be at least 2.')
```
agramfort, line 2303: can you check that setting decoding_n_splits=1 in a custom config raises the exception?

scripts/report/_01_make_reports.py
```diff
@@ -125,40 +125,40 @@ def plot_auto_scores(cfg, subject, session):
...
-#     ax.axhline(0.5, ls='--', lw=0.5, color='black', label='chance')
-
-#     ax.set_xlabel(f'{contrast[0]} ./. {contrast[1]}')
-#     if metric == 'roc_auc':
-#         metric = 'ROC AUC'
-#     ax.set_ylabel(f'Score ({metric})')
-#     ax.legend(loc='center right')
-#     fig.tight_layout()
-
-#     return fig
```
agramfort, line 161: to fix

scripts/report/_01_make_reports.py
```diff
@@ -1514,6 +1519,159 @@ def add_decoding_grand_average(
...
+            )[0]
+            significant_clusters = clusters[significant_cluster_idx]
+
+            # vmax = np.abs(t_vals).max()
+            # vmin = -vmax
+            vmax = max(
+                np.abs(mean_crossval_scores.min() - 0.5),
+                np.abs(mean_crossval_scores.max() - 0.5)
+            ) + 0.5
+            vmin = 0.5 - (vmax - 0.5)
```
agramfort, line 1577: this only makes sense for roc_auc, balanced accuracy or accuracy if data have been balanced before training.

scripts/sensor/_04b_time_frequency_csp.py
```diff
@@ -0,0 +1,436 @@
...
+
+    # Classifier
+    csp = CSP(
+        n_components=4,  # XXX revisit
+        reg=0.1,         # XXX revisit
+        rank='info',
+    )
+    clf = make_pipeline(
+        csp,
+        LinearDiscriminantAnalysis(),
```
agramfort, line 185: I would say we default to LogReg like in other decoding. Unless there is strong evidence that it's worse

scripts/sensor/_04b_time_frequency_csp.py
```diff
@@ -0,0 +1,436 @@
...
+                    'cond_2': condition2,
+                    'f_min': f_min,
+                    'f_max': f_max,
+                    'freq_range_name': freq_range_name,
+                    'mean_crossval_score': np.nan,
+                    'metric': cfg.decoding_metric,
+                },
+                index=[0]
+            )
+            freq_decoding_table = pd.concat(
```
agramfort, line 220: I would avoid calling concat to allow memory as each iteration and use first a list

scripts/sensor/_04b_time_frequency_csp.py
```diff
@@ -0,0 +1,436 @@
...
+            f'({freq_range_name}), '
+            f'Times (s): {round(tmin, 3)}–{round(tmax, 3)}'
+        )
+        logger.info(
+            **gen_log_kwargs(msg, subject=subject, session=session)
+        )
+
+        epochs_filt, y = prepare_epochs_and_y(
+            epochs=epochs, contrast=contrast, fmin=fmin, fmax=fmax, cfg=cfg
+        )
```
agramfort, line 332: so here you filter the same way for each time bin so you seem to redo many time the same costly temporal filtering.

config.py
hoechenberger: Confirmed that it works

scripts/report/_01_make_reports.py
hoechenberger: I've added an `assert` here to ensure we're only working with `roc_auc` for now.

There are other places in the pipeline where we silently assume a chance level of 50%, hence we should probably address this entire issue in a separate PR

scripts/sensor/_04b_time_frequency_csp.py
hoechenberger: Done, let's see how that flies…

scripts/sensor/_04b_time_frequency_csp.py
hoechenberger: I've adjusted the logic to only call it once on a list of dicts

config.py
hoechenberger: For each freq range, we basically perform an independent analysis, and also generate independent figures. See here where I had an "alpha" and a "test" freq range.

Screen Shot 2022-07-20 at 14 26 22

hoechenberger: For each freq range, we basically perform an independent analysis, and also generate independent figures. See here where I had an "alpha" and a "test" freq range.

Screen Shot 2022-07-20 at 14 26 22

PR #481, opened 2021-12-01
Add psutil dependency

hoechenberger:
joblib has some built-in memory leak detection that requires `psutil` to function. I think it therefore makes sense to add `psutil` to our list of dependencies.

https://joblib.readthedocs.io/en/latest/developing.html#release-0-12-2

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,5 @@
 typing_extensions
+psutil
 numpy
 scipy
 scikit-learn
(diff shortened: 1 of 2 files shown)


agramfort:
it's not installed automatically by joblib?

>

hoechenberger:
> it's not installed automatically by joblib?

Nope, it's purely optional.

agramfort:
ok then !

PR #482, opened 2021-12-01
Add epochs to the report. Add plot_psd_for_runs setting.

hoechenberger:
Previously we'd only plot the PSD of raw data for the first run to speed up report generation. Working with @SophieHerbst we discovered that this is not a good choice for us, so this PR:

- changes the default behavior such that we get a PSD plot for each run
- add a new config option, `plot_psd_for_runs`, allowing users to control this behavior

Also, we now add epochs to the report. It appears I simply forgot about them when transitioning to the new MNE Report API.

### Before merging …

### DIFF
--- a/docs/source/changes.md
+++ b/docs/source/changes.md
@@ -114,7 +114,7 @@ authors:
   ({{ gh(477) }} by {{ authors.hoechenberger }} and {{ authors.agramfort }})
 - The new [`plot_psd_for_runs`][config.plot_psd_for_runs] setting can be used
   to control for which runs to add PSD plots of the raw data to the reports.
-  ({{ gh(459) }} by {{ authors.hoechenberger }})
+  ({{ gh(482) }} by {{ authors.hoechenberger }})
 
 ### Behavior changes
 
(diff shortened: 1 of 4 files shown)


docs/source/changes.md
```diff
@@ -112,6 +112,9 @@ authors:
   ({{ gh(460) }} by {{ authors.agramfort }} and  {{ authors.apmellot }})
 - Drastically reduce memory usage during the epoching and ICA steps.
   ({{ gh(477) }} by {{ authors.hoechenberger }} and {{ authors.agramfort }})
+- The new [`plot_psd_for_runs`][config.plot_psd_for_runs] setting can be used
+  to control for which runs to add PSD plots of the raw data to the reports.
+  ({{ gh(459) }} by {{ authors.hoechenberger }})
```
hoechenberger, line 117:
```suggestion
  ({{ gh(482) }} by {{ authors.hoechenberger }})
```

hoechenberger:
@agramfort 
See https://19544-167957354-gh.circle-artifacts.com/0/reports/ds000247/sub-0002_ses-01_task-rest_report.html

We now have epochs before and after cleaning

agramfort (approved):

Issue #332, opened 2021-05-18
Create report just for preprocessing


guiomar:
Hi! I think it could be nice to have the option of creating a report just with the "preprocessing" steps. Currently it seems it looks for the *_ave.fif file, but maybe this is generated in the "sensor" part?

agramfort:
the report needs a lot of love indeed. It is indeed a good idea to split it
by sections (preprocessing, sensor, source)

would you have the bandwidth to help with this?

>

guiomar:
Sure! if you show me some basics, so I can follow the logic and don't mess anything, I can give it a try :)
(I'm just learning all this, so I'll go a bit slowly at the beguining)

agramfort:
can you join the mne discord server https://discord.gg/cQHS5qtV so we can
easily chat?

>

guiomar:
Sure! Thanks!

larsoner:
I think this is more or less taken care of by the section headings in the left bar of the report nowadays so I'll close, but feel free to reopen if it's still not usable

PR #483, opened 2021-12-02
Split report generation function into smaller entities

hoechenberger:
We now have `run_report_preprocessing`, `run_report_sensor`, and `run_report_source`. Should be easier to digest, maintain, and profile!

This is also a step toward resolving #332 (allowing us to produce reports directly after preprocessing)


agramfort:
report are still in one single file?

hoechenberger:
> report are still in one single file?

Yes! User-facing behavior is unchanged!

agramfort (approved):

PR #492, opened 2021-12-27
Forward solution routine

ktavabi:
Looks like the forward solution failsafe routine needs a better check for existing coregistration `trans.fif` file(s)

### Before merging …


hoechenberger:
Great, thank you!

ktavabi:
FYI there were several additional points along the way where `overwrite=True` was useful.

hoechenberger:
> FYI there were several additional points along the way where `overwrite=True` was useful;

The idea would be that we'd always overwrite data in `derivatives` on a re-run of the pipeline. If you know of any other places where `overwrite=True` would be needed, please feel free to open another PR! There were some upstream changes in MNE-Python where an `overwrite` parameter was added to a bunch of save functions, and it usually defaults to `False`.

agramfort:
thx @ktavabi 🙏

PR #495, opened 2021-12-31
Add epochs_metadata_excludes

hoechenberger:
This new setting allows for the exclusion of epochs based on metadata.

In theory this is already possible via `conditions`, but especially with lots of nested event names, this becomes very cumbersome very quickly, because the auto-generated metadata dataframe currently does not create columns for such event groupings. That is, if you have e.g. the following events:

```python
stim/1
stim/2
stim/3
stim/4
stim/5
bad_response
```
and you want to analyze all `stim` trials but exclude all epochs that have a `bad_response` event in the metadata. You cannot simply specify:
```python
conditions = {'stim': 'stim.notna() and bad_response.isna()'}
```
because there won't be a column `'stim'` in the auto-generated metadata. Instead, you'd have to add each event name explicitly, and use backticks to make the forward slashes work:
```python
conditions = {
    'stim':
        '(`stim/1`.notna() or '
        ' `stim/2`.notna() or '
        ' `stim/3`.notna() or '
        ' `stim/4`.notna() or '
        ' `stim/5`.notna()) and'
        'bad_response.isna()'
}
```
Now imagine doing this with even more conditions … This is not very user friendly.
 
Also there might be experiments where **all** participants get to be presented with a `stim`, but you might have groups that only get to see `stim/1` and `stim/2`, while others only get to see `stim/3` and `stim/4`. You'd have to produce different `conditions` for each group, as the columns corresponding to the not-presented stimuli would simply be missing, and the query would fail, complaining about missing columns.

With this PR, the above situation can be addressed by simply specifying:

```python
epochs_metadata_excludes = ['bad_response.notna()']
conditions = ['stim']
```

### Before merging …

### DIFF
--- a/config.py
+++ b/config.py
@@ -821,7 +821,7 @@ will be named with a ``last_`` instead of a ``first_`` prefix.
 epochs_metadata_excludes: Iterable[str] = []
 """
 [Metadata queries][https://mne.tools/stable/auto_tutorials/epochs/30_epochs_metadata.html]
-specifying which epochs to exclude from analysis. If a metadata query fails
+specifying which epochs to exclude from analysis. If a query fails
 because it refers to an unknown metadata column, a warning will be emitted.
 
 ???+ example "Example"
--- a/config.py
+++ b/config.py
@@ -818,6 +818,19 @@ occurrence of matching event types. The columns indicating the event types
 will be named with a ``last_`` instead of a ``first_`` prefix.
 """
 
+epochs_metadata_excludes: Iterable[str] = []
+"""
+[Metadata queries][https://mne.tools/dev/auto_tutorials/epochs/30_epochs_metadata.html]
+specifying which epochs to exclude from analysis. If a metadata query doesn't
+yield any epochs, a warning will be emitted.
+
+???+ example "Example"
+    Exclude all epochs with a `response_missing` event:
+    ```python
+    epochs_metadata_excludes = ['response_missing.notna()']
+    ```
+"""
+
 conditions: Optional[Union[Iterable[str], Dict[str, str]]] = None
 """
 The time-locked events based on which to create evoked responses.
--- a/config.py
+++ b/config.py
@@ -820,9 +820,9 @@ will be named with a ``last_`` instead of a ``first_`` prefix.
 
 epochs_metadata_excludes: Iterable[str] = []
 """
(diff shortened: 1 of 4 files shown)


config.py
```diff
@@ -818,6 +818,19 @@
 will be named with a ``last_`` instead of a ``first_`` prefix.
 """
 
+epochs_metadata_excludes: Iterable[str] = []
+"""
+[Metadata queries][https://mne.tools/dev/auto_tutorials/epochs/30_epochs_metadata.html]
```
hoechenberger, line 823:
```suggestion
[Metadata queries][https://mne.tools/stable/auto_tutorials/epochs/30_epochs_metadata.html]
```

config.py
```diff
@@ -818,6 +818,19 @@
...
+specifying which epochs to exclude from analysis. If a metadata query doesn't
+yield any epochs, a warning will be emitted.
```
hoechenberger, line 825:
```suggestion
specifying which epochs to exclude from analysis. If a metadata query fails
because it refers to an unknown metadata column, a warning will be emitted.
```

docs/source/settings/preprocessing/epochs.md
```diff
@@ -8,5 +8,6 @@
 ::: config.epochs_metadata_tmax
 ::: config.epochs_metadata_keep_first
 ::: config.epochs_metadata_keep_last
+::: config.epochs_metadata_exclude
```
hoechenberger, line 11:
```suggestion
::: config.epochs_metadata_excludes
```

config.py
```diff
@@ -818,6 +818,19 @@
 will be named with a ``last_`` instead of a ``first_`` prefix.
 """
 
+epochs_metadata_excludes: Iterable[str] = []
+"""
+[Metadata queries][https://mne.tools/stable/auto_tutorials/epochs/30_epochs_metadata.html]
+specifying which epochs to exclude from analysis. If a metadata query fails
```
hoechenberger, line 824:
```suggestion
specifying which epochs to exclude from analysis. If a query fails
```

agramfort:
couldn't you rather do:

`event_name.str.contains('stim/') and bad_response.isna()`

or something like this?

hoechenberger:
> couldn't you rather do:
> 
> `event_name.str.contains('stim/') and bad_response.isna()`
> 
> or something like this?

This might work, will give it a shot

hoechenberger:
@agramfort Your proposal works, closing this PR! Thanks!

PR #503, opened 2022-01-05
Subset events to desired conditions before Epochs creation

### LINKED ISSUE #499: Subset events to desired conditions before Epochs creation
Looking at our fancy new metadata tables in the report, I found that the indices of the epochs looked rather sparse:

Screen Shot 2022-01-03 at 21 48 12

I believe the reason is that we're first creating epochs based on all events found in the data (albeit, without preloading them), and only later subset them to include the events (conditions) of interest.

As I don't think there's a way to reset the "epoch index", we should most probably cut down the events to the desired set before we feed them to the `Epochs()` constructor.

The new metadata query selection should still take place only after epochs creation (anything else would cause headaches); then, when browsing the epochs and looking at their indices, it would become immediately obvious which epochs have been dropped through metadata selection (or, later on, via PTP rejection)

This is not super urgent but I think has quite some usability implications, so we should prioritize this.

hoechenberger:
Fixes #499

### Before merging …


config.py
```diff
@@ -2534,6 +2536,35 @@ def make_epochs(
                         event_repeated=event_repeated,
                         reject=None)
 
+    # Only keep conditions that will be analyzed.
+    # Note that we're not selection epochs based on metadata here.
+    if task != 'rest':
```
hoechenberger, line 2541:
```suggestion
    if task.lower() != 'rest':
```

config.py
```diff
@@ -2534,6 +2536,35 @@ def make_epochs(
...
+        if isinstance(conditions, dict):
+            conditions = list(conditions.keys())
+        epochs = epochs[conditions]
+
+        # Reset metadata index and do an open-heart surgery on the drop log to
+        # make everything appear as we never subset the epochs. Otherwise, the
```
hoechenberger, line 2547:
```suggestion
        # make everything appear as if we never subset the epochs. Otherwise, the
```

config.py
```diff
@@ -2534,6 +2536,35 @@ def make_epochs(
...
+        # "sparse" metadata index could confuse the user!
+        #
+        # ⛔️ Don't try this at home, kids! ⛔️
+        drop_log = tuple([
+            epochs.drop_log[i]
+            for i in epochs.metadata.index
+        ])
+        epochs.drop_log = drop_log
+        # Setting metadata must happen after updating `epochs.selection`,
+        # as the index will automatically be changed to `epochs.selection`.
```
hoechenberger, line 2557:
```suggestion
        # as the DataFrame index will automatically be changed to `epochs.selection`.
```

config.py
```diff
@@ -2534,6 +2536,35 @@ def make_epochs(
...
+        # So then all we need to do is set the metadata to itself, and its
```
hoechenberger, line 2558:
```suggestion
        # After that, all we need to do is set the metadata to itself, and its
```

hoechenberger:
Ok, all working as intended!

@agramfort See this produced report for a configuration where we run the metadata query `'index > 0'`:

https://21253-167957354-gh.circle-artifacts.com/0/site/examples/ds000248/sub-01_task-audiovisual_report.html#global17

You'll find that:

- in the metadata table, the row with index 0 is missing
- the rest of the index is without "gaps" 
- in the drop log plot, we can see that a trial was `IGNORED`, so it's immediately obvious that via the metadata query, we omitted one trial

This is good to be merged from my end.

Future improvements I've planned:
- move the drop log plot up, before the ERF plots
- provide more detailed (written) info on what was dropped, and why

config.py
```diff
@@ -2504,25 +2512,44 @@ def make_epochs(
...
+            conditions = list(conditions.keys())
+        else:
+            conditions = list(conditions)  # Ensure we have a list
 
-    if metadata_tmax is None:
-        metadata_tmax = tmax
+        # TODO This function should be made public upstream
+        row_event_names = mne.utils.mixin._hid_match(
+            event_id=event_id, keys=conditions
+        )
```
hoechenberger, line 2533: Any thoughts on that?

agramfort: we should avoid this. Can you do it eg using an EpochsArray just to get the indices?

hoechenberger: Maybe, but I'd like to avoid having to fiddle around too much … I think the cleanest approach would be to make `hid_match` public, or to duplicate it and include it in the Pipeline. WDYT?

agramfort: if would use here something like this for now:

```
import numpy as np
import mne

event_id = {'auditory/left': 1, 'auditory/right': 2, 'visual/left': 3,
            'visual/right': 4}

info = mne.create_info(['A'], 1)
data = np.random.randn(len(event_id), 1, 1)
events = np.zeros((len(event_id), 3), dtype=int)
events[:, 0] = events[:, 2] = np.arange(len(event_id)) + 1
epochs = mne.EpochsArray(data, info, event_id=event_id, events=events)
event_id_hid = epochs['left']
print(event_id_hid.event_id)
```

hoechenberger:
Blocked by https://github.com/mne-tools/mne-python/pull/10233

agramfort (approved):

PR #508, opened 2022-01-17
[tests] update eeg_matchingpennies location

### LINKED ISSUE #507: eeg_matchingpennies, remove git-annex version?
I am in the process of reorganizing the eeg_matchingpennies dataset (once again ...) in https://github.com/bids-standard/bids-examples/pull/265 - there is some information about why in https://github.com/mne-tools/mne-bids/issues/936#issuecomment-1006833527

When I am done there, I'd like to push these changes to the "full data storage" on OSF: https://osf.io/cj2dr/

Usually that means that I have to update the git-annex mirror at https://github.com/sappelhoff/eeg_matchingpennies which is always a bit hacky due to the way I implemented this (back then, datalad didn't have OSF integration, so this is based on a lot of manual steps and running a bunch of scripts in the right order).

Long story short: Would it be a problem if I just removed the git-annex mirror? --> the `eeg_matchingpennies` dataset could still be part of the pipeline via OSF, like "ERPCore":

https://github.com/mne-tools/mne-bids-pipeline/blob/ca17542efd233566fa80d7ecc1ec392add36073e/tests/datasets.py#L20-L25

cc @hoechenberger

sappelhoff:
closes #507 

- I updated the data in https://github.com/bids-standard/bids-examples/pull/265
- I tried pushing to OSF, then noticed that they have a new 5GB limitation, which I exceeded (EDIT: the cap might be [50GB](https://help.osf.io/hc/en-us/articles/360054528874-OSF-Storage-Caps) for *public* repos, maybe this step and the one below occurred in opposite order)
- in the process, [my OSF project](https://osf.io/cj2dr/) got marked as SPAM :-) (meaning they automatically set it to _private_ ... I need to discuss with their support now)
- Now trying to do it via [GIN](https://gin.g-node.org/sappelhoff/eeg_matchingpennies)
- If that doesn't work, I'll upload a small archive to the MNE OSF account as the easiest solution

### DIFF
--- a/LICENSE.txt
+++ b/LICENSE.txt
@@ -1,4 +1,4 @@
-Copyright © 2019-2021, authors of MNE-BIDS-Pipeline
+Copyright © 2019-2022, authors of MNE-BIDS-Pipeline
 All rights reserved.
 
 Redistribution and use in source and binary forms, with or without
--- a/tests/datasets.py
+++ b/tests/datasets.py
@@ -27,9 +27,9 @@ DATASET_OPTIONS: Dict[str, DATASET_OPTIONS_T] = {
         'exclude': []
     },
     'eeg_matchingpennies': {
-        'git': 'https://github.com/sappelhoff/eeg_matchingpennies',
+        'git': 'https://gin.g-node.org/sappelhoff/eeg_matchingpennies',
         'openneuro': '',
-        'osf': '',
+        'osf': '',  # original dataset: 'cj2dr'
         'web': '',
         'include': ['sub-05'],
         'exclude': []


sappelhoff (commented):
this currently fails because of tempita being removed :-)

but good news: I tested this locally, and it seems to work.

So even if it turns out that I prematurely ditched OSF, because they might have 50GB storage after all, I think this is still an improvement. (at least in terms of my maintenance burden, because GIN is easier to work than OSF)

hoechenberger:
Wow this seems like a super smooth transition on the test suite's side!

sappelhoff:
> Wow this seems like a super smooth transition on the test suite's side!

I am also surprised, I really like it. Now I just need the OSF team to make https://osf.io/cj2dr/ public again ... and then I'll have that page redirect to GIN.

sappelhoff:
also works in the bids-validator test suite: https://github.com/bids-standard/bids-validator/blob/4feb91b8aa9506c123d38327485dd5d4cd69e80b/bids-validator/bids_validator/test_bids_validator.py#L16-L18

agramfort (approved):
@hoechenberger merge if happy

PR #510, opened 2022-01-25
Try to fix build with upstream MNE changes

hoechenberger:
### Before merging …

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -9,6 +9,7 @@ scikit-learn
 pandas
 json_tricks
 h5py
+pymatreader
 matplotlib
 seaborn
 nibabel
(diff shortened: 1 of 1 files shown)


requirements.txt
```diff
@@ -9,6 +9,7 @@ scikit-learn
 pandas
 json_tricks
 h5py
+pymatreader
```
agramfort, line 12: should not be necessary

https://github.com/mne-tools/mne-python/blob/main/requirements_hdf5.txt

hoechenberger: Oh, good catch

hoechenberger:
```suggestion
```

hoechenberger:
@agramfort `autoreject` tries to make use of `mne.externals`, hence doesn't work with MNE `main`

agramfort (approved):

hoechenberger:
See https://github.com/autoreject/autoreject/issues/258

Until this has been resolved, our test suite here will partially fail

hoechenberger:
I'm merging this now anyway 😅

agramfort:
see https://github.com/autoreject/autoreject/pull/259

hoechenberger:
Great!!

PR #514, opened 2022-02-13
Enh landmarks config

agramfort:
to be merged after https://github.com/mne-tools/mne-bids/pull/955

### DIFF
--- a/scripts/source/_04_make_inverse.py
+++ b/scripts/source/_04_make_inverse.py
@@ -50,7 +50,7 @@ def run_inverse(*, cfg, subject, session=None):
     forward = mne.read_forward_solution(fname_fwd)
     inverse_operator = make_inverse_operator(info, forward, cov, loose=0.2,
                                              depth=0.8, rank='info')
-    write_inverse_operator(fname_inv, inverse_operator)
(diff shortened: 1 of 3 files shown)


scripts/source/_04_make_inverse.py
```diff
@@ -50,7 +50,7 @@ def run_inverse(*, cfg, subject, session=None):
     forward = mne.read_forward_solution(fname_fwd)
     inverse_operator = make_inverse_operator(info, forward, cov, loose=0.2,
                                              depth=0.8, rank='info')
-    write_inverse_operator(fname_inv, inverse_operator)
+    write_inverse_operator(fname_inv, inverse_operator, overwrite=True)
```
agramfort, line 53: I don't know why we never saw this issue

hoechenberger: This was only changed upstream recently, and in fact I ran into this issue and fixed it a while back, but the change somehow got lost during a rebase. Thanks for fixing it!

hoechenberger (approved):

hoechenberger:
Thank you, @agramfort!

PR #537, opened 2022-04-19
Adds support for arbitrary contrasts

### LINKED ISSUE #536: Proposal to allow more complex contrasts
For my use case, direct contrasts `cond_A - cond_B` are not the most informative.

I was wondering whether we could implement more complex contrasts, e.g. `-.5*cond_1 -.5*cond_2 + cond_3` or similar? `nilearn` does this with [expression_to_contrast_vector](https://github.com/nilearn/nilearn/blob/d43e4ff0e150634c474f834d523ada1529d218d1/nilearn/glm/contrasts.py#L21) by leveraging [pandas.DataFrame.eval](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.eval.html#pandas-dataframe-eval), see [the relevant line here](https://github.com/nilearn/nilearn/blob/d43e4ff0e150634c474f834d523ada1529d218d1/nilearn/glm/contrasts.py#L43)

I understand that it would make safe-naming contrasts tedious, so maybe much like [conditions](https://github.com/mne-tools/mne-bids-pipeline/blob/c40a9f044f33394a6dd73b83f79616cb1d427e80/config.py#L839) we should allow (force?) `contrasts` to be of type `Dict[str, str]`, the first `str` being a name and the second an expression to evaluate at runtime (I don't think we could enforce anything better with the type system available). Then in `config.py` one could write:

```python
contrasts = {"main_effect": "-.5*cond_1 - .5*cond_2 + cond_3",
             "gradient": [...],
             "simple": "cond_1 - cond_2"}
```

I wouldn't mind working on a pull request to do this, with some early input to make sure this is relevant to the maintainers, in particular I think it's worth considering the following:
* This may change some types in `config.py` so potentially break things
* What would be relevant tests for this?

mathias-sm:
As discussed in issue number mne-tools/mne-bids-pipeline#536, this adds the possibility to handle arbitrary contrasts by:
1. Defining a new type for arbitrary contrasts: a specific dict with a name, a list of condition and a list of weights
2. Enforcing that each `contrast` is well formed in the `checks`
3. Defining two getter functions:
    * `get_all_contrasts()` to normalize all contrasts to the new dict types
    * `get_decoding_contrasts()` to normalize compatible contrasts to binary contrasts to be used in decoding
4. Replace `contrast=config.contrasts` with either `contrast=config.get_all_contrasts()` or `get_decoding_contrasts()` accordingly, and update the code to use the `name` field whenever appropriate

I don't like forcing all conditions referenced in a `contrast` to be present in `conditions`. My use case involves a single (complex) contrast over many conditions, but if I put them all in `conditions` then they pop up everywhere in the reports in a way that is minimally useful but takes a lot of time (and space!). This PR removes that constraint, but left two related TODOs to discuss this point.

NB : I have not added new tests of this feature yet. Let's first confirm that the old tests still pass, and then update a test config files to use the new featyre.

Since this was by default in the PR template, I updated it, sorry if this is not appropriate:

Closes #536

### DIFF
--- a/config.py
+++ b/config.py
@@ -2603,10 +2603,7 @@ def get_all_contrasts() -> Iterable[ArbitraryContrast]:
             })
         else:
             normalized_contrasts.append(contrast)
-    if len(normalized_contrasts) == 0:
-        return None
-    else:
-        return normalized_contrasts
+    return normalized_contrasts
 
 
 def get_decoding_contrasts() -> Iterable[Tuple[str, str]]:
@@ -2621,10 +2618,7 @@ def get_decoding_contrasts() -> Iterable[Tuple[str, str]]:
                 cond_1 = contrast["conditions"][0]
                 cond_2 = contrast["conditions"][1]
                 normalized_contrasts.append((cond_1, cond_2))
-    if len(normalized_contrasts) == 0:
-        return None
-    else:
-        return normalized_contrasts
+    return normalized_contrasts
 
 
 def failsafe_run(
--- a/config.py
+++ b/config.py
@@ -2007,7 +2007,8 @@ if bem_mri_images not in ('FLASH', 'T1', 'auto'):
            f'are: "FLASH", "T1", and "auto".')
     raise ValueError(msg)
 
-if contrasts:
+
+def _validate_contrasts(contrasts):
     for contrast in contrasts:
         if isinstance(contrast, tuple):
             if len(contrast) != 2:
@@ -2593,6 +2594,7 @@ def get_fs_subjects_dir():
 
 
 def get_all_contrasts() -> Iterable[ArbitraryContrast]:
+    _validate_contrasts(contrasts)
     normalized_contrasts = []
     for contrast in contrasts:
         if isinstance(contrast, tuple) and len(contrast) == 2:
@@ -2607,6 +2609,7 @@ def get_all_contrasts() -> Iterable[ArbitraryContrast]:
 
 
 def get_decoding_contrasts() -> Iterable[Tuple[str, str]]:
+    _validate_contrasts(contrasts)
     normalized_contrasts = []
     for contrast in contrasts:
         if isinstance(contrast, tuple) and len(contrast) == 2:
--- a/config.py
+++ b/config.py
@@ -2617,7 +2617,7 @@ def get_all_contrasts() -> Iterable[ArbitraryContrast]:
     _validate_contrasts(contrasts)
     normalized_contrasts = []
     for contrast in contrasts:
-        if isinstance(contrast, tuple) and len(contrast) == 2:
+        if isinstance(contrast, tuple):
             normalized_contrasts.append({
                 'name': contrast[0] + "+" + contrast[1],
                 'conditions': list(contrast),
@@ -2632,7 +2632,7 @@ def get_decoding_contrasts() -> Iterable[Tuple[str, str]]:
     _validate_contrasts(contrasts)
     normalized_contrasts = []
     for contrast in contrasts:
-        if isinstance(contrast, tuple) and len(contrast) == 2:
+        if isinstance(contrast, tuple):
             normalized_contrasts.append(contrast)
         else:
             check_len = len(contrast["conditions"]) == 2
--- a/config.py
+++ b/config.py
@@ -2600,7 +2600,7 @@ def get_all_contrasts() -> Iterable[ArbitraryContrast]:
                 'name': contrast[0] + "+" + contrast[1],
                 'conditions': list(contrast),
                 'weights': [1, -1],
-                })
+            })
         else:
             normalized_contrasts.append(contrast)
     if len(normalized_contrasts) == 0:
--- a/config.py
+++ b/config.py
@@ -923,14 +923,24 @@ contrasts: Iterable[
     ]
 ] = []
 """
-The conditions to contrast via a subtraction of ERPs / ERFs. The list has either
-tuples, or ArbitraryContrast dicts. Each element in the list corresponds to one
-contrast. A tuple is a one vs one contrast. A ArbitraryContrast is a dict that
-defines a name, a list of conditions, and the list of associated weights. Pass
-an empty list to avoid calculation of contrasts.
-For the contrasts to be computed, the appropriate condutions must have been
+The conditions to contrast via a subtraction of ERPs / ERFs. The list elements
+can either be tuples or dictionaries (or a mix of both). Each element in the
+list corresponds to a single contrast.
+
+A tuple specifies a one-vs-one contrast, where the second condition is
+subtraced from the first.
+
+If a dictionary, must contain the following keys:
+
+- `name`: a custom name of the contrast
+- `conditions`: the conditions to contrast
+- `weights`: the weights associated with each condition.
+
+Pass an empty list to avoid calculation of any contrasts.
+
+For the contrasts to be computed, the appropriate conditions must have been
 epoched, and therefore the conditions should either match or be subsets of
-`conditions` above
+`conditions` above.
 
 ???+ example "Example"
     Contrast the "left" and the "right" conditions by calculating
--- a/config.py
+++ b/config.py
@@ -916,7 +916,12 @@ if ``None``, no baseline correction is applied.
     ```
 """
 
-contrasts: Iterable[Union[Tuple[str, str], ArbitraryContrast]] = []
+contrasts: Iterable[
+    Union[
+        Tuple[str, str],
+        ArbitraryContrast
+    ]
+] = []
 """
 The conditions to contrast via a subtraction of ERPs / ERFs. The list has either
 tuples, or ArbitraryContrast dicts. Each element in the list corresponds to one
--- a/config.py
+++ b/config.py
@@ -2635,6 +2635,10 @@ def get_decoding_contrasts() -> Iterable[Tuple[str, str]]:
         if isinstance(contrast, tuple):
             normalized_contrasts.append(contrast)
         else:
+            # If a contrast is an `ArbitraryContrast` and satisfies
+            # * has exactly two conditions (`check_len`)
+            # * weights sum to 0 (`check_sum`)
+            # Then the two conditions are used to perform decoding
             check_len = len(contrast["conditions"]) == 2
             check_sum = np.isclose(np.sum(contrast["weights"]), 0)
             if check_len and check_sum:
--- a/config.py
+++ b/config.py
@@ -2028,21 +2028,22 @@ if bem_mri_images not in ('FLASH', 'T1', 'auto'):
     raise ValueError(msg)
 
 
+_keys_arbitrary_contrast = set(ArbitraryContrast.__required_keys__)
+
+
 def _validate_contrasts(contrasts):
     for contrast in contrasts:
         if isinstance(contrast, tuple):
             if len(contrast) != 2:
                 raise ValueError("Contrasts' tuples MUST be two conditions")
-        else:
-            if "name" not in contrast.keys():
-                raise ValueError("One contrast is missing a name")
-            if "conditions" not in contrast.keys():
-                raise ValueError("One contrast is missing a condition list")
-            if "weights" not in contrast.keys():
-                raise ValueError("One contrast is missing a weight list")
+        elif isinstance(contrast, dict):
+            if not _keys_arbitrary_contrast.issubset(set(contrast.keys())):
+                raise ValueError(f"Missing key(s) in contrast {contrast}")
             if len(contrast["conditions"]) != len(contrast["weights"]):
-                raise ValueError(f"Contrast {contrast['name']} has an"
+                raise ValueError(f"Contrast {contrast['name']} has an "
                                  f"inconsistent number of conditions/weights")
+        else:
+            raise ValueError("Contrasts must be tuples or well-formed dicts")
 
 
 def check_baseline(
@@ -2618,11 +2619,13 @@ def get_all_contrasts() -> Iterable[ArbitraryContrast]:
     normalized_contrasts = []
     for contrast in contrasts:
         if isinstance(contrast, tuple):
-            normalized_contrasts.append({
-                'name': contrast[0] + "+" + contrast[1],
-                'conditions': list(contrast),
-                'weights': [1, -1],
-            })
+            normalized_contrasts.append(
+                ArbitraryContrast(
+                    name=(contrast[0] + "+" + contrast[1]),
+                    conditions=list(contrast),
+                    weights=[1, -1]
+                )
+            )
         else:
             normalized_contrasts.append(contrast)
     return normalized_contrasts
--- a/config.py
+++ b/config.py
@@ -36,6 +36,12 @@ from mne_bids import BIDSPath, read_raw_bids
 PathLike = Union[str, pathlib.Path]
 
 
+class ArbitraryContrast(TypedDict):
+    name: str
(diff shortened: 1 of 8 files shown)


hoechenberger:
@mathias-sm LMK when you would like me to review!

mathias-sm:
The version from this morning should be OK for review. I don't think the tests failure is related to my commits.

config.py
```diff
@@ -1987,6 +2007,22 @@ def gen_log_kwargs(
...
+        else:
+            if "name" not in contrast.keys():
+                raise ValueError("One contrast is missing a name")
+            if "conditions" not in contrast.keys():
+                raise ValueError("One contrast is missing a condition list")
+            if "weights" not in contrast.keys():
+                raise ValueError("One contrast is missing a weight list")
+            if len(contrast["conditions"]) != len(contrast["weights"]):
+                raise ValueError(f"Contrast {contrast['name']} has an"
+                                 f"inconsistent number of conditions/weights")
```
agramfort, line 2024: I don’t think validation should be done here. It needs to be done also when you have a custom config file so after the first import of config

config.py
```diff
@@ -2556,6 +2592,35 @@ def get_fs_subjects_dir():
                 .resolve())
 
 
+def get_all_contrasts() -> Iterable[ArbitraryContrast]:
```
agramfort, line 2595: I suggest to do the validation here

agramfort (commented):
Looks clean !

config.py
mathias-sm: OK, I can see why and I'm happy to move it there. My reasoning was that moving the checks in `get`ters would duplicate the logic, as we have two getters, but I could abstract away a `_validate_contrasts(...)` to be called from both.

agramfort:
Yes ! 🙏

mathias-sm:
Actually I'm not sure, that change will mean that we validate the contrasts in `config` with each step that calls a `get`ter, which is suboptimal. Is there a place to validate the `config` as a whole?

hoechenberger:
> that change will mean that we validate the contrasts in `config` with each step that calls a `get`ter, which is suboptimal.

I don't think it's an issue really, at least not in terms of performance impact

config.py
```diff
@@ -934,6 +942,18 @@
                  ('visual/left', 'visual/right'),
                  ('auditory', 'visual')]
     ```
+
+    Contrast the "left" and the "right" regardless of side, and compute an
+    arbitrary contrast with a gradient of weights:
+    ```python
+    conditions = ['auditory/left', 'auditory/right',
```
agramfort, line 949: Why do you need both conditions and contrasts?

hoechenberger: We can only calculate contrasts from conditions that were previously specified in `conditions`; otherwise the indexing we do fails.

config.py
```diff
@@ -934,6 +942,18 @@
...
+                  'visual/left', 'visual/right']
+    contrasts = [('auditory/left', 'auditory/right'),
+                 {'name': 'gradedContrast',
+                  'conditions': ['auditory/left', 'auditory/right',
+                                 'visual/left', 'visual/right'],
+                  'weights': [-1.5, -.5, .5, 1.5]}]
```
hoechenberger, line 955:
```suggestion
    contrasts = [
        ('auditory/left', 'auditory/right'),
        {
            'name': 'gradedContrast',
            'conditions': [
                'auditory/left',
                'auditory/right',
                'visual/left',
                'visual/right'
            ],
            'weights': [-1.5, -.5, .5, 1.5]
        }
    ]
```

config.py
mathias-sm: We don't, that's why I left as a "TODO" just above. So far, contrasts stated `The condition names must be specified in ``conditions`` above.`, and indeed the code used `all_evoked` computed on `conditions` to compute the contrasts. I think this is too restrictive, so I don't do this anymore and instead go back to the `epochs`, and wanted your input on this.

agramfort: agreed. Using epochs seems best !

mathias-sm: OK, so I removed the `TODOs`, and cleaned (all!) examples of `contrasts`.

agramfort (approved):
@hoechenberger I let you review/merge if happy

thx @mathias-sm !

hoechenberger:
Will take a look after lunch break 🥙😎☕️

config.py
```diff
@@ -910,30 +916,49 @@
     ```
 """
 
-contrasts: Iterable[Tuple[str, str]] = []
+contrasts: Iterable[Union[Tuple[str, str], ArbitraryContrast]] = []
```
hoechenberger, line 919: Let's do it like this until we drop support for Python <3.10 and can replace `Union` with `|`

```suggestion
contrasts: Iterable[
    Union[
        Tuple[str, str],
        ArbitraryContrast
    ]
] = []
```

config.py
```diff
@@ -910,30 +916,49 @@
...
-specified in ``conditions`` above. Pass an empty list to avoid calculation
-of contrasts.
+The conditions to contrast via a subtraction of ERPs / ERFs. The list has either
+tuples, or ArbitraryContrast dicts. Each element in the list corresponds to one
+contrast. A tuple is a one vs one contrast. A ArbitraryContrast is a dict that
+defines a name, a list of conditions, and the list of associated weights. Pass
+an empty list to avoid calculation of contrasts.
+For the contrasts to be computed, the appropriate condutions must have been
+epoched, and therefore the conditions should either match or be subsets of
+`conditions` above
```
hoechenberger, line 928:
```suggestion
The conditions to contrast via a subtraction of ERPs / ERFs. The list elements
can either be tuples or dictionaries (or a mix of both). Each element in the
list corresponds to a single contrast.

A tuple specifies a one-vs-one contrast, where the second condition is
subtraced from the first.

If a dictionary, must contain the following keys:

- `name`: a custom name of the contrast
- `conditions`: the conditions to contrast
- `weights`: the weights associated with each condition.

Pass an empty list to avoid calculation of any contrasts.

For the contrasts to be computed, the appropriate conditions must have been
epoched, and therefore the conditions should either match or be subsets of
`conditions` above.
```

config.py
```diff
@@ -1988,6 +2013,23 @@ def gen_log_kwargs(
...
+        else:
+            if "name" not in contrast.keys():
+                raise ValueError("One contrast is missing a name")
+            if "conditions" not in contrast.keys():
+                raise ValueError("One contrast is missing a condition list")
+            if "weights" not in contrast.keys():
+                raise ValueError("One contrast is missing a weight list")
+            if len(contrast["conditions"]) != len(contrast["weights"]):
+                raise ValueError(f"Contrast {contrast['name']} has an"
+                                 f"inconsistent number of conditions/weights")
```
hoechenberger, line 2030: Can we avoid these checks by just trying to cast the passed `dict` to an `ArbitraryContrast`?

config.py
```diff
@@ -2556,6 +2598,37 @@ def get_fs_subjects_dir():
                 .resolve())
 
 
+def get_all_contrasts() -> Iterable[ArbitraryContrast]:
+    _validate_contrasts(contrasts)
+    normalized_contrasts = []
+    for contrast in contrasts:
+        if isinstance(contrast, tuple) and len(contrast) == 2:
```
hoechenberger, line 2605: By here it **must** by of length 2; unless our validation is broken
```suggestion
        if isinstance(contrast, tuple):
```

config.py
```diff
@@ -2556,6 +2598,37 @@ def get_fs_subjects_dir():
...
+            normalized_contrasts.append({
+                'name': contrast[0] + "+" + contrast[1],
+                'conditions': list(contrast),
+                'weights': [1, -1],
```
hoechenberger, line 2609: Why don't you create an `ArbitraryContrast` directly?

config.py
```diff
@@ -2556,6 +2598,37 @@ def get_fs_subjects_dir():
...
+        else:
+            normalized_contrasts.append(contrast)
+    return normalized_contrasts
+
+
+def get_decoding_contrasts() -> Iterable[Tuple[str, str]]:
+    _validate_contrasts(contrasts)
+    normalized_contrasts = []
+    for contrast in contrasts:
+        if isinstance(contrast, tuple) and len(contrast) == 2:
```
hoechenberger, line 2620: Same as above
```suggestion
        if isinstance(contrast, tuple):
```

config.py
```diff
@@ -2556,6 +2598,37 @@ def get_fs_subjects_dir():
...
+            normalized_contrasts.append(contrast)
+        else:
+            check_len = len(contrast["conditions"]) == 2
+            check_sum = np.isclose(np.sum(contrast["weights"]), 0)
+            if check_len and check_sum:
```
hoechenberger, line 2625: Can you please add a comment to the code, explaining why we need this?

scripts/report/_01_make_reports.py
```diff
@@ -492,7 +492,7 @@ def run_report_sensor(
     else:
         conditions = cfg.conditions.copy()
 
-    conditions.extend(cfg.contrasts)
+    conditions.extend([c["name"] for c in cfg.all_contrasts])
```
hoechenberger, line 495: More explicit:
```suggestion
    conditions.extend([contrast["name"] for contrast in cfg.all_contrasts])
```

scripts/report/_01_make_reports.py
```diff
@@ -712,7 +711,7 @@ def run_report_source(
...
```
hoechenberger, line 714: Same as above:
```suggestion
    conditions.extend([contrast["name"] for contrast in cfg.all_contrasts])
```

scripts/report/_01_make_reports.py
```diff
@@ -752,10 +751,9 @@ def run_report_source(
                                      subject=subject, session=session))
 
         if condition in cfg.conditions:
-            title = f'Source: {config.sanitize_cond_name(condition)}'
+            title = f'Condition: {config.sanitize_cond_name(condition)}'
         else:  # It's a contrast of two conditions.
-            # XXX Will change once we process contrasts here too
-            continue
+            title = f'Contrast: {config.sanitize_cond_name(condition)}'
```
hoechenberger, line 756: Can you please revert this? We don't support contrasts in source space yet, and naming the source estimates `Condition` collides with our evoked data

scripts/report/_01_make_reports.py
```diff
@@ -906,7 +904,7 @@ def run_report_average(*, cfg, subject: str, session: str) -> None:
     else:
         conditions = cfg.conditions.copy()
 
-    conditions.extend(cfg.contrasts)
+    conditions.extend([c["name"] for c in cfg.all_contrasts])
```
hoechenberger, line 907: Same as above
```suggestion
    conditions.extend([contrast["name"] for contrast in cfg.all_contrasts])
```

scripts/report/_01_make_reports.py
```diff
@@ -996,8 +993,13 @@ def run_report_average(*, cfg, subject: str, session: str) -> None:
...
         else:  # It's a contrast of two conditions.
-            # XXX Will change once we process contrasts here too
-            continue
+            title = f'Average contrast: {condition}'
+            cond_str = config.sanitize_cond_name(condition)
+            tags = (
+                'source-estimate',
+                'contrast',
+                config.sanitize_cond_name(condition).lower().replace(' ', '')
+            )
```
hoechenberger, line 1002: Please revert or comment out: we do not yet support source-level contrasts

scripts/sensor/_01_make_evoked.py
```diff
@@ -69,11 +69,10 @@ def run_evoked(*, cfg, subject, session=None):
                                      session=session))
 
         for contrast in cfg.contrasts:
-            cond_1, cond_2 = contrast
-            evoked_diff = mne.combine_evoked([all_evoked[cond_1],
-                                              all_evoked[cond_2]],
-                                             weights=[1, -1])
-            all_evoked[contrast] = evoked_diff
+            conditions = [epochs[x].average() for x in contrast["conditions"]]
```
hoechenberger, line 72: Please don't simply name a list of evokeds "conditions". I don't have a good alternative in mind right now, but `conditions` will lead to confusion.

scripts/sensor/_02_sliding_estimator.py
```diff
@@ -176,10 +176,11 @@ def main():
     # Here we go parallel inside the :class:`mne.decoding.SlidingEstimator`
     # so we don't dispatch manually to multiple jobs.
     logs = []
+    contrasts = config.get_decoding_contrasts()
     for subject, session, (cond_1, cond_2) in itertools.product(
         config.get_subjects(),
         config.get_sessions(),
-        config.contrasts
+        contrasts
```
hoechenberger, line 183:
```suggestion
        contrasts=config.get_decoding_contrasts()
```

scripts/source/_04_make_inverse.py
```diff
@@ -74,6 +74,8 @@ def run_inverse(*, cfg, subject, session=None):
     else:
         conditions = cfg.conditions
 
+    conditions.extend([c["name"] for c in cfg.contrasts])
```
hoechenberger, line 77: More explicit
```suggestion
    conditions.extend([contrast["name"] for contrast in cfg.contrasts])
```

scripts/source/_05_group_average.py
```diff
@@ -44,6 +44,8 @@ def morph_stc(cfg, subject, fs_subject, session=None):
         else:
             conditions = cfg.conditions
 
+    conditions.extend([c["name"] for c in cfg.contrasts])
```
hoechenberger, line 47: More explicit
```suggestion
    conditions.extend([contrast["name"] for contrast in cfg.contrasts])
```

scripts/source/_05_group_average.py
```diff
@@ -89,6 +91,8 @@ def run_average(cfg, session, mean_morphed_stcs):
     else:
         conditions = cfg.conditions
 
+    conditions.extend([c["name"] for c in cfg.contrasts])
```
hoechenberger, line 94: More explicit
```suggestion
    conditions.extend([contrast["name"] for contrast in cfg.contrasts])
```

scripts/report/_01_make_reports.py
mathias-sm: Is there something specific about contrasts and source space, beyond "just" using the appropriate combined evoked? If not, then I implemented _that_. Sorry if I'm getting this wrong.

agramfort (approved):
LGTM ! thx @mathias-sm 

I'll let @hoechenberger merge / edit if he feels like it

scripts/report/_01_make_reports.py
hoechenberger: We discussed this in person and the gist is that we should address source level contrasts in a separate PR (in fact, there's already one)

config.py
mathias-sm: Done

config.py
mathias-sm: Can't find a straightforward way to bend python's type system to my will and do that.... Still streamlined that check

scripts/report/_01_make_reports.py
mathias-sm: Thanks, reverted

mathias-sm:
OK, this version feels final to me. Many comments and commits for relatively few changes in the end, should we squash those? LMK what you prefer.

hoechenberger:
No please do NOT squash, I'm in the middle of a review 😅

We always squash when merging, GH does that for us automatically.

agramfort:
thx @mathias-sm !

PR #546, opened 2022-05-02
Try to fix CI

hoechenberger:
Something is wrong with CI, but I cannot reproduce it locally. Buckle up for a journey…


hoechenberger:
Closing in favor of #546

PR #547, opened 2022-05-02
Try to make parallelization work again with latest MNE-Python development version

hoechenberger:
Let's see if CI passes then…


hoechenberger:
@larsoner 

Since https://github.com/mne-tools/mne-python/pull/10567, two thirds of our CI jobs fail due to resource overconsumption. Locally on my machine everything looks normal, and memory profiling doesn't reveal anything awfully bad. Yet, on Circle, the jobs seem to exceed the memory restrictions.

Reverting MNE to before https://github.com/mne-tools/mne-python/pull/10567 fixes the issue.

larsoner:
To me it looks like you have some `parallel_func` code that should have been integrated into MNE-Python (DRY) that probably does not play nicely with what MNE-Python does/uses.

That issue aside, one possibility is that the use of `parallel_backend` in MNE-BIDS-Pipeline causes MNE-Python to now see that there are, say, 8 jobs available for parallelization where previously there was only one. So if you parallelize over subjects for example in MNE-BIDS-Pipeline, you should set `n_jobs=1` to your MNE-Python calls to avoid overconsumption. Otherwise you could end up with parallelization over 8 subjects for example, where each of those threads/processes thinks in MNE-Python that it should be using `n_jobs=8` in filtering for example.

hoechenberger:
But we're using a parallel context manager, so this problem shouldn't occur, no..?

larsoner:
The problem is that when MNE-Python thinks it has `n_jobs>1` the code path changes to one that can use up to 2x the memory, e.g.:

https://github.com/mne-tools/mne-python/blob/00e96d940ac666f66c5a268f17564c470e141236/mne/filter.py#L204-L212

So if you were using 2GB on CircleCI before, you're using 4GB now and it dies

larsoner:
... or that is *at least one* problem, not sure if it's the cause of the failures here or not.

hoechenberger:
But wasn't the idea of switching to `n_jobs=None` to prevent exactly this kind of behavior? We are running all MNE functions inside a `joblib.parallel_backend()` context, and keeping `n_jobs` at the default (i.e., `None`) everywhere except during context creation.

hoechenberger:
I'm also realizing now that many more checks are passing now after the adjustments I made here; only 2 jobs that are using loky still fail (ds000246, ds003392), and 7 that use Dask; but I'll ignore Dask for now

larsoner:
> But wasn't the idea of switching to n_jobs=None to prevent exactly this kind of behavior? We are running all MNE functions inside a joblib.parallel_backend() context, and keeping n_jobs at the default (i.e., None) everywhere except during context creation.

I don't think so. I think we are doing the right thing -- you are creating a context with `n_jobs>1` and now MNE-Python respects this, and uses a different, parallelized code path. One downside of respecting your context manager choice is that it now hits a code path that is less memory efficient.

hoechenberger:
I don't get it. I thought we wanted to force MNE to respect the processing limits imposed from the "outside", i.e., a parallel context. If you now say I have to take care of passing the "correct" `n_jobs` to MNE in addition to creating a parallel context, or otherwise MNE will blow up, what do we actually gain? I clearly must be missing something here. We had the same number of jobs before that change, and all worked fine; now it appears MNE is multiplying the number of jobs / required resources. This clearly couldn't have been the goal of this upstream change? Doesn't the parallel context not serve as a semaphore that's supposed to prevent **exactly** this kind of behavior? I'm extremely confused.

hoechenberger:
> I don't think so. I think we are doing the right thing -- you are creating a context with `n_jobs>1` and now MNE-Python respects this, and uses a different, parallelized code path.

For the failing `ds003392` test, we actually use `n_jobs=1` and yet it fails:

https://app.circleci.com/pipelines/github/mne-tools/mne-bids-pipeline/2862/workflows/d4b51bb0-3d2a-49bc-98f7-1b5a9375474a/jobs/26128

larsoner:
> I thought we wanted to force MNE to respect the processing limits imposed from the "outside", i.e., a parallel context.

It will/should. If you are in a parallel context with n_jobs=8, MNE should use up to 8 jobs now. Before it would use 1 unless you explicitly passed it `n_jobs=8`. When I said this above I was wrong:

> So if you parallelize over subjects for example in MNE-BIDS-Pipeline, you should set n_jobs=1 to your MNE-Python calls to avoid overconsumption. Otherwise you could end up with parallelization over 8 subjects for example, where each of those threads/processes thinks in MNE-Python that it should be using n_jobs=8 in filtering for example.

There shouldn't ever be for example 64 jobs running.

I *think* the problem here is the code path memory issue, but I'm not totally sure.

> For the failing ds003392 test, we actually use n_jobs=1 and yet it fails:

Indeed there might be some bug somewhere. If you are in no context or -- equivalently -- a `parallel_backend(..., n_jobs=1)` context, then we should stay in the memory-efficient code paths in MNE, and also in `parallel_func` saying the number of jobs being used. Perhaps we should add a `logger.debug` in our efficient and inefficient code paths, and then you can run these with `verbose='debug'`. It will be a lot to read but you can probably find if it's the issue or not.

hoechenberger:
@larsoner Disabling frequency filtering via 2df233c in fact fixes the problem…

hoechenberger:
Ok perfect, now the only stuff that's failing is using Dask. But Loki and no parallelization are both working correctly now.

hoechenberger:
@agramfort I've switched all CI tests to use `loky` (or no parallelization at all) for now, i.e., I've disabled Dask so everything passes again. I will merge this so I can proceed working on new features for the pipeline, and we should try to fix Dask at a later time.

PR #553, opened 2022-05-09
WIP: Highlight significantly different time periods in evoked plot

hoechenberger:
This is currently WIP.

I'm not happy with what I have right now – just showing the contrast with
some highlights. I don't think this is very useful.

However, I'll try to show plots of the evokeds both contrasted
conditions, which would probably make more sense.

Just leaving this here as a WIP for now.

### Before merging …


hoechenberger:
Not sure we'll want this anytime soon (or ever), closing.

PR #557, opened 2022-05-12
Correct hardcoded lowpass filter on raw data.

### LINKED ISSUE #556: Lowpass filter at 40Hz is hardcoded in the config.py file
In the preprocessing, during the first script (`maxfilter`), raw data is imported with the function `import_experimental_data`. But in this function, there is another function called `_find_bad_channels` in which the raw data is filtered with a lowpass at 40Hz. If I choose a higher h_freq in the config it is not taken into account.

Here is the link to this specific line:
https://github.com/mne-tools/mne-bids-pipeline/blob/main/config.py#L3168

apmellot:
### Before merging …

Closes #556.

### DIFF
--- a/config.py
+++ b/config.py
@@ -3138,8 +3138,6 @@ def _find_bad_channels(cfg, raw, subject, session, task, run) -> None:
 
     Modifies ``raw`` in-place.
     """
-    from joblib import parallel_backend
-
     if not (cfg.find_flat_channels_meg or cfg.find_noisy_channels_meg):
         return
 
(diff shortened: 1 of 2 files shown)


config.py
```diff
@@ -3165,10 +3165,11 @@ def _find_bad_channels(cfg, raw, subject, session, task, run) -> None:
 
     # Filter the data manually before passing it to find_bad_channels_maxwell()
     # This reduces memory usage!
```
agramfort, line 3167: Is this comment still relevant now? Can we have raw without preload up to this point to avoid memory issues?

sappelhoff (approved):

hoechenberger:
I'm sorry, I forgot about this one. I'll take a look ASAP, most probably tonight.

hoechenberger:
@apmellot FYI I'm looking into this right now!

hoechenberger:
We don't seem to run into any memory issues on Circle, which is generally a good sign. I've just pushed some small cleanups; once CI goes green, we can merge this. I left the old cold around (commented out) for reference in case we discover any future memory issues.

hoechenberger:
Argh, celebrated too early!! Now Circle fails, I suppose it's an out-of-memory issue. Will take a look.

hoechenberger:
Thank you, @apmellot!

agramfort:
thx @apmellot @hoechenberger 🙏

PR #558, opened 2022-05-12
Add decoding based on full epochs

hoechenberger:
### Before merging …


hoechenberger:
Reports:

- [grand average, single contrast](https://output.circle-artifacts.com/output/job/3e60caf3-e851-4666-ac65-97d1255f065b/artifacts/0/reports/ERP_CORE/sub-average_ses-N400_task-N400_report.html#global10)
- [grand average, multiple contrasts](https://output.circle-artifacts.com/output/job/2575281d-7490-4790-baec-16971040f8be/artifacts/0/reports/ds000117/sub-average_ses-meg_task-facerecognition_report.html#global17) (unfortunately only a single participant, hence the boxplots are not boxes at all…)
- [single subject](https://output.circle-artifacts.com/output/job/efce5b31-4df4-4a31-83b8-e3a7e0a76dfd/artifacts/0/reports/ds000246/sub-0001_task-AEF_report.html#global34)


@agramfort This is ready for review / merge

scripts/sensor/_02_decoding_full_epochs.py
```diff
@@ -0,0 +1,202 @@
+"""
+=================
+Decoding pairs of conditions based on entire epochs.
+=================
+
+A sliding estimator fits a separate logistic regression model for every time
+point. The end result is an averaging effect across sensors.
```
agramfort, line 7: This needs update I think

scripts/sensor/_02_decoding_full_epochs.py
```diff
@@ -0,0 +1,202 @@
...
+            Vectorizer(),  # So we can pass the data to scikit-learn
+            LogisticRegression(
+                solver='liblinear',  # much faster than the default
+                n_jobs=1
+            )
+        )
+
+        # Now, actually run the classification, and evaluate it via a
+        # cross-validation procedure.
+        cv = StratifiedKFold(shuffle=True, n_splits=cfg.decoding_n_splits)
```
agramfort, line 114: Using shuffle without fixing the seed is dangerous.

agramfort (commented):
Just a few nitpicks. Just I would not use a slider in the report as now the time by time is not visible by default and hence much harder to discover. 

🙏

scripts/sensor/_02_decoding_full_epochs.py
hoechenberger: Good catch!!! Simply forgot that one!

hoechenberger:
>  I would not use a slider in the report as now the time by time is not visible by default and hence much harder to discover.

Absolutely! This problem occurs in other scenarios as well, I believe. I'm planning to rework the slider for that reason.

But in this particular case, maybe I should display a facet grid of all full-epochs decoding results (for all conditions) like I'm already doing for the grand average. I'll rework it.

hoechenberger:
@agramfort I think we're good!

See
https://output.circle-artifacts.com/output/job/7dbc9ee8-6d6f-4ad5-9cf9-9b35ca00f573/artifacts/0/reports/ds000117/sub-01_ses-meg_task-facerecognition_report.html#global54

agramfort:
🙏

PR #562, opened 2022-06-10
Enable time generalization for ds000246 and ds003392

hoechenberger:


mwe.py
```diff
@@ -0,0 +1,21 @@
...
+    # steps=('sensor/decoding_entire_epochs',)
+    steps=('report',),
+    n_jobs='8',
+    # subject='015',
+    # steps=('sensor/decoding_time_by_time', 'sensor/group_average')
+    # steps=('sensor/sliding_estimator', 'sensor/group_average', 'report'),
+    # steps=('sensor/group_average', 'report'),
+    # task='ERN'
+    # steps=('sensor/group_average', 'report'),
+    # config='~/Development/mne-bids-pipeline/tests/configs/config_ds000248_no_mri.py',
```
agramfort, line 20: Was this merged on purpose ?

hoechenberger: Argh. Absolutely not. Should've waited for your review … will remove this.

hoechenberger: Deleted it directly on `main` now!

PR #563, opened 2022-06-14
ENH: Caching infrastructure, maxwell_fliter and frequency_filter steps

agramfort:
@hoechenberger let me know what you think

@larsoner if your curious...


larsoner:
Only *two* `cleanup` after `seems to work`, then things are green -- and at only +87/-4 -- this qualifies as a proper *tour de force* (it's a French phrase I believe, use Google Translate if you need to) compared to the [dumpster fires I produce](https://github.com/mne-tools/mne-python/pull/10763#commits-pushed-7a2aa87) :) I'll look with jet-lagged eyes soon...

hoechenberger:
Preserving @larsoner's dumpster fire for the next generation:

image


Apropos dumpsters and French, this is what Paris currently looks like thanks to a strike:

https://www.dailymotion.com/video/x38xy02


But I'm depressing.

agramfort:
I need to find the time to clean this up but I am happy that you like it too ;)

larsoner:
Too simple, I refuse to believe it works.

It does make me feel better that your next commit broke CIs, even if the failure is just spurious :)

larsoner:
@hoechenberger there is something weird going on with that dataset, maybe that version got deleted? Using the non-versioned name I can access https://s3.amazonaws.com/openneuro.org/ds000247/dataset_description.json

larsoner:
I think the natural next step here is to refactor out any input path usage **within** the function that does the heaving lifting to use `input_files['whatever']`. This will allow the `get_input_files=` for the decorator to operate on the same set of files as anything within the function.

In other words, within any processing script, 

1. Any implicit or explicit load/read should have to operate on `input_files[<something>]` that comes from `get_input_files(...)` (and not generate any such paths internally), and
2. Any implicit or explicit save/write should operate on `output_files[<something>]` internally, and this `output_files` dict should be returned by the function

larsoner:
@hoechenberger any idea how to fix the ds000247 issue?

https://app.circleci.com/jobs/github/mne-tools/mne-bids-pipeline/28377?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

It shows up in all PRs currently

hoechenberger:
I've restarted the ds000247 CI run, should hopefully work now

hoechenberger:
We're getting a 403 now ... progress!

scripts/preprocessing/_01_maxfilter.py
```diff
@@ -24,25 +24,85 @@
...
+                    import_experimental_data, import_er_data, import_rest_data)
 from config import parallel_func
 
 
 logger = logging.getLogger('mne-bids-pipeline')
 
 
-@failsafe_run(on_error=on_error, script_path=__file__)
-def run_maxwell_filter(*, cfg, subject, session=None, run=None):
+def _get_reference_run_path(
```
hoechenberger, line 38: what's with this function, it doesn't seem to be used anywhere?

larsoner: Indeed `git grep` says it's no longer used, feel free to push a commit to delete it!

config.py
```diff
@@ -2763,7 +2775,14 @@ def get_decoding_contrasts() -> Iterable[Tuple[str, str]]:
 def failsafe_run(
     on_error: OnErrorT,
     script_path: PathLike,
+    get_input_fnames: Optional[Callable] = None,
 ):
+    if memory_location is None or get_input_fnames is None:
+        # No caching is needed
+        memory = Memory(location=None)  # no op
```
larsoner, line 2782: FYI @agramfort you had this as `location='.'` which was not a no-op. I assume this was an oversight and not intentional... that, or I don't understand `Memory` or its docstring properly.

But I left the default of `memory_location='.'` in the config for now, which means by default after this PR users will see benefits and all caching will probably be in the `mne-bids-pipeline` directory

larsoner:
FYI I force-pushed so it's easier to see what I did to fix the path problem for `_02_frequency_filter.py`. I'll continue working with my data on this branch to see how far I get 👍

larsoner:
@hoechenberger @agramfort can we review+merge, assuming it comes back green? Otherwise reviewing will become very difficult.

FWIW this now successfully enables caching on the first two pipeline steps on a new (to me) dataset that I'm analyzing!

config.py
hoechenberger: We should dump the cache in the derivatives directory, don't you agree?

larsoner: Not sure which is better, but you're probably right. Maybe we alias `memory_location = True` to mean "in the derivatives folder"? I don't want to use a string default because the user might actually want that specific path (relative to `mne-bids-pipeline`)

tests/configs/config_ds000248.py
```diff
@@ -44,8 +44,10 @@ def noise_cov(bp):
 recreate_bem = True
 recreate_scalp_surface = True
 
-N_JOBS = 2
+# N_JOBS = 2
+N_JOBS = 1
 
+on_error = "debug"
```
agramfort, line 50: This should be reverted before merge

config.py
larsoner: This has been implemented as:
```
If not None (or False), caching will be enabled and the cache files will be
stored in the given directory. The default (True) will use a
``'.joblib_cache'`` subdirectory in the BIDS derivative root of the dataset.
```

larsoner:
Found a few more bugs that actually made the caching not work for all subjects. I switched to making sure it works all the way through for one subject and it did, so finally good to go here I think assuming CIs come back green!

config.py
```diff
@@ -3425,39 +3533,25 @@ def _fix_stim_artifact_func(
...
     cfg
         The local configuration.
-    subject
-        The subject to import.
-    session
-        The session to import.
-    run
-        The run to import.
+    bids_path_in
+        XXX
```
agramfort, line 3545: Still to fix

config.py
```diff
@@ -3481,52 +3575,58 @@ def import_experimental_data(
...
 
     Parameters
     ----------
     cfg
         The local configuration.
-    subject
-        The subject for whom to import the empty-room data.
-    session
-        The session for which to import the empty-room data.
+    bids_path_er_in
```
agramfort, line 3587: Do a pass on xxx

tests/configs/config_ds000248.py
```diff
@@ -44,7 +44,8 @@ def noise_cov(bp):
 recreate_bem = True
 recreate_scalp_surface = True
 
-N_JOBS = 2
+# N_JOBS = 2
+N_JOBS = 1
```
agramfort, line 48: To revert

agramfort (commented):
Also just use joblib default for cache dir name

larsoner:
Addressed @agramfort comments, okay for you @hoechenberger ?

hoechenberger:
@larsoner Could you add a changelog entry please?

agramfort:
Let’s add the entry in the next PR

Thx @larsoner

larsoner:
Where is the changelog? I didn't see it referenced in `CONTRIBUTING` and don't see a file with a name like that in the `docs` folder

larsoner:
Ahh docs/source/changes.md, I just missed it

PR #566, opened 2022-06-21
remove the save parameter that was not used

agramfort:
### Before merging …


larsoner:
Error is just the ds000247 one that happens on every PR, thanks @agramfort !

PR #567, opened 2022-06-22
Avoid printing nested tracebacks if on_error='abort'

hoechenberger:
So it's easier to figure out where and when the error occurred, without having to scroll around so much.


agramfort (approved):

larsoner:
Thanks @hoechenberger !

PR #569, opened 2022-07-06
WIP: Clean up & unify parallelization approach in decoding scripts

hoechenberger:
Unfortunately, no speed gains when going parallel within participants; so we avoid this.

I assume it's because we simply create too much overhead when distributing and later collecting these large numbers of small data chunks across cores.

But this PR at least makes it such that the decoding scripts follow the same parallelization approaches as all the other scripts.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -259,6 +259,12 @@ jobs:
 
   test_ds000246:
     <<: *imageconfig
+    # The resource_class feature allows configuring CPU and RAM resources for
+    # each job. Different resource classes are available for different
+    # executors.
+    #
+    # https://circleci.com/docs/2.0/configuration-reference/#resourceclass
+    resource_class: large
     steps:
       - attach_workspace:
           at: ~/
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -259,7 +259,7 @@ jobs:
 
   test_ds000246:
     <<: *imageconfig
-    no_output_timeout: 15m
+    no_output_timeout: 15m  # XXX should be revisited at some point
     steps:
       - attach_workspace:
           at: ~/
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -259,12 +259,6 @@ jobs:
 
   test_ds000246:
     <<: *imageconfig
-    # The resource_class feature allows configuring CPU and RAM resources for
-    # each job. Different resource classes are available for different
-    # executors.
-    #
-    # https://circleci.com/docs/2.0/configuration-reference/#resourceclass
-    resource_class: large
     no_output_timeout: 15m
     steps:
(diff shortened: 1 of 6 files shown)


.circleci/config.yml
```diff
@@ -259,6 +259,13 @@ jobs:
 
   test_ds000246:
     <<: *imageconfig
+    # The resource_class feature allows configuring CPU and RAM resources for
+    # each job. Different resource classes are available for different
+    # executors.
+    #
+    # https://circleci.com/docs/2.0/configuration-reference/#resourceclass
+    resource_class: large
+    no_output_timeout: 15m
```
larsoner, line 268: Just an idea -- IIRC the NumPy / SciPy build scripts print something every ~1 min or so while building their code. It helps let people know that nothing is broken during a long processing stage, and has the nice side effect that things like this are not necessary. Any idea if we could add something like this? I assume it's not trivial, in which case feel free to turn this into a new issue.

hoechenberger:
This makes things take longer on Circle, I don't understand why.

larsoner:
> This makes things take longer on Circle, I don't understand why.

Just a few possibilities I can think of based on a quick look at the changes:

1. it could be a bug with how we're handling n_jobs
2. it could be that the parallelization offered by Circle isn't truly parallel (cores are virtual and share an actual physical core somewhere)
3. it could be that the time-limiting steps were parallel before (e.g., in sklearn somewhere) so there is nothing to be gained by moving the paralellization "outward" to our level, and only losses to incur in terms of more data-passing overhead, etc. The comment `n_jobs=1,  # liblinear parallelizes automatically` suggests that joblib-based parallelization may now lead to an oversubscription of CPU cores, which has a cost in terms of switching (e.g., CPU cache can't be used as efficiently)
4. it could be that moving the pipeline instantiation out of a `with config.get_parallel_backend():` clause has made you miss out on some secret magic under the hood somehow

I'm not sure the best way to debug this. I would do it all locally first to see if you can replicate the slowdown. If you can, it eliminates (2) (or at least means 2 isn't the only issue). Then I'd go back to what you had before and make as small changes as possible while monitoring CPU use. If on `main` we already have all physical cores used almost all the time then I wouldn't expect to gain much from this PR -- but at least you shouldn't lose much if everything is implemented properly here either.

hoechenberger:
> 3\. The comment `n_jobs=1,  # liblinear parallelizes automatically` suggests that joblib-based parallelization may now lead to an oversubscription of CPU cores, which has a cost in terms of switching (e.g., CPU cache can't be used as efficiently)

No, this is excluded, as we control the number of BLAS threads via `inner_max_number_threads` (or something like that) :)

hoechenberger:
Giving up on this one for now, needs a fresh attempt sometime in the future.

PR #576, opened 2022-07-29
DOC: Add to changes

larsoner:
Pushing with a `circle full` since we hadn't done it for a while in #563 (whoops!)

### DIFF
--- a/scripts/preprocessing/_02_frequency_filter.py
+++ b/scripts/preprocessing/_02_frequency_filter.py
@@ -82,7 +82,7 @@ def get_input_fnames_frequency_filter(**kwargs):
     in_files[f'raw_run-{run}'] = bids_path_in
 
     # TODO: No need to process empty room (I guess?)
-    if cfg.noise_cov not in ('rest', 'emptyroom'):
+    if config.noise_cov not in ('rest', 'emptyroom'):
         return in_files
 
     noise_task = "rest" if config.noise_cov == "rest" else "noise"
@@ -94,11 +94,11 @@ def get_input_fnames_frequency_filter(**kwargs):
             raw_noise_fname_in.update(split='01')
         in_files["raw_noise"] = raw_noise_fname_in
     else:
-        if cfg.noise_cov == 'rest':
+        if config.noise_cov == 'rest':
             in_files["raw_rest"] = bids_path_in.copy().update(run=None,
                                                               task=noise_task)
         else:
-            assert cfg.noise_cov == 'noise'
+            assert config.noise_cov == 'noise'
             ref_bids_path = bids_path_in.copy().update(
                 run=cfg.mf_reference_run,
                 extension='.fif',
@@ -301,7 +301,6 @@ def get_config(
         drop_channels=config.drop_channels,
         find_breaks=config.find_breaks,
         min_break_duration=config.min_break_duration,
-        noise_cov=config.noise_cov,
         t_break_annot_start_after_previous_event=config.t_break_annot_start_after_previous_event,  # noqa:E501
         t_break_annot_stop_before_next_event=config.t_break_annot_stop_before_next_event,  # noqa:E501
     )
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -509,9 +509,9 @@ jobs:
             export DS=ds003392
             python tests/run_tests.py --download=0 ${DS}
             mkdir -p ~/reports/${DS}
-            cp ~/mne_data/derivatives/mne-bids-pipeline/${DS}/*/**/*.html ~/reports/${DS}/
-            cp ~/mne_data/derivatives/mne-bids-pipeline/${DS}/*/**/*.json ~/reports/${DS}/
-            cp ~/mne_data/derivatives/mne-bids-pipeline/${DS}/*/**/*.tsv ~/reports/${DS}/
+            cp ~/mne_data/derivatives/mne-bids-pipeline/${DS}/*/*/*.html ~/reports/${DS}/
+            cp ~/mne_data/derivatives/mne-bids-pipeline/${DS}/*/*/*.json ~/reports/${DS}/
+            cp ~/mne_data/derivatives/mne-bids-pipeline/${DS}/*/*/*.tsv ~/reports/${DS}/
             cp ~/mne_data/derivatives/mne-bids-pipeline/${DS}/*.xlsx ~/reports/${DS}/
       - store_artifacts:
           path: /home/circleci/reports/ds003392
--- a/docs/source/changes.md
+++ b/docs/source/changes.md
@@ -9,6 +9,7 @@ authors:
   dengemann: "[Denis A. Engemann](https://github.com/dengemann)"
   apmellot: "[Apolline Mellot](https://github.com/apmellot)"
   mathiassm: "[Mathias Sablé-Meyer](https://github.com/mathias-sm)"
+  larsoner: "[Eric Larson](https://github.com/larsoner)"
 ---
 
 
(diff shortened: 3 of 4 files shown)


larsoner:
@agramfort @hoechenberger feel free to push commits if you see the problem. None of these three are trivial to me :(

hoechenberger:
@larsoner
I think it's yet again an issue with OpenNeuro. I checked for ds000117 and the `*_meg.json` file is not being downloaded. Via their [web UI](https://openneuro.org/datasets/ds000117/versions/1.0.5), you can see that there are duplicates:

Screen Shot 2022-07-30 at 10 13 17

I believe they started appearing after https://github.com/OpenNeuroOrg/openneuro/issues/2635 was resolved.

Using the [GraphQL API](https://openneuro.org/crn/graphql) to query:

```graphql
# Write your query or mutation here
query {
  dataset(id: "ds000117") {
    latestSnapshot {
      id
      files(
        prefix: "sub-01/ses-meg/meg/"
      ) {
        urls
      }
    }
  }
}
```
indeed one can see that the duplicates are missing from the directory listing:
```
{
  "data": {
    "dataset": {
      "latestSnapshot": {
        "id": "ds000117:1.0.5",
        "files": [
          {
            "urls": [
              "https://openneuro.org/crn/datasets/ds000117/snapshots/12470d393b827e9b2d877554a9a8e8a20ce4ea41/files/sub-01:ses-meg:meg:sub-01_ses-meg_coordsystem.json"
            ]
          },
          {
            "urls": [
              "https://openneuro.org/crn/datasets/ds000117/snapshots/12470d393b827e9b2d877554a9a8e8a20ce4ea41/files/sub-01:ses-meg:meg:sub-01_ses-meg_task-facerecognition_run-01_events.tsv"
            ]
          },
          {
            "urls": [
              "https://openneuro.org/crn/datasets/ds000117/snapshots/12470d393b827e9b2d877554a9a8e8a20ce4ea41/files/sub-01:ses-meg:meg:sub-01_ses-meg_task-facerecognition_run-02_events.tsv"
            ]
          },
          {
            "urls": [
              "https://openneuro.org/crn/datasets/ds000117/snapshots/12470d393b827e9b2d877554a9a8e8a20ce4ea41/files/sub-01:ses-meg:meg:sub-01_ses-meg_task-facerecognition_run-03_events.tsv"
            ]
          },
          {
            "urls": [
              "https://openneuro.org/crn/datasets/ds000117/snapshots/12470d393b827e9b2d877554a9a8e8a20ce4ea41/files/sub-01:ses-meg:meg:sub-01_ses-meg_task-facerecognition_run-04_events.tsv"
            ]
          },
          {
            "urls": [
              "https://openneuro.org/crn/datasets/ds000117/snapshots/12470d393b827e9b2d877554a9a8e8a20ce4ea41/files/sub-01:ses-meg:meg:sub-01_ses-meg_task-facerecognition_run-05_events.tsv"
            ]
          },
          {
            "urls": [
              "https://openneuro.org/crn/datasets/ds000117/snapshots/12470d393b827e9b2d877554a9a8e8a20ce4ea41/files/sub-01:ses-meg:meg:sub-01_ses-meg_task-facerecognition_run-06_events.tsv"
            ]
          },
          {
            "urls": [
              "https://openneuro.org/crn/datasets/ds000117/snapshots/12470d393b827e9b2d877554a9a8e8a20ce4ea41/files/sub-01:ses-meg:meg:sub-01_ses-meg_headshape.pos"
            ]
          },
          {
            "urls": [
              "https://openneuro.org/crn/datasets/ds000117/snapshots/12470d393b827e9b2d877554a9a8e8a20ce4ea41/files/sub-01:ses-meg:meg:sub-01_ses-meg_task-facerecognition_run-01_meg.fif"
            ]
          },
          {
            "urls": [
              "https://openneuro.org/crn/datasets/ds000117/snapshots/12470d393b827e9b2d877554a9a8e8a20ce4ea41/files/sub-01:ses-meg:meg:sub-01_ses-meg_task-facerecognition_run-02_meg.fif"
            ]
          },
          {
            "urls": [
              "https://openneuro.org/crn/datasets/ds000117/snapshots/12470d393b827e9b2d877554a9a8e8a20ce4ea41/files/sub-01:ses-meg:meg:sub-01_ses-meg_task-facerecognition_run-03_meg.fif"
            ]
          },
          {
            "urls": [
              "https://openneuro.org/crn/datasets/ds000117/snapshots/12470d393b827e9b2d877554a9a8e8a20ce4ea41/files/sub-01:ses-meg:meg:sub-01_ses-meg_task-facerecognition_run-04_meg.fif"
            ]
          },
          {
            "urls": [
              "https://openneuro.org/crn/datasets/ds000117/snapshots/12470d393b827e9b2d877554a9a8e8a20ce4ea41/files/sub-01:ses-meg:meg:sub-01_ses-meg_task-facerecognition_run-05_meg.fif"
            ]
          },
          {
            "urls": [
              "https://openneuro.org/crn/datasets/ds000117/snapshots/12470d393b827e9b2d877554a9a8e8a20ce4ea41/files/sub-01:ses-meg:meg:sub-01_ses-meg_task-facerecognition_run-06_meg.fif"
            ]
          }
        ]
      }
    }
  },
  "extensions": {
    "openneuro": {
      "version": "4.8.2"
    },
    "cacheControl": {
      "version": 1,
      "hints": [
        {
          "path": [
            "dataset"
          ],
          "maxAge": 0
        }
      ]
    }
  }
}
```

hoechenberger:
I've opened an issue at https://github.com/OpenNeuroOrg/openneuro/issues/2643

hoechenberger:
Oh wait. Apparently the paths have changed

hoechenberger:
With the change I just pushed and with the changes from this yet-to-be-merged PR for MNE-BIDS, it works for me locally for ds000117:
https://github.com/mne-tools/mne-bids/pull/1030

scripts/preprocessing/_02_frequency_filter.py
```diff
@@ -301,7 +301,6 @@ def get_config(
         drop_channels=config.drop_channels,
         find_breaks=config.find_breaks,
         min_break_duration=config.min_break_duration,
-        noise_cov=config.noise_cov,
```
larsoner, line 304: I think the idea is that this local `cfg` should have everything that is needed to run and `get_input_files`, and this breaks that convention.

In terms of following this standard, if we do this (use of `config` directly) in other cases, we risk breaking the caching because `joblib` won't see any changes in `cfg` (I think). In this *particular* case it will still see the changes I think because the input files will change, for the most part. One case that isn't covered is if it's a callable -- now if the code of that callable changes, the caching won't work. (Actually I suspect it will show up as "always new" because by default the `__repr__` of a function includes its `id`/address, which will change on almost every run even if the code doesn't change. But maybe `joblib` is smarter about handling this already...?)

For now we could mark this with `# TODO FIX` just to get things green, though, since this seems like a complex problem with no easy solution, and the `callable` noise_cov case is rare relative to the others. If we do this, we might want to make it *always* recompute in the case of a callable `noise_cov` for safety purposes (i.e., we are not smart enough yet to know if someone changes the code within this function).

larsoner:
@hoechenberger it looks like you missed my comment above (your merge was only 4 min after my comment so understandable!) -- can you take a look?

scripts/preprocessing/_02_frequency_filter.py
hoechenberger: I'm currently on my phone so I cannot look, but I believe we're doing the same thing in a few other places where we accept callables, for example I think for that T1 path config option

But there might be others as well 

Finding a clean solution for all of these would be great.

larsoner:
Okay, thanks for the info and getting this green so quickly @hoechenberger . That would have have taken me quite some time to figure out...

PR #543, opened 2022-04-26
Various improvements to our documentation

hoechenberger:
- Use latest mkdocstrings-python so we get default parameter values in the docs
- Reduce table row height in example docs
- Fix icons used in the report download buttons in the example gallery
- Update the URL displayed by the browser when clicking on an anchor link in the docs, so it will be easier to share anchors
- Change section order on example pages
- Change order of documentation pages, so that the next page after the landing page is installation & getting started
- Fix missing / broken links to rendered examples in our gallery
- Some more minor stuff
    
Closes #196
Closes #447

PR #577, opened 2022-07-29
STY: Fix doc table line height

### LINKED ISSUE #575: DOC: CSS is messed up on some entries
For me this https://mne.tools/mne-bids-pipeline/features/steps.html looks like

Screenshot from 2022-07-28 19-08-48

larsoner:
Closes #575

@hoechenberger you added this in #543 where I see:

> Reduce table row height in example docs

Do you want to try pushing a different fix that doesn't cause the problem in #575, or just live with this revert?

### DIFF
--- a/docs/source/css/extra.css
+++ b/docs/source/css/extra.css
@@ -13,7 +13,6 @@ p {
 .md-typeset table td,
 .md-typeset table th {
   font-size: 0.7rem;
-  line-height: 0.5;
 }
 
 /* Larger font in code, example, info, warning blocks */
--- a/docs/source/css/extra.css
+++ b/docs/source/css/extra.css
@@ -13,6 +13,7 @@ p {
 .md-typeset table td,
 .md-typeset table th {
   font-size: 0.7rem;
+  padding: .5em 1.25em !important;
 }
 
 /* Larger font in code, example, info, warning blocks */


hoechenberger:
@larsoner This change was added to alter the "Demonstrated features" tables from this:
Screen Shot 2022-07-29 at 20 51 58

to this:
Screen Shot 2022-07-29 at 20 52 25

So maybe we need to apply a more restrictive CSS selector…

docs/source/css/extra.css
```diff
@@ -13,7 +13,6 @@ p {
 .md-typeset table td,
 .md-typeset table th {
   font-size: 0.7rem;
-  line-height: 0.5;
```
drammock, line 16:
```suggestion
  padding: .5em 1.25em !important;
```

larsoner:
Okay I didn't see any way to be more selective given the HTML, so I looked into wrapping it a bit. I figured I could add something to `gen_examples.py` to wrap the table in a div like `<div class="demonstrated-features-table">` but I can't get `mkdocs` render the markdown table *inside* that div.

@hoechenberger feel free to give it a shot if you know how. I was just going to change the `padding: .9375em 1.25em` to something smaller...

docs/source/css/extra.css
larsoner: Yes we could do this but it's not selective. But if @hoechenberger is okay with affecting all tables I'm happy to commit

hoechenberger:
All looking good. Thanks @larsoner!

PR #585, opened 2022-08-07
rest condition should have task name starting with rest not == to rest

agramfort:
ok for you @sappelhoff @hoechenberger ?

see at https://bids-specification.readthedocs.io/en/stable/04-modality-specific-files/01-magnetic-resonance-imaging-data.html

The task label included in the file name is derived from this "TaskName" field by removing all non-alphanumeric ([a-zA-Z0-9]) characters. For example "TaskName" "faces n-back" will correspond to task label facesnback. A RECOMMENDED convention is to name resting state task using labels beginning with rest.

### DIFF
--- a/config.py
+++ b/config.py
@@ -3127,7 +3127,7 @@ def make_epochs(
     - No rejection thresholds will be applied.
     - No baseline-correction will be performed.
     """
-    if task.lower() == 'rest':
+    if task.lower().startswith('rest'):
         stop = raw.times[-1] - rest_epochs_duration
         assert epochs_tmin == 0., "epochs_tmin must be 0 for rest"
         assert rest_epochs_overlap is not None, \
@@ -3786,7 +3786,7 @@ if 'MKDOCS' not in os.environ:
 
 # Another check that depends on some of the functions defined above
 if (get_task() is not None and
-        get_task().lower() != 'rest' and
+        not get_task().lower().startswith('rest') and
         conditions is None and
         'MKDOCS' not in os.environ):
     msg = ('Please indicate the name of your conditions in your '
--- a/config.py
+++ b/config.py
@@ -430,7 +430,7 @@ itself, nor to the source analysis stage.
 
 reader_extra_params: dict = {}
 """
-Parameters to be passed to read_raw_bids calls when importing raw data.
+Parameters to be passed to `read_raw_bids()` calls when importing raw data.
 
 ???+ example "Example"
     Enforce units for EDF files:
--- a/config.py
+++ b/config.py
@@ -428,6 +428,17 @@ itself, nor to the source analysis stage.
     ```
 """
 
+reader_extra_params: dict = {}
+"""
+Parameters to be passed to read_raw_bids calls when importing raw data.
+
+???+ example "Example"
+    Enforce units for EDF files:
+    ```python
+    reader_extra_params = {"units": "uV}
+    ```
+"""
+
 ###############################################################################
 # BREAK DETECTION
 # ---------------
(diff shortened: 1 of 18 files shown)


hoechenberger:
On the one hand, I agree we're currently being too strict.

On the other hand, we might produce false-positives here. Imagine a task `restless`, or `restore` or `restricted`, or `restaurant`, … you get the idea.

I'm starting to think we should introduce a config option, allowing the user to specify whether the data at hand is a resting-state recording or not.

WDYT?

agramfort:
yes I think you are right. How would you write such an option @hoechenberger ?

hoechenberger:
@agramfort 

Since we currently only allow to process a single task, we could simply have

```python
task_is_resting_state: bool = False
```

Super low-effort and not very flexible, but I think absolutely sufficient for now. WDYT?

hoechenberger:
> @agramfort
> 
> Since we currently only allow to process a single task, we could simply have
> 
> ```python
> task_is_resting_state: bool = False
> ```

Maybe better:

```python
task_is_resting_state: Optional[bool] = None
```
and if `None`, we apply the existing heuristic (task name equal to "rest"); and with a boolean value, the user can force the treatment of any task as either resting-state or not.

agramfort:
looks pretty ok https://output.circle-artifacts.com/output/job/27fa0c6f-01aa-4c51-8044-aaf015d4bdff/artifacts/0/site/examples/ds003775.html !

this one is ready for review @hoechenberger 🙏

agramfort:
@hoechenberger please have a look at this one when you have a moment. :pray:

config.py
```diff
@@ -428,6 +428,17 @@ class ArbitraryContrast(TypedDict):
     ```
 """
 
+reader_extra_params: dict = {}
+"""
+Parameters to be passed to read_raw_bids calls when importing raw data.
```
hoechenberger, line 433:
```suggestion
Parameters to be passed to `read_raw_bids()` calls when importing raw data.
```

config.py
```diff
@@ -428,6 +428,17 @@ class ArbitraryContrast(TypedDict):
...
+
+???+ example "Example"
+    Enforce units for EDF files:
+    ```python
+    reader_extra_params = {"units": "uV}
```
hoechenberger, line 438:
```suggestion
    reader_extra_params = {"units": "uV"}
```

config.py
```diff
@@ -860,8 +871,8 @@ class ArbitraryContrast(TypedDict):
 (value) to a more legible one (value).
 
 This is a **required** parameter in the configuration file, unless you are
-processing resting-state data. If left as `None` and [`task`][config.task]
-is not `'rest'`, we will raise an error.
+processing resting-state data. If left as `None` and
+[`task_is_rest`][config.task_is_rest] is not True, we will raise an error.
```
hoechenberger, line 875:
```suggestion
[`task_is_rest`][config.task_is_rest] is not `True`, we will raise an error.
```

config.py
```diff
@@ -899,6 +910,12 @@ class ArbitraryContrast(TypedDict):
     ```
 """
 
+task_is_rest: bool = False
+"""
+Set to true is the task is or should be processed as resting
+state data.
```
hoechenberger, line 916:
```suggestion
Whether the task should be treated as resting-state data.
```

config.py
```diff
@@ -3785,9 +3805,7 @@ def save_logs(logs):
 
 
 # Another check that depends on some of the functions defined above
-if (get_task() is not None and
-        get_task().lower() != 'rest' and
-        conditions is None and
+if (not task_is_rest and conditions is None and
         'MKDOCS' not in os.environ):
```
hoechenberger, line 3809:
```suggestion
if (
    not task_is_rest and
    conditions is None and
    'MKDOCS' not in os.environ
):
```

tests/configs/config_ds003775.py
```diff
@@ -0,0 +1,47 @@
...
+
+parallel_backend = 'loky'
+dask_open_dashboard = True
+
+on_error = 'continue'
+# on_error = 'abort'
+# on_error = 'debug'
+
+# log_level = 'debug'
+log_level = 'info'
```
hoechenberger, line 45:
```suggestion
log_level = 'info'
```

hoechenberger (approved):
Some nitpicks, otherwise LGTM!

agramfort:
🙏 @hoechenberger

hoechenberger:
Thanks @agramfort!!!

PR #587, opened 2022-08-16
Infant

larsoner:
### Before merging …

Two small tweaks:

1. Only emit the "freesurfer takes forever" message if we *don't* emit the "subject already found, skipping" message
2. Use `failsafe_run` for the `mkheadsurf` step. It helped me debug the failure and find out that my `sub-007` recon was interrupted so never produced the `T1.mgz`

For (2) at some point we might want to come up with some "did recon-all actually complete" check at some point. But for now this at least makes it easier to see it


hoechenberger:
Thanks @larsoner!

PR #588, opened 2022-08-17
BUG: Use patch information in computing normals

larsoner:
### Before merging …

`add_dist='patch'` doesn't add much time, but it improves the source space normal definitions by using the average normal in a surface patch (i.e., of all high-res verts for which the given source space vert is the closest source space vert)


agramfort (approved):

hoechenberger:
@larsoner This is a good addition, but could you either push a new commit or force-push the existing changes with a commit message that contains the phrase `[circle full]` so we run the entire CI pipeline?

I'm also wondering if this should be our default, and only if one **doesn't** want to run all tests, one should need to supply such an amended commit message. We're often forgetting to run all tests before merging.

Issue #590, opened 2022-08-18
DOC: Not all params documented


larsoner:
Looking at https://mne.tools/mne-bids-pipeline/settings/source/inverse.html I was surprised there was no `loose` param. But then:
```
$ git grep loose
config.py:# loose: Union[float, Literal['auto']] = 0.2
```
We should have some CI process whereby these are checked. I suggest:

1. Write a little `pytest` test for this
2. Put it in `tests/test_doc.py`
3. Add a GH actions run that runs `pytest tests`
4. Add `__test__ = False` to the top of `tests/run_test.py` (and whatever else we [need](https://stackoverflow.com/questions/50610557/how-to-mark-one-function-not-a-test-for-pytest)) so that `pytest` does not try to collect/run it

I can do this if people are on board

hoechenberger:
+1

agramfort:
+1

larsoner:
... wait a minute -- `loose` and `depth` are actually *not* parameters because they were commented out 1.5 years ago in https://github.com/mne-tools/mne-bids-pipeline/pull/220 . Was this intentional? I don't see discussion of it in the PR so it seems like it might have been testing cruft.

hoechenberger:
Could have been intentional, I remember thinking about getting rid of those  parameters as we never used them during any of our analyses. But if you need them now it's good to have them back, I suppose :)

PR #592, opened 2022-08-21
ENH: Reenable loose and depth

larsoner:
### Before merging …

Toward #590, let's see if the parameters are auto-documented. If so, then there is probably no need for #590. If not, then I'll add a test that fails, then fix it.

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -1,4 +1,7 @@
 name: Run the test suite
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.number }}-${{ github.event.ref }}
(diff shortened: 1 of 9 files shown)


larsoner:
- Adding to namespace built fine: df6718a
- Adding a test for the doc failed (on a whole bunch of stuff locally but I fixed all but `loose` and `depth`: ce3bde2 / https://github.com/mne-tools/mne-bids-pipeline/runs/7935862117?check_suite_focus=true
- Add depth and loose to docs: ca785e4

Should come back green, ready for review/merge from my end!

agramfort (approved):

agramfort:
Thx @larsoner

PR #595, opened 2022-09-04
ENH: Add ssp_meg option

larsoner:
### Before merging …

When using maxfiltered data, `meg='combined'` usually makes sense. Maybe we should have this default to `True` meaning "combined when using maxfilter and separate otherwise", but this should maybe be an upstream change first in MNE-Python. In the meantime, let's expose the option at least.

### DIFF
--- a/config.py
+++ b/config.py
@@ -1063,6 +1063,12 @@ Whether to calculate the EOG projection vectors based on the the averaged or
 on individual EOG epochs.
 """
 
+ssp_meg: str = 'separate'
+"""
+Whether to compute SSP vectors for MEG channels separately (`'separate'`)
+or jointly (`'combined'`).
+"""
+
 ssp_reject_ecg: Optional[
     Union[
         Dict[str, float],
--- a/config.py
+++ b/config.py
@@ -1066,7 +1066,10 @@ on individual EOG epochs.
 ssp_meg: str = 'separate'
 """
 Whether to compute SSP vectors for MEG channels separately (`'separate'`)
-or jointly (`'combined'`) for magnetometers and gradiomenters.
+or jointly (`'combined'`) for magnetometers and gradiomenters. When using
+MaxFilter, magnetometer and gradiometer signals are synthesized from multipole
+moments jointly and are no longer independent, so it can be useful to estimate
+projectors from all MEG sensors simultaneously.
 """
 
 ssp_reject_ecg: Optional[
--- a/config.py
+++ b/config.py
@@ -569,8 +569,8 @@ warning:
 
 mf_st_duration: Optional[float] = None
 """
-There are two kinds of maxfiltering: SSS (signal space separation) and tSSS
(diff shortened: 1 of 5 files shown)


config.py
```diff
@@ -1063,6 +1063,12 @@ class ArbitraryContrast(TypedDict):
 on individual EOG epochs.
 """
 
+ssp_meg: str = 'separate'
+"""
+Whether to compute SSP vectors for MEG channels separately (`'separate'`)
+or jointly (`'combined'`).
```
hoechenberger, line 1069:
```suggestion
or jointly (`'combined'`) for magnetometers and gradiomenters.
```

hoechenberger (commented):
I like this idea! Could you add a changelog entry please?

config.py
```diff
@@ -1063,6 +1063,12 @@ class ArbitraryContrast(TypedDict):
 on individual EOG epochs.
 """
 
+ssp_meg: str = 'separate'
```
hoechenberger, line 1066:
```suggestion
ssp_meg: Literal['separate', 'combined'] = 'separate'
```

config.py
```diff
@@ -1063,6 +1063,12 @@ class ArbitraryContrast(TypedDict):
...
+"""
+Whether to compute SSP vectors for MEG channels separately (`'separate'`)
+or jointly (`'combined'`) for magnetometers and gradiomenters.
```
agramfort, line 1069: can you add a note here for why one option is better than another? I have to admit I never looked into this. Should we say it's recommended to use "combined" when using maxfilter?

larsoner: Because mag and grad signals are jointly created/synthesized by MaxFilter (they are not independent afterward) it makes sense mathematically to estimate the projectors from them jointly

hoechenberger: If the case is so clear-cut, can we default to None here and set it to combine if the user requested Maxwell-filtering, and separate otherwise?

larsoner: This brings me back to my top comment:

> When using maxfiltered data, meg='combined' usually makes sense. Maybe we should have this default to True meaning "combined when using maxfilter and separate otherwise", but this should maybe be an upstream change first in MNE-Python. In the meantime, let's expose the option at least.

I think @agramfort was never totally convinced of the utility here, so I'm hesitant to automatically enable it for people

hoechenberger: > This brings me back to my top comment:

Oh, I missed that one, sorry.

agramfort:
ok so should be make it "auto" and default to "combined" if maxfilter is on?

larsoner:
Ideally I think we would do this in MNE-Python first, but I guess we can add it here instead

larsoner:
(the backward compat break there would be a pain)

hoechenberger (approved):

hoechenberger:
Thanks @larsoner! CI should pass.

agramfort (approved):

PR #600, opened 2022-09-09
BUG: Fix URL

### LINKED ISSUE #539: add ds004107 to datasets
we should add the dataset to our examples:

https://openneuro.org/datasets/ds004107/

### LINKED ISSUE #599: BUG: Broken testing dataset
Locally I get:
```
$ python tests/download_test_data.py -d ds001971

----------------------

👋 Hello! This is openneuro-py 2022.1.0. Great to see you! 🤗

   👉 Please report problems 🤯 and bugs 🪲 at
      https://github.com/hoechenberger/openneuro-py/issues

🌍 Preparing to download ds001971 …
  File "/Users/larsoner/python/mne-bids-pipeline/tests/download_test_data.py", line 155, in <module>
    main(dataset)
  File "/Users/larsoner/python/mne-bids-pipeline/tests/download_test_data.py", line 144, in main
    _download(ds_name=ds_name, ds_path=ds_path)
  File "/Users/larsoner/python/mne-bids-pipeline/tests/download_test_data.py", line 125, in _download
    download_func(ds_name=ds_name, ds_path=ds_path)
  File "/Users/larsoner/python/mne-bids-pipeline/tests/download_test_data.py", line 34, in _download_via_openneuro
    openneuro.download(
  File "/Users/larsoner/opt/miniconda3/envs/mne/lib/python3.10/site-packages/openneuro/download.py", line 656, in download
    raise RuntimeError(f'Could not find path '
RuntimeError: Could not find path sub-001/eeg/sub-001_task-AudioCueWalkingStudy_run-01_events.tsvsub-001/eeg/sub-001_task-AudioCueWalkingStudy_run-01_eeg.set in the dataset. Please check your includes.
```
@hoechenberger how do you debug and fix these things?

larsoner:
### Before merging …

- [x] Add config for 4107
- [x] Add test that all tests show up in CircleCI config
- [x] Add 4107 and whatever else we need to tests

Okay the basic fix was easy. ~~Draft because I'm not sure how this download has ever worked, as this error appears in the `blame` from 2 years ago when it was added. It appears not to be used on CircleCI -- should it be? If so, I'll add another little sanity check test here that any dataset in `datasets.py` has at least some matching code in `.circleci/config.yml`.~~

Closes #599
Closes #539


larsoner:
I'll probably take care of https://github.com/mne-tools/mne-bids-pipeline/issues/539 here as well since it should be easy enough hopefully

agramfort (approved):

tests/run_tests.py
```diff
@@ -27,170 +27,99 @@ def fetch(dataset=None):
...
+# }
+#
+TEST_SUITE: Dict[str, Dict] = {
+    'ds003392': {},
+    'ds004229': {},
+    'ds004107': {},
+    'ds000117': {},
+    'ds003775': {},
+    'eeg_matchingpennies': {
+        'dataset': 'eeg_matchingpennies',
```
larsoner, line 57: I saw that a `TypedDict` was used in `datasets.py`, and this is nice syntatic sugar, but I think it hurts maintainability in the end here. Things can be simplified *a lot* and become more readable with defaults if we abandon it (at least until [PEP655 lands in 3.11 and we require it so we can use NotRequired](https://peps.python.org/pep-0655/)). By making the entries optional and having reasonable defaults for them:

1. Code is more DRY, less chance for error
2. Easier to add new datasets (see 4107 in the list above is a single easy line and everything else "just works")
3. It's immediately obvious where each dataset deviates from the default processing scheme enumerated above

So I've implemented this here.

hoechenberger: I disagree with these changes, as they reduce flexibility should we need it, and make it **less** clear what's going on. I specifically designed this here API to be very explicit and, yes, maybe a bit verbose, but with the advantage that no magic happens e.g. in terms of how the configuration files are located in the filesystem etc. The `dataset`, `config`, and `steps` keys should not be optional. Also, without a `TypedDict`, my IDE doesn't provide me any help as to which keys are valid. So this actually makes adding new test cases **harder**.

Interim reconsilation proposition:
Keep it a `TypedDict` but make **all** keys optional. So you can have your cake, and I can eat mine too:

```python
class TestOptionsT(TypedDict, total=False):
    ....
```

drammock: > they reduce flexibility should we need it

Can you explain how?  I'm a bit of a TypedDict n00b so it's not obvious to me how @larsoner's proposal ties your hands here.

> and make it **less** clear what's going on.

This I think is a matter of perspective. Some would say "OK, I see what the defaults are, I can easily scan the `TEST_SUITE` entries to see which ones deviate from the defaults. Clear."

> The `dataset`, `config`, and `steps` keys should not be optional.

If they have a default value, they're not really optional though? Maybe this is just semantics (optional for user to type out, vs optional to exist in the dict at all), but I see value in *not* forcing people to type out `ds003392` multiple times (in key, dataset, and config) if it's possible to derive it from the key.

> without a `TypedDict`, my IDE doesn't provide me any help

Fair point.

larsoner: > as they reduce flexibility should we need it

I don't see how this is the case by providing sensible defaults?

> I specifically designed this here API to be very explicit and, yes, maybe a bit verbose, but with the advantage that no magic happens e.g. in terms of how the configuration files are located in the filesystem etc. ... The dataset, config, and steps keys should not be optional

To generalize your idea here, I think you could (nearly) equivalently argue in the spirit of explicitness/verboseness that functions should also have no default arguments, and users should always have to pass them. I don't think it's really sustainable or best practice. If something is the case often enough (maybe 50%+ of the time?), you should give it a it default and use it.

Coming back to this case, take `'config'` -- it is the same for every dataset except some ERP_CORE variants. It seems like a waste to specify it every time except for those cases. In this PR, it's immediately clear (and more clear on than on `main`) that the same config is reused in each of these cases, just with a new `task`.

And for the changes here, I'm able to just add this line:
```
    'ds004107`: {},
```
and everything "just works", and it's clear why with the comment above the dict (just like a function signature). In `main` I have to add a bunch of redundant stuff.

In the end, though, it's about balancing DRY vs explicitness vs risk of error (in different ways!), so I think we'll have to agree to disagree on how we should strike the balance here.

Maybe @agramfort can vote on whether 1) these three keys need to be obligatory, or 2) if they should all be optional to the extent possible, and then we can move on?

> Keep it a TypedDict but make all keys optional. So you can have your cake, and I can eat mine too:

I've pushed a commit for this part at least so far.

agramfort:
Personally I prefer this dryer option. +1 merge after the CIs are green.  Note that the new dataset shall be added to the documentation page too.

larsoner:
I think there is a bug with 1971 with `participants.tsv`:
```

participant_id  participant_id  sex     age     handedness
sub-001 CQ2     F       18-38   R
sub-002 BT4     F       28      R
sub-003 CQ1     F       18-38   R
sub-004 CQ3     F       18-38   R
sub-005 CP9     M       18-38   R
sub-006 CA7     M       18-38   R
sub-007 CB6     M       32      R
sub-008 BV2     M       38      R
sub-009 CP8     F       18-38   R
sub-010 CP1     F       23      R
sub-011 CP2     M       18-38   R
sub-012 CO9     F       18-38   R
sub-013 CP5     M       18-38   R
sub-014 CL1     M       18-38   R
sub-015 BL7     M       28      R
sub-016 CO3     M       18-38   R
sub-017 CE2     M       18-38   R
sub-018 CV6     F       18-38   R
sub-019 CJ2     F       18-38   R
sub-020 CY8     M       18-38   R
```
It has two `participant` columns. I'll need to exclude it for now I think.

tests/run_tests.py
hoechenberger: Ok, let's move on here … `total=False` makes me sufficiently happy because I still get help from my IDE when I wish to add a new dataset; and you can omit everything you don't need, unless I'm overlooking something here. Thanks everybody for sharing your thoughts!

hoechenberger (approved):

PR #603, opened 2022-09-12
ENH: Allow decim of GAT

### LINKED ISSUE #602: MAINT: Speed up ERP CORE runs
They [take a while](https://app.circleci.com/pipelines/github/mne-tools/mne-bids-pipeline/3142/workflows/de070ac0-5b7d-4d15-b9fd-e0586990af2f):

Screen Shot 2022-09-10 at 3 51 02 PM

And about 15 minutes in one run is spent on time-by-time decoding:

Screen Shot 2022-09-10 at 3 57 29 PM

@hoechenberger have you optimized these for run time already? If not, I might try cropping runs, decimating epochs, etc. Also for decoding across time:

https://mne.tools/mne-bids-pipeline/settings/sensor/statistics.html#config.decoding_time_generalization

We could perhaps add a `decoding_time_generalization_decim` which would be an *additional* decimation factor (on top of the `decim` used when creating epochs) to apply. In general the across-time stuff is fairly smooth:

https://mne.tools/mne-bids-pipeline/examples/ERP_CORE/sub-015_ses-N400_task-N400_report.html

So decimating by a factor of 2 or 4 would get us a squared-factor (i.e., 4x or 16x) speedup.

larsoner:
### Before merging …

Closes #602

### DIFF
--- a/scripts/report/_01_make_reports.py
+++ b/scripts/report/_01_make_reports.py
@@ -242,10 +242,7 @@ def _plot_time_by_time_decoding_scores(
     ax.plot(times, mean_scores, ls='-', lw=2, color='black',
             label='mean')
 
-    extra = ''
-    if decim > 1:
-        extra = f'\n(decim={decim})'
-    ax.set_xlabel(f'Time (s){extra}')
+    _label_time_by_time(ax, xlabel='Time (s)', decim=decim)
     if metric == 'roc_auc':
         metric = 'ROC AUC'
     ax.set_ylabel(f'Score ({metric})')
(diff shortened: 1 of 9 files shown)


scripts/report/_01_make_reports.py
```diff
@@ -232,7 +233,7 @@ def _plot_time_by_time_decoding_scores(
         max_scores = np.diag(max_scores)
         min_scores = np.diag(min_scores)
 
-    fig, ax = plt.subplots()
+    fig, ax = plt.subplots(constrained_layout=True)
```
larsoner, line 236: I want to change all of these (see https://github.com/mne-tools/mne-bids-pipeline/issues/602#issuecomment-1243836602). I can do it in a separate PR if it would help, but it seemed easy enough to do it here

hoechenberger:
Thanks @larsoner!

PR #605, opened 2022-09-12
ENH: Add --debug to run_tests.py

### LINKED ISSUE #578: MAINT: `run_tests.py` should have `--debug` option
It would be nice to do `python tests/run_tests.py ds000248 --debug` and have it

1. Automatically set `n_jobs=1`
2. Set `on_error='debug'`

I went to do this but it quickly became complicated because `failsafe_run` requires `on_error` as an input, and every script imports `on_error` from config... so we'd have to add a `get_on_error` like we have for `get_n_jobs` or something. Not entirely sure...

One question is, why is `on_error` a parameter for `failsafe_run` at all? It seems like it should just use the `on_error` already in the `config.py` namespace where it's defined.

larsoner:
### Before merging …

- [x] Factor `on_error` out of the decorator
- [x] Check that all command-line arguments are consumed in `run.py` instead of dangerously/silently allowing them (which `fire.Fire` does) by using `**kwargs` and making sure it's empty
- [x] Change default for "download" to be zero/off
- [x] Add support for `--download` to mean "yes download"
- [x] Add support for `--debug` to mean "run in debug mode" (along with `--debug=[0|1]` for consistency)

Closes #578

### DIFF
--- a/config.py
+++ b/config.py
@@ -2601,6 +2601,14 @@ def get_n_jobs() -> int:
     return n_jobs
 
 
+def get_on_error() -> int:
+    if interactive:
+        this_on_error = 'debug'
+    else:
+        this_on_error = os.getenv('MNE_BIDS_STUDY_ON_ERROR', on_error)
+    return this_on_error
+
+
 dask_client = None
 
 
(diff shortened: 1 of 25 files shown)


larsoner:
To test, just run something like `python tests/run_tests.py ds000246 --debug` after adding `1/0` in some step

agramfort (approved):

agramfort:
Thx 🙏 @larsoner

PR #606, opened 2022-09-19
ENH: Add M/EEG processing

### LINKED ISSUE #601: BUG: MEG+EEG?
I just tried doing:
```
ch_types = ['meg', 'eeg']
```
and got
```
ValueError: EEG data can only be analyzed separately from other channel types. Please adjust `ch_types` in your configuration.
```
Is it not possible to analyze M/EEG data jointly?

larsoner:
### Before merging …

- [x] Add support for MEG+EEG data
- [x] Reenable SSP for EEG
- [x] ~~Add~~ Check that conditional that empty-room cov cannot be used for MEG+EEG data *EDIT: Already exists*
- [x] CIs green
- [x] Reports checked for proper sensor type plots (MEG and EEG) for each step (SSP, sensor plots/evoked, inverse)

@hoechenberger @agramfort sound like a plan? Also, why was SSP for EEG not supported / explicitly an error before? Looks like the error conditional that I removed here was added in https://github.com/mne-tools/mne-bids-pipeline/pull/92 or so.

Closes #601


larsoner:
Not quite ready yet, some plots missing from [the report](https://output.circle-artifacts.com/output/job/aeddfcc1-a417-47fc-b228-5bcf2df15dc4/artifacts/0/site/examples/ds000248.html)

larsoner:
Ready for review/merge from my end.

Actually what I thought was missing was actually correct -- the empty-room plots *should* only have MEG channels. All other plots seem to reasonably contain MEG+EEG.

Also fixes a bug where STIM channels were included in decoding (whoops!). Now when we `if cfg.analyze_channels; epocs.pick(cfg.analyze_channels)` in a few places, we `else: epochs.pick(cfg.ch_types)` to restrict to the relevant data channels.

https://output.circle-artifacts.com/output/job/11d22b26-cfd0-4c25-a1f0-ea9197d5c3c0/artifacts/0/site/examples/ds000248/sub-01_task-audiovisual_report.html

larsoner:
Okay there was still a bug with picking. I changed the default `cfg.analyze_channels` from `'all'` which would include STIM channels to `'ch_types'`, which means "use cfg.ch_types".

I also fixed a doc bug where the "used in this example" was not determined properly because things like `config['decode']` is only half the check, we also need `config['conditions']` to be specified in order for decoding to be done.

agramfort (approved):
🎉 🙏

PR #607, opened 2022-09-20
ENH: Add more step caches

larsoner:
### Cleanups

- [x] Clean up startup logging
- [x] Fix bug where only the first run was used for SSP
- [x] Fix bug where `memory_location=None` was not working properly (by always using `ConditionalStepMemory`)
- [x] Add `--no-cache` to `run_tests.py` allow running without any cache
- [x] Unify log formatting (use coloredlogs default, probably good enough)

### Caching implemented

- [x] Epochs
- [x] Run ICA
- [x] Run SSP
- [x] Apply SSP
- [x] Apply ICA
- [x] PTP reject
- [x] Make evoked
- [x] Decoding full epochs
- [x] Decoding time_by_time

### Next follow-up PR (not here):

- Add `.pop` to read steps and `assert len(in_files) == 0` at the end of each step (as in the `make_evoked` step)
- Time frequency
- Report (maybe?)
-  Add re-run test checks on CircleCI... somehow?

### DIFF
--- a/config.py
+++ b/config.py
@@ -1684,7 +1684,7 @@ which is typically the pre-stimulus period.
 
 If ``'emptyroom'``, the noise covariance matrix will be estimated from an
 empty-room MEG recording. The empty-room recording will be automatically
-selected based on recording date and time.
+selected based on recording date and time. This cannot be used with EEG data.
 
 If ``''rest'``, the noise covariance will be estimated from a resting-state
 recording (i.e., a recording with `task-rest` and without a `run` in the
--- a/config.py
+++ b/config.py
@@ -2015,7 +2015,7 @@ def gen_log_kwargs(
     step_name = f'{script_path.parent.name}/{script_path.stem}'
 
     extra = {
-        'step': f'📊 {step_name}'
+        'step': f'⏳️ {step_name}'
     }
     if subject:
         extra['subject'] = subject
--- a/config.py
+++ b/config.py
@@ -2051,10 +2051,7 @@ _epochs_split_size = '2GB'
 if "MNE_BIDS_STUDY_CONFIG" in os.environ:
     cfg_path = pathlib.Path(os.environ['MNE_BIDS_STUDY_CONFIG'])
 
-    if cfg_path.exists():
-        msg = f'Using custom configuration: {cfg_path}'
-        logger.info(**gen_log_kwargs(message=msg))
-    else:
+    if not cfg_path.exists():
         msg = ('The custom configuration file specified in the '
                'MNE_BIDS_STUDY_CONFIG environment variable could not be '
                'found: {cfg_path}'.format(cfg_path=cfg_path))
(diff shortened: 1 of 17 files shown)


config.py
```diff
@@ -2012,7 +2015,7 @@ def gen_log_kwargs(
     step_name = f'{script_path.parent.name}/{script_path.stem}'
 
     extra = {
-        'step': step_name
+        'step': f'⏳️ {step_name}'
```
agramfort, line 2018: fancy 😉

run.py
```diff
@@ -251,11 +261,14 @@ def process(
         script_modules = [*SCRIPT_MODULES['init'], *script_modules]
 
     logger.info(
-        "👋 Welcome aboard the MNE BIDS Pipeline!\n"
+        "👋 Welcome aboard the MNE BIDS Pipeline!"
+    )
+    logger.info(
+        f"🧾 Using configuration: {config}"
```
agramfort, line 267: @hoechenberger would be so proud of you 😆

run.py
```diff
@@ -251,11 +261,14 @@ def process(
...
     )
 
     for script_module in script_modules:
-        logger.info(f'🚀 Now running script: {script_module.__name__} 👇')
+        logger.info(f'🚀 Now running script:  {script_module.__name__} 👇')
```
agramfort, line 271:
```suggestion
        logger.info(f'🚀 Now running script: {script_module.__name__} 👇')
```

scripts/sensor/_02_decoding_full_epochs.py
```diff
@@ -36,12 +36,15 @@
...
-def run_epochs_decoding(*, cfg, subject, condition1, condition2, session=None):
-    msg = f'Contrasting conditions: {condition1} – {condition2}'
-    logger.info(**gen_log_kwargs(message=msg, subject=subject,
-                                 session=session))
-
+def get_input_fnames_epochs_decoding(**kwargs):
+    cfg = kwargs.pop('cfg')
+    subject = kwargs.pop('subject')
+    session = kwargs.pop('session')
+    # TODO: Somehow remove these?
```
agramfort, line 43: what do you mean?

scripts/sensor/_03_decoding_time_by_time.py
```diff
@@ -52,12 +52,15 @@ def fit(self, *args, **kwargs):
...
-    logger.info(**gen_log_kwargs(message=msg, subject=subject,
-                                 session=session))
-
+def get_input_fnames_time_decoding(**kwargs):
+    cfg = kwargs.pop('cfg')
+    subject = kwargs.pop('subject')
+    session = kwargs.pop('session')
+    # TODO: Somehow remove these?
+    del kwargs['condition1']
+    del kwargs['condition2']
```
agramfort, line 61: idem

scripts/sensor/_02_decoding_full_epochs.py
larsoner: These `kwargs` are not used in `get_input_fnames_*`, so ideally they wouldn't be passed to it at all.

It suggests to me that there is some refactoring to be done at some point. For example maybe we should enforce that `get_input_fnames_*` should only ever take `cfg, subject, session, run` kwargs (at most) until proven otherwise. But these are considerations for another PR, hence the comment to think about it later.

agramfort:
thx @larsoner !

PR #612, opened 2022-09-23
WIP: Add table with all examples

larsoner:
### Before merging …

I have wanted a table showing the capabilities of all examples in one place. This is at least a proof of concept that it can be done. @hoechenberger any interest in taking over to make it more readable? If a sticky first column (via CSS) is good enough I could do that, but thought you might have more ideas.

### DIFF
--- a/docs/source/examples/gen_examples.py
+++ b/docs/source/examples/gen_examples.py
@@ -1,3 +1,5 @@
+#!/bin/env python
+
 import os
 import shutil
 import sys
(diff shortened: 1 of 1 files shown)


larsoner:
https://output.circle-artifacts.com/output/job/9165ac24-09bb-428c-b67a-18dbd23f1154/artifacts/0/site/examples/examples.html

agramfort:
I like that !

larsoner:
Maybe good enough to start? If so let's merge and we can iterate with better consistency and readability in follow-up PRs when people are motivated

agramfort:
I would say yes

PR #614, opened 2022-09-28
ENH: Add SSP joint plots to reports

larsoner:
### Before merging …

Locally at least it produces something reasonable:

Screenshot from 2022-09-28 15-59-55

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -259,7 +259,7 @@ jobs:
       - run:
           <<: *bashenv
       - run:
-          name: Get ds003392
+          name: Get ds004229
           command: |
             python ~/project/tests/download_test_data.py -d ds004229
       - save_cache:
@@ -320,7 +320,7 @@ jobs:
       - run:
           name: test ds000117
           command: |
-            export DS=ds000117
+            DS=ds000117
             python tests/run_tests.py --download=0 ${DS}
             mkdir -p ~/reports/${DS}
             cp -av ~/mne_data/derivatives/mne-bids-pipeline/${DS}/*/**/*.html ~/reports/${DS}/
@@ -346,7 +346,7 @@ jobs:
       - run:
           name: test ds003775
           command: |
-            export DS=ds003775
+            DS=ds003775
             python tests/run_tests.py --download=0 ${DS}
             mkdir -p ~/reports/${DS}
             cp -av ~/mne_data/derivatives/mne-bids-pipeline/${DS}/*/**/*.html ~/reports/${DS}/
@@ -372,7 +372,7 @@ jobs:
       - run:
           name: test ds001971
           command: |
-            export DS=ds001971
+            DS=ds001971
             python tests/run_tests.py --download=0 ${DS}
             mkdir -p ~/reports/${DS}
             cp -av ~/mne_data/derivatives/mne-bids-pipeline/${DS}/*/**/*.html ~/reports/${DS}/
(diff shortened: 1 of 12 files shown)


hoechenberger:
Nice! Can we change the order such that the topoplot comes first?

larsoner:
You mean horizontally? I don't think we should here. The order comes from the upstream joint plot function where we chose the order presumably with consensus and review already

larsoner:
@hoechenberger how do we resolve:

    RuntimeError: Error 403 when trying to download
    /home/circleci/mne_data/ds004229/sub-102/meg/sub-102_task-amnoise_meg.fif from
    https://s3.amazonaws.com/openneuro.org/ds004229/sub-102/meg/sub-102_task-amnoise_meg.fif?versionId=Mfnz6drCpmRBzW1NlZA7C4o5.ZCQjWae

.circleci/config.yml
```diff
@@ -942,47 +949,6 @@ jobs:
...
-            - data-cache-ds003392-{{ .Revision }}
-      - restore_cache:
-          keys:
-            - data-cache-ds004229-{{ .Revision }}
-      - restore_cache:
-          keys:
-            - data-cache-eeg_matchingpennies-{{ .Revision }}
-      - restore_cache:
-          keys:
-            - data-cache-ERP_CORE-{{ .Revision }}
```
larsoner, line 985: I expect these to break things -- I plan to fix them here

hoechenberger:
> You mean horizontally? I don't think we should here. The order comes from the upstream joint plot function where we chose the order presumably with consensus and review already

Oh i didn't realize this came from upstream, I actually didn't look at your code before writing that comment! Let's keep it the way it is then.

hoechenberger:
> @hoechenberger how do we resolve:
> 
> 
> 
>     RuntimeError: Error 403 when trying to download
> 
>     /home/circleci/mne_data/ds004229/sub-102/meg/sub-102_task-amnoise_meg.fif from
> 
>     https://s3.amazonaws.com/openneuro.org/ds004229/sub-102/meg/sub-102_task-amnoise_meg.fif?versionId=Mfnz6drCpmRBzW1NlZA7C4o5.ZCQjWae
> 
> 
> 
> 


If it persists we need to open an issue with OpenNeuro. Can you ping me if the problem doesn't go away in its own? I can then take over.

larsoner:
> Can you ping me if the problem doesn't go away in its own? I can then take over.

It has been like this through 4 or 5 tries over about 24 hours. It's that long enough?

hoechenberger:
> > Can you ping me if the problem doesn't go away in its own? I can then take over.
> 
> 
> 
> It has been like this through 4 or 5 tries over about 24 hours. It's that long enough?

Ouch. Yes. I will try to take a look.

hoechenberger:
@larsoner I've opened a bug report: https://github.com/OpenNeuroOrg/openneuro/issues/2677

.circleci/config.yml
agramfort: Did you actually break something?

agramfort (approved):
If nothing breaks. Lets merge ! 🙏 @larsoner

hoechenberger:
I'd also say it's okay to move ahead here and merge if @larsoner is okay with this. It might take another day or two for the OpenNeuro issue to get resolved…

larsoner:
We can wait. I want to fix the thing I likely broke on CircleCI here, too

larsoner:
For now I'm just going to disable ds004229. That way we can get green CIs on these PRs and merge.

PR #615, opened 2022-09-29
ENH: Cache coreg, bem, cov

larsoner:
### Before merging …

- [x] Add a magic `in_files['__force_run__'] = True` to force rerunning even if the call is in the cache
- [x] Cache `_02_coreg_surfaces.py`
- [x] Remove redundant `-head.fif` creation from `_01_bem_surfaces.py` in favor of `_02_coreg_surfaces.py`
- [x] Cache `_01_bem_surfaces.py`
- [x] Add a magic `in_files['__unknown_inputs__'] = True` that gets detected to force running without cache (e.g., for cov-as-callable)
- [x] Cache `_03_make_cov.py`
- [x] For all steps above -- but `_02_coreg_surfaces.py` and `_01_bem_surfaces.py` in particular -- skipping can only occur if the function has been called before. If people ship FS subjects with BEM and/or head surfaces, they will be recomputed (because they won't exist in the cache until *mne-bids-pipeline* computes them). We should change this to (re)enable skipping if the files already exist.

Not super thoroughly tested, but the ideas are there at least.

Questions

### DIFF
--- a/config.py
+++ b/config.py
@@ -2954,21 +2954,16 @@ class ConditionalStepMemory():
             in_files = None
             if self.get_input_fnames is not None:
                 in_files = kwargs['in_files'] = self.get_input_fnames(**kwargs)
-            assert isinstance(in_files, (dict, type(None))), type(in_files)
-
-            run_without_memory = False
             if self.memory is None:
-                run_without_memory = True
-
-            # Deal with cases (e.g., custom cov) where input files are unknown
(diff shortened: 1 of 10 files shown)


hoechenberger:
@larsoner This looks good from a distance; I don't have the time to carefully examine all details here, but I'd say once you're happy and nothing breaks, let's merge this!

PR #622, opened 2022-10-06
MAINT: Move to standard flat folder structure

### LINKED ISSUE #620: MAINT: Refactor project code to mne_bids_pipeline/*.py
I think *a lot* of the internal mechanics might be simplified if we can make everything private by moving `run.py` to `mne_bids_pipeline/_run.py` etc. I've seen quite a bit of redundant code and gymnastics in all the automation routines that could be simplified by this. Just a few examples:

1. No need for `sys.path` insertion because we can use proper package-relative imports like `from ._scripts import ...` rather than absolute `from scripts import ...` which failed before (for me at least)
2. Code dedup between `config.py` and `run.py`, e.g., related to logging and imports
3. Moving all helper functions out of `config.py` to a new `_utils.py` to make the `config.py` "pure" in the sense that it can contain just all the default config values -- so we can both use it ourselves, *and* copy-paste it to give people a default config (probably while commenting out all lines like in standard Unix config files)
4. Simplification of doc building gymnastics due to interactions between variables (e.g., https://github.com/mne-tools/mne-bids-pipeline/pull/619 would be easier I think)

If we support `python run.py ...` for people rather than making them transition to `mne_bids_pipeline ...` we have to wait on this stuff :(

_Originally posted by @larsoner in https://github.com/mne-tools/mne-bids-pipeline/issues/611#issuecomment-1265976118_

### LINKED ISSUE #624: "decim" information in axis label isn't always rendered correctly
I think this was introduced in #603.

- Sometimes the information is just cut off:
  Screen Shot 2022-10-07 at 11 09 05
- Sometimes it's displayed as an array (?):
  Screen Shot 2022-10-07 at 11 09 14

(Examples are from https://output.circle-artifacts.com/output/job/e043e4f2-7911-4df9-bd5c-08e0b37933f9/artifacts/0/site/examples/ERP_CORE/sub-average_ses-P3_task-P3_report.html)

Generally, I think the decim information should be placed in the same line as the rest of the axis label.

cc @larsoner

larsoner:
Closes #620
Closes #624

I couldn't get an editable install working with the `setuptools` backend because apparently this requires the *system* setuptools to be v62 or later:

https://stackoverflow.com/a/69711730/2175965

Even on Ubuntu 22.04 this isn't the case. So let's switch to `poetry`. Locally I get:
```
$ pip install -ve .[tests]
...
Successfully installed mne-bids-pipeline-0.0.0.post622.dev0+262e571
$ cd ~
$ python
>>> import mne_bids_pipeline
>>> mne_bids_pipeline.__version__
'0.0.0.post622.dev0+262e571'
```
So everything is about what I'd expect.

This changes CircleCI


hoechenberger:
@larsoner I'm not a fan of moving to poetry; have you tried disabling build isolation via

```shell
pip install --no-build-isolation -e .
```
?

larsoner:
> I'm not a fan of moving to poetry; have you tried disabling build isolation via

No I hadn't tried that -- it seems to have worked!

hoechenberger:
> > I'm not a fan of moving to poetry; have you tried disabling build isolation via
> 
> No I hadn't tried that -- it seems to have worked!

Nice!!! :)

agramfort (approved):

larsoner:
https://output.circle-artifacts.com/output/job/3bf22735-0c13-4d13-918c-9a84d55473c9/artifacts/0/site/settings/general.html

mne_bids_pipeline/tests/configs/config_ERP_CORE.py
```diff
@@ -33,13 +33,12 @@
 task = os.environ.get('MNE_BIDS_STUDY_TASK')
 sessions = [task]
 
-subjects = ['015', '016', '017', '018', '019']
-
+subjects = ['015', '016', '017', '018']
```
larsoner, line 36: @hoechenberger okay to reduce this to 4? 4 or 6 make more sense than 5 because `N_JOBS=2`, 4 will be faster than 5 but 5 should be about the same amount of execution time as 6.

hoechenberger: I think I picked 5 because some effects are only visible when averaging across a certain minimum number of participants. We can try 4, but need to look at the grand average reports to ensure that the results still look somewhat "convincing".

pyproject.toml
```diff
@@ -29,7 +29,7 @@ dependencies = [
     "coloredlogs",
     "python-picard",
     "fire",
-    "pyqt5",
+    "qtpy",
```
hoechenberger, line 32: Nice, this makes it easier to work with in a conda-based environment too

mne_bids_pipeline/tests/configs/config_ERP_CORE.py
hoechenberger: @larsoner I'm looking into this right now and adjusting parameters. will push to this branch once i have results

larsoner: @hoechenberger it didn't actually affect the run time all that much. So we can also just revert if you want

larsoner: Yes please do -- it's okay on my Linux machine but pretty bad on my M1

hoechenberger: Any idea why that might be? I also have the impression it used to be faster on my Intel Mac

larsoner: For me it slams all cores even if I set `OPENBLAS_NUM_THREADS=2` for example. I think it might thus be due in part to a parallelization oversubscription, as we run `N_JOBS=2` and I think both try to use all cores... but I haven't fully checked.

larsoner:
@hoechenberger okay for me to revert back to 5 subjects, push, and mark for merge-when-green? I think we can improve the decim/n_jobs in a follow-up PR. In the meantime this one has some nice features and speedups (~32 minutes to run tests down to ~25).

larsoner:
@hoechenberger I just pushed a `ica_decim = 2` to at least help a little bit. Feel free to merge as is or push commits to make it work better. Or if you'd like to investigate further I can revert the `ica_decim` here and we can do it in another PR...

mne_bids_pipeline/tests/configs/config_ERP_CORE.py
hoechenberger: I found this weird behavior, but I suspect it's a bug in Python itself?

```shell
❯ python -c 'import platform; print(platform.processor())'
i386
❯ arch -arm64 python -c 'import platform; print(platform.processor())'
arm
❯ lipo -archs `which python`
arm64
```
So the binary is clearly ARM only, but when run, Python reports it's running on x86 unless one uses the `arch§ command, which to my knowledge should only be of use with fat binaries…

hoechenberger: I just ran with `n_jobs` set to 1 and this actually finished faster than with `2`, yet still utilized all CPU cores. Definitely resource oversubscription.

We're running the ERP CORE tests through Dask, if I'm not mistaken. So the problem might lie there…

larsoner: > I found this weird behavior, but I suspect it's a bug in Python itself?

I do not get that on my ARM/M1 machine with conda-forge Python:
```
$ python -c 'import platform; print(platform.processor())'
arm
```
so I suspect something is wrong with your `which python` / nativeness of the CPython you're using.

hoechenberger: Extremely weird…

PR #623, opened 2022-10-06
fix forward computation when MEG only is used but data contains EEG channels

agramfort:
@larsoner this bug was introduced with the forward code refactoring.


scripts/source/_04_make_forward.py
```diff
@@ -119,6 +119,15 @@ def run_forward(*, cfg, subject, session, in_files):
 
     # Info
     info = mne.io.read_info(in_files.pop('info'))
+    info = mne.pick_info(
+        info,
+        mne.pick_types(
+            info,
+            meg="meg" in cfg.ch_types,
```
larsoner, line 126: IIRC it has to be more than this I think since `ch_types` can also be `'mag'` or `'grad'` or lists of these

agramfort:
thx @larsoner 🙏

PR #625, opened 2022-10-08
ENH: Add CSP

hoechenberger:
Migrating changes from #479 one by one.

Closes #479
Closes #414

Todo:

- [x] Group average
- [x] Report group average

- [x] Report single-subject

### Before merging …


hoechenberger:
@agramfort See the "Todo" checklist in the PR description for what still needs to be done

I think an important point that's not listed there is, we need a good testing dataset… the one that's used in the MNE examples does not exist in BIDS AFAIK. So if you have any idea which dataset we could use for testing, that would be helpful

larsoner:
Naively I would try `eeg_matchingpennies` but if that doesn't work then it seems like even something like `sample` should work for auditory vs visual

larsoner:
... if you do go with `sample` maybe add it to the `ds000248_ica` run, since the main run is one of the longest-duration tests we have

https://app.circleci.com/pipelines/github/mne-tools/mne-bids-pipeline/3376/workflows/c5092e69-5fe1-4499-85b0-ef9b35c267c4

but `eeg_matchingpennies` is even shorter currently so if it works, it's a better choice testing-wise.

larsoner:
Have you tried it on any of the datasets that have multiple participants? I would just add it to the configs, run them, and see if any of them work out. It's probably hours or days faster than waiting for @agramfort to un-bury himself in emails to respond :)

larsoner:
For now you could push with that change so that this can at least be tested and merged, and we can wait on a better dataset until later. Or maybe it's possible to get some ERP_CORE component working with CSP with the right type of filtering, if you push then @agramfort or I could try

larsoner:
I pushed a commit to add it to ERP_CORE_ERN which has multiple subjects. The ROC_AUC is maybe good enough? Worth checking...

I also made the message more compact so it can actually be readable on a single line in a terminal, and uniformly formatted so that all lines line up nicely:
```
[2022-10-17 14:32:05] │ ⏳️ sensor/_05_decoding_csp sub-015 ses-ERN Contrast: response/incorrect – response/correct,  4.0– 7.0 Hz (theta)-0.400–-0.200 sec:  roc_auc=0.692
```
(the `Time` and `Freqs` that were in there are redundant with the units and knowledge of the params that must be specified to actually run CSP, so it seems a shame to use terminal output chars for them)

agramfort:
@hoechenberger @larsoner can you push with no ci skip so we can see the rendered reports?

hoechenberger:
There are no reports yet
I see todo in the top comment 
I need to migrate that code from
The previous PR

larsoner:
@hoechenberger do you want me to try copy-pasting the group averaging and time-freq plotting here? I could do it today or tomorrow

larsoner:
... from #479, I mean

hoechenberger:
Yes please feel free to go ahead, I won't have time before the weekend

larsoner:
Ready for review/merge from my end:

- [Fixed features table](https://output.circle-artifacts.com/output/job/6f045867-f435-4c17-ac3f-bf3b4c36f5f1/artifacts/0/site/examples/examples.html#demonstrated-features)
- [Single subject report](https://output.circle-artifacts.com/output/job/6f045867-f435-4c17-ac3f-bf3b4c36f5f1/artifacts/0/site/examples/ERP_CORE/sub-015_ses-ERN_task-ERN_report.html)
- [Average report](https://output.circle-artifacts.com/output/job/6f045867-f435-4c17-ac3f-bf3b4c36f5f1/artifacts/0/site/examples/ERP_CORE/sub-average_ses-ERN_task-ERN_report.html)

larsoner:
Still all good after merge with `main` to use new framework:

- [Fixed features table](https://output.circle-artifacts.com/output/job/2e7b522c-4d2b-48c3-b2c5-982b60b64885/artifacts/0/site/examples/examples.html#demonstrated-features)
- [Single subject report](https://output.circle-artifacts.com/output/job/2e7b522c-4d2b-48c3-b2c5-982b60b64885/artifacts/0/site/examples/ERP_CORE/sub-015_ses-ERN_task-ERN_report.html)
- [Average report](https://output.circle-artifacts.com/output/job/2e7b522c-4d2b-48c3-b2c5-982b60b64885/artifacts/0/site/examples/ERP_CORE/sub-average_ses-ERN_task-ERN_report.html)

agramfort:
I still have a script running on this one. I need to look at produced
reports to check it looks correct

give me 2 days max

hoechenberger:
If you want my feedback too you'll also have to wait until the weekend, as I'm sick with a fever 🤒

agramfort:
ok I continued my testing with @SophieHerbst 's data. Currently the report rendering fails if I use something list this:

```
decoding_csp_freqs = {
        'custom_range': [4, 8, 14]
    }
```

it seems it only works if the values are tuples of size 2.

Problem appears in the report code when doing

```python
            for freq_range_name in cfg.decoding_csp_freqs.keys():
                results = csp_freq_results.loc[
                    csp_freq_results['freq_range_name'] == freq_range_name, :
                ]
                all_decoding_scores.append(results['scores'].item())
```

as `freq_range_name` is the same for many frequency bands results has more than one row and the `.item()` fails. To be honest I feel that having all these options of specifying the frequency bands (with tuples, frequency gaps, arrays etc.) makes the code complicated. 

@larsoner or @hoechenberger do you have time to look?

larsoner:
> decoding_csp_freqs = {
>        'custom_range': [4, 8, 14]
>    }

I'm not sure what the docstring says, but to me this is a bad definition of the range. The value should be an array-like of length 2 specifying the lower and upper bounds, so this should raise an error. Maybe instead you want:

```
decoding_csp_freqs = {
    'custom_range_1': [4, 8],
    'custom_range_2': [8, 12],
}
```
This should work and be supported

larsoner:
... if you agree with me that the value should only be an array-like-of-length-2, I'll fix the code and docstring to be this way.

agramfort:
let's do this then. I will change my

'custom': np.arange(5, 41, 1.5)

in my config file.

hoechenberger:
> ... if you agree with me that the value should only be an array-like-of-length-2, I'll fix the code and docstring to be this way.

But then something like @agramfort's


> 'custom': np.arange(5, 41, 1.5)

won't work anymore, and I think it's essential we support this case…

larsoner:
> I think it's essential we support this case…

I'm not convinced here. Currently there are these documented options:

1. Create two frequency bins, one for 4–8 Hz, and another for 8–14 Hz:
    ```
    decoding_csp_freqs = {
        'custom_range': [4, 8, 14]
    }

2. Create 5 equidistant frequency bins from 4 to 14 Hz:
    ```
    decoding_csp_freqs = {
        'custom_range': np.linspace(
            start=4,
            stop=14,
            num=5+1  # We need one more to account for the endpoint!
        )
    }
    ```

3. Create two frquency bins with a "gap" between them, one for 4–8 Hz, and another for 10–14 Hz:
    ```
    decoding_csp_freqs = {
        'lower_range': [4, 8],
        'upper_range': [10, 14]
    ]
    `

I think the only one we *really* need is (3). It's trivial enough for users to specify a couple of entries (1) and do a `linspace` equivalent themselves (2) that we shouldn't bother supporting them internally.

If you're not convinced on this, maybe we can just vote on which to keep? If so, I'm -1 on (1) and (2), and +1 on (3). We can see what @agramfort and @drammock think as well.

drammock:
How is option 1 different from option 2?  Is it just `list` vs `array`?  (seems like internally they would use the same code path).  And if that's right, then wrapping that in a for-loop over each key in `decoding_csp_freqs` and you have one code path that handles all 3 options (unless I'm missing something).

this example:

> 'custom': np.arange(5, 41, 1.5)

would yield 23 bands.  If that's typical, that seems a bit onerous to adapt to pattern (3).

larsoner:
> How is option 1 different from option 2? 

For Option 1, i.e., when `val[2] > val[1]`, it means you should have the equivalent of
```
decoding_csp_freqs = {
    f'{val[0]}_{val[1]}': (val[0], val[1]),
    f{val[1]}_{val[2]}': (val[1], val[2]),
}
```
i.e., the equivalent in (3) form:
```
vals = [4, 8, 12]
for lower, upper in zip(vals[:-1], vals[1:]):
    decoding_csp_freqs[f'{lower}_{upper}'] = (lower, upper)
```
> If that's typical, that seems a bit onerous to adapt to pattern (3).

To me it's just a few lines, so not too painful
```
lims = np.arange(5, 41, 1.5)
for lower, upper in zip(lims[:-1], lims[1:]:
    decoding_csp_freqs[f'{lower}_{upper}'] = (lower, upper)
```
So people can choose what to call these, etc.

But if you agree this is too much to ask of users, then yes some variant of what you wrote above could be adapted to handle this.

larsoner:
... but actually I'm realizing now there is ambiguity between (1) and (2) because it's *not* an `arange`, it's a `linspace`. So
```
decoding_csp_freqs = [4, 8, 14]
```
could be interpreted as boundary edges, or as "14 bands between 4 and 8". And if it's list vs ndarray this seems like a very bad idea because we treat "array-like" the same in our code all over the place (and reasonably so) -- to introduce a difference here to make the behavior different seems like a new pattern people are likely to get wrong.

larsoner:
Wrong button :)

drammock:
oooooh, now I see (I think).  You were internally translating a `list` into `linspace` params?  That seems error-prone.

drammock:
to me a clean / hard-to-mess-up UX would be "pass an iterable of any length, items will be treated as band edges. If you need non-adjacent bands, you can add additional keys."

larsoner:
Works for me

larsoner:
Actually, I think that was might actually how it is/was meant to be and I was misinterpreting the arguments. If so, sorry for the noise!

larsoner:
Okay should be good now:

- [Fixed features table](https://output.circle-artifacts.com/output/job/c1ac61df-7cb9-4b49-9d40-ff83644a74c4/artifacts/0/site/examples/examples.html#demonstrated-features)
- [Single subject report](https://output.circle-artifacts.com/output/job/c1ac61df-7cb9-4b49-9d40-ff83644a74c4/artifacts/0/site/examples/ERP_CORE/sub-015_ses-ERN_task-ERN_report.html)
- [Average report](https://output.circle-artifacts.com/output/job/c1ac61df-7cb9-4b49-9d40-ff83644a74c4/artifacts/0/site/examples/ERP_CORE/sub-average_ses-ERN_task-ERN_report.html)

agramfort (approved):

hoechenberger:
🎈

PR #627, opened 2022-10-09
run epochs decoding in parallel over subjects (no SlidingEstimator here)

agramfort:


mne_bids_pipeline/scripts/sensor/_02_decoding_full_epochs.py
```diff
@@ -183,23 +185,22 @@ def main():
         logger.info(**gen_log_kwargs(message=msg))
         return
 
-    # Here we go parallel inside the :class:`mne.decoding.SlidingEstimator`
-    # so we don't dispatch manually to multiple jobs.
```
agramfort, line 187: this was just wrong

larsoner:
Failures look real and related to multiprocessing

hoechenberger:
Thanks @agramfort!

PR #628, opened 2022-10-12
BUG: Fix bug with naming

larsoner:
Fixes:

> RuntimeWarning: This filename (.../derivatives/mne-bids-pipeline/sub-007/meg/sub-007_task-whatever_power+condition+tfr.h5) does not conform to MNE naming conventions. All tfr files should end with -tfr.h5 or _tfr.h5

Is using `-tfr` instead of `+tfr` the correct fix here @hoechenberger ? Or do we need to update MNE-Python to be more tolerant here somehow?

### DIFF
--- a/mne_bids_pipeline/scripts/report/_01_make_reports.py
+++ b/mne_bids_pipeline/scripts/report/_01_make_reports.py
@@ -986,9 +986,9 @@ def run_report_sensor(
     for condition in conditions:
         cond = config.sanitize_cond_name(condition)
         fname_tfr_pow_cond = str(fname_tfr_pow.copy()).replace("+condition+",
-                                                               f"+{cond}-")
+                                                               f"+{cond}+")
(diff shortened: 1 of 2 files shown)


hoechenberger:
Hello, this is not the correct fix. A dash in BIDS separates entity name and value only

larsoner:
... which begs the question, what is the correct fix?

hoechenberger:
Ignore the warning here or add a switch to MNE to disable the warning, I suppose!

larsoner:
It seems problematic to have a check in MNE-Python that MNE-BIDS-Pipeline can't satisfy. It suggests that one or the other is wrong.

To me a better solution would be to permit `-<stuff>+tfr.h5` in MNE-Python, or to change the naming scheme here -- depending on which of the two packages is doing something wrong/non-standard according to BIDS-compliant naming schemes.

larsoner:
Better like this?

agramfort (approved):

PR #631, opened 2022-10-12
BUG: Fix bugs with reject caching

### LINKED ISSUE #630: BUG: Logging for rejection incorrect
For rejection thresholds, logging omits the subject, so you end up with stuff like:
```
[2022-10-12 14:47:28] │ ⏳️ preprocessing/_06_ptp_reject sub-405 Input:  sub-405_task-lexicaldecision_proc-ssp_epo.fif
[2022-10-12 14:47:28] │ ⏳️ preprocessing/_06_ptp_reject sub-405 Output: sub-405_task-lexicaldecision_proc-clean_epo.fif
[2022-10-12 14:47:29] │ ⏳️ preprocessing/_06_ptp_reject Generating rejection thresholds using autoreject …
[2022-10-12 14:47:32] │ ⏳️ preprocessing/_06_ptp_reject sub-405 Using PTP rejection thresholds: {'mag': 3.017558331475184e-12, 'grad': 1.0491168421379384e-10}
```
Note the missingc `sub-405` in the third line -- we should fix this. This seems minor, but it actually is indicative that we are not consistent, and looking at the code the handling of `*reject` parameters sometimes implicitly violates the `config` -> `cfg` SimpleNamespace conversions we do, thereby making caching not guaranteed to work properly.

To fix this, I propose to:

1. Finally start the `mne_bids_pipeline/_utils.py` helper file
2. Add `_get_reject` to it
3. Refactor epoching/SSP/ICA to use this `_get_reject`, passing in only `cfg.*` vars, no `config.*` vars (which could accidentally break caching)

larsoner:
Closes #630

Not sure if we need `changes.md` for this since it's mostly internal, but I can add one if it helps.

1. Add `mne_bids_pipeline/_utils.py`
2. Use this for rejection
3. Refactor `ica_reject` to compute (possibly using autoreject) exactly once, and persist to disk using `json`

### DIFF
--- a/mne_bids_pipeline/scripts/preprocessing/_04b_run_ssp.py
+++ b/mne_bids_pipeline/scripts/preprocessing/_04b_run_ssp.py
@@ -18,6 +18,7 @@ import config
 from config import gen_log_kwargs, failsafe_run, _update_for_splits
 from config import parallel_func, _script_path
 
+from ..._utils import _get_reject
 
 logger = logging.getLogger('mne-bids-pipeline')
 
@@ -95,16 +96,15 @@ def run_ssp(*, cfg, subject, session, in_files):
         logger.info(**gen_log_kwargs(message=msg, subject=subject,
                                      session=session))
         # Enough to start
-        # TODO: This config.get_ssp_reject violates that we should only use
-        # cfg.values in this function, so caching might not work properly here.
-        # This should be refactored at some point.
         if len(proj_epochs) >= minimums[kind]:
-            if rejects[kind] == 'autoreject_global':
-                reject_epochs = proj_epochs
-            else:
-                reject_epochs = None
-            reject_ = config.get_ssp_reject(
-                ssp_type=kind, epochs=reject_epochs)
+            reject_ = _get_reject(
+                subject=subject,
+                session=session,
+                reject=rejects[kind],
+                ch_types=cfg.ch_types,
+                epochs=proj_epochs,
+                decim=1,
+            )
             proj_epochs.drop_bad(reject=reject_)
         # Still enough after rejection
         if len(proj_epochs) >= minimums[kind]:
@@ -156,6 +156,7 @@ def get_config(
         n_proj_eog=config.n_proj_eog,
         n_proj_ecg=config.n_proj_ecg,
         ssp_meg=config.ssp_meg,
+        ch_types=config.ch_types,
         use_maxwell_filter=config.use_maxwell_filter,
     )
     return cfg
--- a/mne_bids_pipeline/scripts/preprocessing/_06_ptp_reject.py
+++ b/mne_bids_pipeline/scripts/preprocessing/_06_ptp_reject.py
@@ -20,6 +20,7 @@ import config
 from config import gen_log_kwargs, failsafe_run, _update_for_splits
 from config import parallel_func
 
+from ..._utils import _get_reject, _read_json
 
 logger = logging.getLogger('mne-bids-pipeline')
 
@@ -45,6 +46,9 @@ def get_input_fnames_drop_ptp(**kwargs):
     in_files = dict()
     in_files['epochs'] = bids_path.copy().update(
         processing=cfg.spatial_filter)
+    if cfg.spatial_filter == 'ica':
+        in_files['reject'] = bids_path.copy().update(
+            processing='ica', suffix='reject', extension='.json')
     return in_files
 
 
@@ -62,10 +66,20 @@ def drop_ptp(*, cfg, subject, session, in_files):
 
     # Get rejection parameters and drop bad epochs
     epochs = mne.read_epochs(in_files.pop('epochs'), preload=True)
-    reject = config.get_reject(epochs=epochs)
+    reject = _get_reject(
+        subject=subject,
+        session=session,
+        epochs=epochs,
+        reject=cfg.reject,
+        ch_types=cfg.ch_types,
+        decim=1)
+    if cfg.spatial_filter == 'ica':
+        ica_reject = _read_json(in_files.pop('reject'))
+    else:
+        ica_reject = None
 
-    if cfg.ica_reject is not None:
-        for ch_type, threshold in cfg.ica_reject.items():
+    if ica_reject is not None:
+        for ch_type, threshold in ica_reject.items():
             if (ch_type in reject and
                     threshold < reject[ch_type]):
                 # This can only ever happen in case of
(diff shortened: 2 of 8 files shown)


agramfort (commented):
@larsoner can I ask you to comment in the PR to help me understand what you fix here?

If you add autoreject as an option for ICA I don’t think we should offer this option. At least it would deserve discussion

hoechenberger:
I'll only be able to review tomorrow but I'd like to voice that I'm very strongly -1 against naming a module something as generic as "utils", it will become the catch-all for everything over time. Let's better think of better names / namespaces

mne_bids_pipeline/scripts/preprocessing/_04b_run_ssp.py
```diff
@@ -95,16 +96,15 @@ def run_ssp(*, cfg, subject, session, in_files):
         logger.info(**gen_log_kwargs(message=msg, subject=subject,
                                      session=session))
         # Enough to start
-        # TODO: This config.get_ssp_reject violates that we should only use
```
larsoner, line 98: @agramfort this is the important fix -- avoid using `config` inside a cached function, as it potentially silently breaks caching logic

larsoner:
I'll refractor a bit more into namespaces, and also make it clear autoreject is not an option for ICA.

larsoner:
Basically the long term vision is that nothing inside a cached function should call `config.<func>` because it potentially silently breaks our `SimpleNamespace` encapsulation/isolation of used configuration variables. Refactoring these into simple functions in `utils` or some other namespace that take arguments (like `cfg.ica_reject` directly) make immediately clear what configuration variables must be available.

mne_bids_pipeline/scripts/preprocessing/_04b_run_ssp.py
```diff
@@ -156,6 +156,7 @@ def get_config(
         n_proj_eog=config.n_proj_eog,
         n_proj_ecg=config.n_proj_ecg,
         ssp_meg=config.ssp_meg,
+        ch_types=config.ch_types,
```
larsoner, line 159: And here you can see a real potential bug due to bad encapsulation, where rejection parameters could depend on ch_types because it was silently used inside config.get_ssp_reject. This PR forces us to actually have this variable in the SimpleNamespace to use it, rather than allowing it to silently sneak in via a config function call.

I can explain Friday if it's not clear

mne_bids_pipeline/scripts/preprocessing/_06_ptp_reject.py
```diff
@@ -121,9 +135,11 @@ def get_config(
         reject_tmin=config.reject_tmin,
         reject_tmax=config.reject_tmax,
         spatial_filter=config.spatial_filter,
-        ica_reject=config.get_ica_reject(),
+        ica_reject=config.ica_reject,
         deriv_root=config.get_deriv_root(),
         decim=config.decim,
+        reject=config.reject,
```
larsoner, line 141: And perhaps the worst offender, where changing `reject` wouldn't be detected by our caching mechanisms because it was silently used inside a config function call rather than being explicitly passed to a function isolated from config (where variables must be passed explicitly)

agramfort:
@larsoner should I merge or wait for your refactoring here?

larsoner:
Wait for refactoring, also @hoechenberger will want to check the namespaces when I'm done

mne_bids_pipeline/scripts/preprocessing/_04b_run_ssp.py
```diff
@@ -88,23 +89,23 @@ def run_ssp(*, cfg, subject, session, in_files):
         projs[kind] = []
         if not any(n_projs[kind]):
             continue
-        proj_epochs = epochs_fun[kind](raw, ch_name=ch_name[kind])
+        proj_epochs = epochs_fun[kind](
+            raw, ch_name=ch_name[kind], decim=cfg.decim)
```
larsoner, line 93: I consider it a bugfix that we now apply `decim=cfg.decim` here. We probably should have done it before

mne_bids_pipeline/scripts/preprocessing/_06_ptp_reject.py
```diff
@@ -62,10 +63,27 @@ def drop_ptp(*, cfg, subject, session, in_files):
...
     # Get rejection parameters and drop bad epochs
     epochs = mne.read_epochs(in_files.pop('epochs'), preload=True)
-    reject = config.get_reject(epochs=epochs)
+    reject = _get_reject(
+        subject=subject,
+        session=session,
+        reject=cfg.reject,
+        ch_types=cfg.ch_types,
+        param='reject',
+        epochs=epochs,
```
larsoner, line 72: ... and another bugfix (I think) -- these epochs are already decimated, but when `autoreject_global` was used, `decim=decim` was being passed so they were being decimated again.

larsoner:
Okay this one is good to go I think @hoechenberger

larsoner:
(CIs can be ignored, they were green before I pushed two tiny cosmits with `[ci skip]`)

mne_bids_pipeline/scripts/preprocessing/_04b_run_ssp.py
agramfort: I am not sure. Doc says: https://github.com/mne-tools/mne-bids-pipeline/blob/main/mne_bids_pipeline/config.py#L738
and here it's epochs. Now you could argue that decim should apply to raw not epochs but I don't think it's a bug here.

agramfort (approved):

mne_bids_pipeline/scripts/preprocessing/_04b_run_ssp.py
larsoner: Here I mean that we weren't EOG/ECG decimating epochs, even though we should have been (for consistency with task epochs). So I think this PR is more consistent.

Elsewhere this PR fixes a bug where we were decimating epochs *twice*, meaning you end up fitting on the uncleaned, unrejected epochs (good) with decim-squared decimation applied instead of just decim (bad).

hoechenberger (approved):

hoechenberger:
Thanks @larsoner!

larsoner:
Uhoh
```
RuntimeError: Query failed: "Unknown argument "prefix" on field "files" of type "Snapshot"."
```

hoechenberger (approved):

PR #633, opened 2022-10-13
MAINT: Cleaner logic for empty-room and rest data

### LINKED ISSUE #632: BUG: Handling of rest and empty-room
Naively I would have expected that if there are empty room and/or resting state runs in a dataset, they would be analyzed regardless of what the plan is for `noise_cov` during inversion. Thinking about things like connectivity analyses + evoked analyses on the same dataset with the same processing this makes the most sense.

But in the code we have logic like (here for the `frequency_filter` step):
```
    if (cfg.process_er or config.noise_cov == 'rest') and run == cfg.runs[0]:
        noise_task = "rest" if config.noise_cov == "rest" else "noise"
        if cfg.use_maxwell_filter:
            raw_noise_fname_in = bids_path_in.copy().update(
                run=None, task=noise_task
            )
            in_files["raw_noise"] = raw_noise_fname_in
            _update_for_splits(in_files, "raw_noise", single=True)
        else:
            if config.noise_cov == 'rest':
                in_files["raw_rest"] = bids_path_in.copy().update(
                    run=None, task=noise_task)
            else:
                assert config.noise_cov == 'noise'
                ref_bids_path = bids_path_in.copy().update(
                    run=cfg.mf_reference_run,
                    extension='.fif',
                    suffix='meg',
                    root=cfg.bids_root,
                    check=True
                )
                in_files["raw_er"] = ...
```
... which is complicated, to say the least. There is even a bug hidden by non-covered test lines here, because `assert noise_cov == 'noise'` should be `assert noise_cov == 'emptyroom'`. And I think `noise_cov = 'ad_hoc'` is not addressed properly here, either!

I propose we change this to a much simpler scheme like:
```
if has_empty_room:
    in_files['raw-noise'] = ...
if has_resting_state:
    in_files['raw-rest'] = ...
```
This can be implemented at the maxwell filtering stage and the frequency filtering stage. Then the logic about what to use for `noise_cov` + `inverse` only lives in those functions -- and when the param is changed, it only changes those functions.

larsoner:
### Before merging …

- [x] Implement ideas in #632

The logic and flow seems simpler and more transparent to me now. `noise_cov` is no longer used at all in `maxfilter` or `frequency_filter` steps, which is nice!

Closes #632

### DIFF
--- a/mne_bids_pipeline/config.py
+++ b/mne_bids_pipeline/config.py
@@ -3953,7 +3953,7 @@ if (
     raise ValueError(msg)
 
 
-def _update_for_splits(files_dict, key, *, single=False):
+def _update_for_splits(files_dict, key, *, single=False, allow_missing=False):
     if not isinstance(files_dict, dict):  # fake it
         assert key is None
         files_dict, key = dict(x=files_dict), 'x'
@@ -3961,7 +3961,11 @@ def _update_for_splits(files_dict, key, *, single=False):
     if bids_path.fpath.exists():
         return bids_path  # no modifications needed
     bids_path = bids_path.copy().update(split='01')
-    assert bids_path.fpath.exists(), f'Missing file: {bids_path.fpath}'
+    missing = not bids_path.fpath.exists()
+    if not allow_missing:
+        assert not missing, f'Missing file: {bids_path.fpath}'
+    if missing:
+        return bids_path.update(split=None)
     files_dict[key] = bids_path
     # if we only need the first file (i.e., when reading), quit now
     if single:
--- a/mne_bids_pipeline/config.py
+++ b/mne_bids_pipeline/config.py
@@ -3796,8 +3796,7 @@ def import_er_data(
 def import_rest_data(
     *,
     cfg: SimpleNamespace,
-    subject: str,
-    session: Optional[str] = None
+    bids_path_in: BIDSPath,
 ) -> mne.io.BaseRaw:
     """Import resting-state data for use as a noise source.
 
@@ -3805,10 +3804,8 @@ def import_rest_data(
     ----------
     cfg
         The local configuration.
-    subject
-        The subject for whom to import the empty-room data.
-    session
-        The session for which to import the empty-room data.
+    bids_path_in : BIDSPath
+        The path.
 
     Returns
     -------
@@ -3819,7 +3816,7 @@ def import_rest_data(
     cfg.task = 'rest'
 
     raw_rest = import_experimental_data(
-        cfg=cfg, subject=subject, session=session
+        cfg=cfg, bids_path_in=bids_path_in,
     )
     return raw_rest
 
--- a/mne_bids_pipeline/config.py
+++ b/mne_bids_pipeline/config.py
@@ -3761,9 +3761,9 @@ def import_er_data(
     # rank mismatches will still occur (eventually for some configs).
     # But at least using the union here should reduce them.
     # TODO: We should also uso automatic bad finding on the empty room data
+    raw_ref = mne_bids.read_raw_bids(bids_path_ref_in,
+                                     extra_params=cfg.reader_extra_params)
     if cfg.use_maxwell_filter:
-        raw_ref = mne_bids.read_raw_bids(bids_path_ref_in,
-                                         extra_params=cfg.reader_extra_params)
         # We need to include any automatically found bad channels, if relevant.
         # TODO this is a bit of a hack because we don't use "in_files" access
         # here, but this is *in the same step where this file is generated*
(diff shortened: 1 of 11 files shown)


mne_bids_pipeline/scripts/preprocessing/_02_frequency_filter.py
```diff
@@ -68,39 +67,34 @@ def get_input_fnames_frequency_filter(**kwargs):
...
+        do = dict(rest=cfg.process_rest, noise=cfg.process_er)
+        for task in ('rest', 'noise'):
+            key = f'raw_{task}'
+            if cfg.use_maxwell_filter:
+                raw_fname = bids_path_in.copy().update(
+                    run=None, task=task)
             else:
-                assert config.noise_cov == 'noise'
-                ref_bids_path = bids_path_in.copy().update(
-                    run=cfg.mf_reference_run,
```
larsoner, line 97: I think this logic was wrong before -- `mf_reference_run` shouldn't influence non-maxfiltered data. It should just use `runs[0]` to find the empty-room data, right? If so, the logic here should be correct

mne_bids_pipeline/config.py
```diff
@@ -3788,19 +3796,16 @@ def import_er_data(
 def import_rest_data(
     *,
     cfg: SimpleNamespace,
-    subject: str,
-    session: Optional[str] = None
+    bids_path_in: BIDSPath,
```
larsoner, line 3799: I think this was just not being used/tested at all on `main`. `import_experimental_data` cannot be called without a `bids_path_in` argument, and it was below...

mne_bids_pipeline/scripts/preprocessing/_01_maxfilter.py
```diff
@@ -217,7 +217,7 @@ def run_maxwell_filter(*, cfg, subject, session, run, in_files):
         rank_exp = mne.compute_rank(raw_sss, rank='info')['meg']
         rank_noise = mne.compute_rank(raw_noise_sss, rank='info')['meg']
 
-        if config.noise_cov == 'rest':
+        if task == 'rest':
```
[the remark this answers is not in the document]
larsoner: Did you notice that this code is inside a loop over `for task in ('rest', 'noise')`? In other words, if they have resting state data and haven't opted out of processing it, they will get this warning if there is a rank mismatch between their resting state data and their actual-task data. I think it makes sense to emit this warning regardless of what they set `noise_cov` to, no?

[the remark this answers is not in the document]
larsoner: (In other words, it sounds like you're worried that this `task == 'rest'` conditional will only be hit if they have set their `config.task` to `'rest'`, but that's not what this conditional is actually checking)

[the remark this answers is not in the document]
hoechenberger: Ah alright! I was only looking at this in my phone and didn't see enough context. What you say sounds reasonable!

mne_bids_pipeline/config.py
```diff
@@ -226,6 +226,14 @@ class ArbitraryContrast(TypedDict):
 automatically.
 """
 
+process_rest: bool = True
```
agramfort, line 229: Looks good but I would not process rest by default. For cam-can that would mean a lot of computations that I typically don't need when looking at task data.

mne_bids_pipeline/config.py
```diff
@@ -216,7 +216,7 @@ class ArbitraryContrast(TypedDict):
     The ``emptyroom`` subject will be excluded automatically.
 """
 
-process_er: bool = False
+process_er: bool = True
```
agramfort, line 219: I think we should call this `process_empty_room` as `er` is pretty impossible to guess.

mne_bids_pipeline/config.py
larsoner: I think Cam-can is an exception here really. To argue the default should be False means arguing that the default behavior should be "only process a subset of the data, not as much as possible". Are you sure this is the best choice for most datasets / principle of least surprise for people?

larsoner: To give context, datasets I work on with both task and rest data have 30 minutes of task and 5 of rest. And I always want them processed jointly for analysis. I expect more people to be in this position than in the position "I have access to both types of data but want to leave unprocessed the rest data".

Thinking about in the future potentially adding _05_rest_envelope_correlations.py or something, it would also be nice to have this on by default :)

But you work with more datasets and researchers than me, so if you're not convinced by this I'll change it!

larsoner: @agramfort from a chat with @hoechenberger he's on board with the "process all data until told otherwise" idea, so I'm going to keep it that way and merge for now. But let me know if you disagree with this and I can open a follow-up PR to change the default back

larsoner:
Locally I tested a couple OpenNeuro datasets, so as long as the non-OpenNeuro datasets pass I'll merge

PR #634, opened 2022-10-14
ENH: Add find_empty_room step

### LINKED ISSUE #629: ENH: add find_empty_room step
@agramfort experienced some "caching slowdowns" of unknown origin. Then locally for me when I ran a dataset, I also noticed it taking 5-10 sec per subject during the `maxwell_filter` step just to say "cached, skipping".

I tracked this down to the *filename resolution* step of `scripts/preprocessing/_01_maxfilter::get_input_fnames_maxwell_filter`, specifically this line:
```
in_files["raw_er"] = ref_bids_path.find_empty_room()
```
This is because, in order to find the empty room file, mne-bids will at least sometimes resort to reading info from all of the empty room files to find the one with the closest measurement date. (I think ideally this would be stored during the BIDS-ification step, but it's not always done, and we should accommodate these use cases if we can.)

To avoid this, I propose we add a `scripts/init/_01_match_empty_room.py` file that does this *once* ahead of time for datasets that need it, and then saves the match in some new file. This step can be cached just like all the others, and should only need to be run once per dataset (unless the `raw` data actually changes, which fortunately sane caching logic should take care of for us).

The only open question in this implementation is: where can we store the mapping from subject->empty room file? Can we create a new sidecar file somewhere? @hoechenberger

larsoner:
Not sure if a `changelog.md` is really necessary here since this is mostly for internal account purposes.

Also tests https://github.com/hoechenberger/openneuro-py/pull/75 implicitly. Locally I actually get some failures with `ds000248` when downloading the dataset, I wonder if something is broken. We'll see if CircleCI has the same problem...

Eventually we'll want to add the set of empty-room files MNE-BIDS will look at for empty-room files, but that will require refactoring MNE-BIDS a bit. So for now let's assume that users have a complete dataset (if they want one) from the beginning.

Closes #629

### DIFF
--- a/mne_bids_pipeline/_io.py
+++ b/mne_bids_pipeline/_io.py
@@ -1,18 +1,25 @@
 """I/O helpers."""
 
 import json_tricks
+from types import SimpleNamespace
+from mne_bids import BIDSPath
+
+from ._typing import PathLike
 
 
-def _write_json(fname, data):
+def _write_json(fname: PathLike, data: dict) -> None:
     with open(fname, 'w') as f:
         json_tricks.dump(data, fp=f, allow_nan=True, sort_keys=False)
 
 
-def _read_json(fname):
+def _read_json(fname: PathLike) -> dict:
     with open(fname, 'r') as f:
         return json_tricks.load(f)
 
 
-def _empty_room_match_path(run_path, cfg):
+def _empty_room_match_path(
+    run_path: BIDSPath,
+    cfg: SimpleNamespace
+) -> BIDSPath:
     return run_path.copy().update(
         extension='.json', suffix='emptyroommatch', root=cfg.deriv_root)
--- a/mne_bids_pipeline/_io.py
+++ b/mne_bids_pipeline/_io.py
@@ -6,3 +6,13 @@ import json_tricks
 def _write_json(fname, data):
     with open(fname, 'w') as f:
         json_tricks.dump(data, fp=f, allow_nan=True, sort_keys=False)
+
+
+def _read_json(fname):
+    with open(fname, 'r') as f:
+        return json_tricks.load(f)
+
+
+def _empty_room_match_path(run_path, cfg):
+    return run_path.copy().update(
+        extension='.json', suffix='emptyroommatch', root=cfg.deriv_root)
--- a/mne_bids_pipeline/_io.py
+++ b/mne_bids_pipeline/_io.py
@@ -1,19 +1,20 @@
 """I/O helpers."""
 
-import json_tricks
 from types import SimpleNamespace
+
+import json_tricks
 from mne_bids import BIDSPath
 
 from ._typing import PathLike
 
 
 def _write_json(fname: PathLike, data: dict) -> None:
-    with open(fname, 'w') as f:
+    with open(fname, 'w', encoding='utf-8') as f:
         json_tricks.dump(data, fp=f, allow_nan=True, sort_keys=False)
 
 
 def _read_json(fname: PathLike) -> dict:
-    with open(fname, 'r') as f:
+    with open(fname, 'r', encoding='utf-8') as f:
         return json_tricks.load(f)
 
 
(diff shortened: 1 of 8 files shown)


larsoner:
Ready for review/merge from my end @agramfort @hoechenberger . @agramfort this should fix your slowdowns on NFS

mne_bids_pipeline/_io.py
```diff
@@ -6,3 +6,13 @@
...
     with open(fname, 'w') as f:
         json_tricks.dump(data, fp=f, allow_nan=True, sort_keys=False)
+
+
+def _read_json(fname):
+    with open(fname, 'r') as f:
+        return json_tricks.load(f)
+
+
+def _empty_room_match_path(run_path, cfg):
```
hoechenberger, line 16: I know we don't do this consistently with existing code, but could you please add type annotations to the function signatures, both for input and return values? Just for these helpers here in this module, not for the actual processing step script

pyproject.toml
```diff
@@ -56,7 +56,7 @@ tests = [
     "jinja2",
     "black",  # function signature formatting
     "livereload",
-    "openneuro-py",
+    "openneuro-py @ git+https://github.com/larsoner/openneuro-py@iterate#egg=openneuro-py",
```
hoechenberger, line 59: this can be reverted now

larsoner:
```suggestion
    "openneuro-py",
```

mne_bids_pipeline/_io.py
larsoner: Done, feel free to take a look to see if I did it correctly. I'm still a type annotation novice

hoechenberger:
@larsoner I've pushed some additional type annotations & cleanups, should be ready for merging now

hoechenberger (approved):

mne_bids_pipeline/scripts/init/_01_find_empty_room.py
```diff
@@ -0,0 +1,135 @@
...
+from config import _update_for_splits, failsafe_run, gen_log_kwargs
+from mne_bids import BIDSPath
+
+from ..._io import _empty_room_match_path, _write_json
+from ..._logging import logger
+
+
+def get_input_fnames_find_empty_room(
+    *,
+    subject: Optional[str],
```
hoechenberger, line 16:
```suggestion
    subject: str,
```

mne_bids_pipeline/scripts/init/_01_find_empty_room.py
```diff
@@ -0,0 +1,135 @@
...
+        # empty-room matches.
+        pass
+    return in_files
+
+
+@failsafe_run(script_path=__file__,
+              get_input_fnames=get_input_fnames_find_empty_room)
+def find_empty_room(
+    *,
+    subject: Optional[str],
```
hoechenberger, line 58:
```suggestion
    subject: str,
```

agramfort (approved):

hoechenberger:
@larsoner I'm not sure what's going on, but after merging this PR (which passed CI tests) into `main`, a number of tests are now failing on `main`, related to some `KeyErrors`:
https://app.circleci.com/pipelines/github/mne-tools/mne-bids-pipeline/3369/workflows/17213b4c-70ac-4dff-8682-2d7c1a5e31de

Or maybe it's not related to this PR at all, IDK, but we have a problem

larsoner:
The other PR is green, so maybe it was some temporary dataset issue

larsoner:
I feel like I saw this locally on Friday when working on the openneuro-pr PR. I opened this

https://github.com/mne-tools/mne-bids/issues/1085

And will look in more detail tomorrow

dengemann:
Hi @larsoner @hoechenberger I gave empty room finding a try.
My setup is: empty room subject, but no sidecar.

The date matching seems to work but I found the output printed confusing.
The output below suggests that no date-matching empty room files were found but `find_empty_room()` indeed worked (no sidecar). 

Screenshot 2022-10-28 at 15 37 29

Is this only confusing to me or would it make sense to make the output more explicit, e.g., "empty room search via sidecar failed." -> "date-matched search returned 1 file" or something like that.

hoechenberger:
Yes it seems confusing indeed, I don't thing this is intended behavior, @larsoner?

larsoner:
No it's not intended, it just means I didn't replicate the mne-bids logic properly. To fix this I think we should use `return_candidates=True` if available (by inspecting func args) and only report how many were checked if the candidates can be returned

PR #638, opened 2022-10-21
BUG: Fix bug with oversubscription in ICA on M1

### LINKED ISSUE #626: BUG: Resource oversubscription in ERP_CORE ICA
I just ran with `n_jobs` set to 1 and this actually finished faster than with `2`, yet still utilized all CPU cores. Definitely resource oversubscription.

We're running the ERP CORE tests through Dask, if I'm not mistaken. So the problem might lie there…

_Originally posted by @hoechenberger in https://github.com/mne-tools/mne-bids-pipeline/pull/622#discussion_r990508420_

larsoner:
### Before merging …

Fixes the oversubscription problem on my machine. I opened https://github.com/dask/dask/issues/9588 for dask. Until they fix it (if they ever do), we can improve things. This code:
```
$ time mne_bids_pipeline --config mne_bids_pipeline/tests/configs/config_ERP_CORE.py --task ERN --steps=preprocessing/_04a_run --cache=0
```
now decreases from 3m17.387s on `main` to 1m7.732s on this PR. I checked, and there seems to be no need to do this for `loky` backend because the `inner_num_threads` must set it properly in `joblib` (or in `sklearn`, not sure).

Also fixes a bug where `--help` could only be detected by `fire.Fire` when it was the only argument. Now if it appears anywhere in the arg list the `--help` is shown.

Closes #626

### DIFF
--- a/docs/source/changes.md
+++ b/docs/source/changes.md
@@ -388,3 +388,5 @@ authors:
   ({{ gh(606) }} by {{ authors.larsoner }})
 - Fix bug where only the first run was used to compute SSP
   ({{ gh(607) }} by {{ authors.larsoner }})
+- Fix bug with CPU oversubscription when using the Dask backend on macOS M1
+  ({{ gh(638) }} by {{ authors.larsoner }})
--- a/mne_bids_pipeline/run.py
+++ b/mne_bids_pipeline/run.py
@@ -331,6 +331,10 @@ def process(
 
 
 def main():
+    # Fire does not seem to detect a "--help" in all locations, so let's do it
+    # manually.
+    if '--help' in sys.argv or '-h' in sys.argv:
+        sys.argv = sys.argv[:1] + ['--help']
     fire.Fire(process)
 
 
--- a/mne_bids_pipeline/scripts/preprocessing/_04a_run_ica.py
+++ b/mne_bids_pipeline/scripts/preprocessing/_04a_run_ica.py
@@ -17,6 +17,7 @@ from types import SimpleNamespace
 
 import pandas as pd
 import numpy as np
+import threadpoolctl
 
 import mne
 from mne.report import Report
@@ -266,6 +267,10 @@ def run_ica(*, cfg, subject, session, in_files):
     out_files['report'] = bids_basename.copy().update(
         processing='ica+components', suffix='report', extension='.html')
     del bids_basename
+
+    # Prevent oversubscription of CPUs in dask on M1 (probably a dask bug)
+    if cfg.parallel_backend == 'dask':
+        threadpoolctl.threadpool_limits(limits=1, user_api='openmp')
 
     # Generate a list of raw data paths (i.e., paths of individual runs)
     # we want to create epochs from.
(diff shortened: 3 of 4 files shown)


larsoner:
FYI this has been fixed in latest Dask 2022.10.00, but the fix here seems to work fine with it as well

hoechenberger:
Should we simply bump the Dask requirement then? Instead of introducing new code here

hoechenberger:
Great work in any case!

larsoner:
I still think it's useful to have it here, no idea when people will get this latest dask version...

hoechenberger:
Whenever they pip-install the (soon to be released) MNE-BIDS-Pipeline package? 😅

larsoner:
Yeah, not if they already have `dask` installed, though, unless we add version requirement for a new enough version. I guess we could do that instead, but we risk less compatibility/more breakages as we add such things. I'll add an "active comment" to only enable this if the dask version is older than `2022.10.0`. I'm not sure when it was actually fixed between 2022.05 and 2022.10, but it at least works on 2022.10. I looked at the changelog and didn't see anything

hoechenberger:
> unless we add version requirement for a new enough version

Yes this is exactly what I had in mind when I said, let's bump the requirement :)

hoechenberger:
I'm fine with the switch, but I think it's not necessary

hoechenberger:
... besides you're adding a new dependency ...... 😅😅😅😅

I'll leave it to @agramfort ro decide. I'm okay with merging this but I think the easier solution would just be requiring a newer dask version and we'd be done :)

larsoner:
> ... besides you're adding a new dependency ...... 😅😅😅😅

Yes but it's for `packaging` which is required all over the place nowadays, and `threadpoolctl` (with no version pin) which is lightweight and has no dependencies of its own. So I'd argue it's much less invasive to a user -- or potential maintenance on us -- than a `dask` pin.

> I'm okay with merging this but I think the easier solution would just be requiring a newer dask version and we'd be done :)

With this PR, but not necessarily with maintenance burden. Any time you require latest-and-greatest of a package, you decrease adoptability and increase installation issues. For one example, think about `mnelab` for requiring `PySide6` for a while (6 months? 12 months?) and we still can't include it in mne-installers. As another example even here (!), `dask` 2022.10.0 isn't even compatible with the `jupyter` I had installed via `pip` because they disagree about which `tornado` should be used... with how `conda` is at resolving conflicts for example, every version pin makes this problem potentially worse. So it's really nice to avoid them when possible.

larsoner:
> Who uses Dask with the pipeline at this stage anyway? It was broken until just recently....

One last point here... you're thinking about everyone coming from a clean install. If you think about users wanting to install `mne-bids-pipeline` into an existing install, the issues become much more complicated. They do `pip install mne-bids-pipeline` expecting a pure-Python package to just install itself, all of a sudden a bunch of stuff has to be upgraded because we pinned `dask` to this month's version. Or worse, they `conda install mne-bids-pipeline` and end up with a `solving...` for 30 minutes.

agramfort:
And threadpoolctl is already a dependency of joblib.

hoechenberger:
Thanks @larsoner!

PR #639, opened 2022-10-25
rest does not have the run information and has no events to fix

agramfort:
found the problem when running the new CSP code on @SophieHerbst data @hoechenberger 

Once merge I'll report next on the CSP pull request.

### DIFF
--- a/mne_bids_pipeline/_import_data.py
+++ b/mne_bids_pipeline/_import_data.py
@@ -511,12 +511,13 @@ def import_experimental_data(
     _create_bipolar_channels(cfg=cfg, raw=raw, subject=subject,
                              session=session, run=run)
     _drop_channels_func(cfg=cfg, raw=raw, subject=subject, session=session)
-    _rename_events_func(
-        cfg=cfg, raw=raw, subject=subject, session=session, run=run
-    )
     _find_breaks_func(cfg=cfg, raw=raw, subject=subject, session=session,
                       run=run)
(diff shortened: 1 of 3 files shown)


mne_bids_pipeline/_import_data.py
```diff
@@ -511,12 +511,13 @@ def import_experimental_data(
     _create_bipolar_channels(cfg=cfg, raw=raw, subject=subject,
                              session=session, run=run)
     _drop_channels_func(cfg=cfg, raw=raw, subject=subject, session=session)
-    _rename_events_func(
-        cfg=cfg, raw=raw, subject=subject, session=session, run=run
-    )
     _find_breaks_func(cfg=cfg, raw=raw, subject=subject, session=session,
                       run=run)
-    _fix_stim_artifact_func(cfg=cfg, raw=raw)
+    if cfg.task != "rest":
```
larsoner, line 516: I think you also need to check cfg.task_as_rest or something like that

mne_bids_pipeline/_run.py
```diff
@@ -330,6 +330,9 @@ def _update_for_splits(
     bids_path = files_dict[key]
     if bids_path.fpath.exists():
         return bids_path  # no modifications needed
+    if bids_path.copy().update(run=None).fpath.exists():
```
larsoner, line 333: Shouldn't the user have runs=[None] or something in this case?

larsoner:
Good enough for now, if it breaks something or needs to be more general we can follow up. Thanks @agramfort !

PR #644, opened 2022-11-07
BUG: Add hashing of sidecar files

### LINKED ISSUE #643: BUG: Hash sidecars properly
@hoechenberger we have this comment regarding our hashing

https://github.com/mne-tools/mne-bids-pipeline/blob/eb37cf9682da60441ddfd55c6efc21d25fd0b0d2/mne_bids_pipeline/_run.py#L185

For this, do you think it makes sense to call `bids_path.find_matching_sidecar(...)` on all `BIDSPath` instances in `in_files` and add them to the list of files we hash (if such sidecars exist / it returns non-`None`)? If so, should I just iterate over extensions `.json` and `.tsv` to check, or would you recommend something else?

I assume this is the idea behind the comment at least. Then there will be specific other places we'll want to do this, like currently implemented in `_01_find_empty_room`, and probably needed but not implemented for the fiducial matching stuff in coreg (but maybe a BIDSPath to the T1 will already take care of this using the above idea?).

*EDIT*: There are several TODOs in `mne_bids_pipeline/_import_data.py` that might be relevant as well

larsoner:
### Before merging …

Closes #643

### DIFF
--- a/docs/source/changes.md
+++ b/docs/source/changes.md
@@ -173,7 +173,8 @@ authors:
   ({{ gh(603) }} by {{ authors.larsoner }})
 - Caching of pipeline enabled for many pipeline steps
   by default using [`memory_location=True'`][mne_bids_pipeline.config.memory_location]
-  ({{ gh(563) }}, {{ gh (600) }}, {{ gh(608) }}, {{ gh(615) }}, {{ gh(618) }}
+  ({{ gh(563) }}, {{ gh (600) }}, {{ gh(608) }}, {{ gh(615) }}, {{ gh(618) }},
+  {{ gh(644) }}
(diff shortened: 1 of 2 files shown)


hoechenberger (approved):
This looks good to me!!!

larsoner:
CircleCI doesn't appear to be reporting any statuses today, but it's green so I'll merge:

- https://app.circleci.com/pipelines/github/mne-tools/mne-bids-pipeline/3407/workflows/597a7ff7-917c-4f75-be9e-b08bd8c79ad6
- https://output.circle-artifacts.com/output/job/e7e17d71-2c4c-4cae-8f47-88686ae8390a/artifacts/0/site/examples/examples.html#demonstrated-features

PR #647, opened 2022-11-09
ENH: Show progress for fitting

### LINKED ISSUE #570: Print heartbeat log messages
Some steps take quite a while to complete without printing log messages. This causes issues with CircleCI, which may believe the job is stuck; but it's also an issue for users, who cannot be sure that we're still operational either.

The least we should do is periodically print some heartbeat messages.

Better yet would be some useful status info.

Affected steps are e.g. the time-by-time decoding with our ds000246 configuration.

cc @larsoner

larsoner:
### Before merging …

Closes #570

Unfortunately we can't really get this to work for `dask` -- workers can't write to the host to give updates, and the `memmap` approach that works for joblib does not appear to work for dask -- so there no progress bar is shown. The alternative would be to show a progress bar that doesn't update at all, which seems worse than not showing one in the first place.

Locally if I comment out the `parallel_backend = 'dask'` and use the bugfix branch https://github.com/mne-tools/mne-python/pull/11311, then I get nice updates:

https://user-images.githubusercontent.com/2365790/200845655-80dfae6d-698a-4b6c-938f-5638db749475.mp4

### DIFF
--- a/docs/source/changes.md
+++ b/docs/source/changes.md
@@ -198,6 +198,8 @@ authors:
    (CSP).
    ({{ gh(625) }} by {{ authors.crsegerie }}, {{ authors.agramfort }},
     {{ authors.hoechenberger }}, and {{ authors.larsoner }})
+- Add progress bar for time-by-time decoding
+  ({{ gh(647) by {{ authors.larsoner }})
 
 ### Behavior changes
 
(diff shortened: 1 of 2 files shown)


larsoner:
CircleCI is all good https://app.circleci.com/pipelines/github/mne-tools/mne-bids-pipeline/3410/workflows/45fe4374-9e57-4737-b0cc-01a83a8a0ab4

hoechenberger:
Great! Thanks @larsoner

PR #649, opened 2022-11-09
MAINT: Simplify by removing fire

larsoner:
This PR proposes to remove `fire` in support of `optparse`. `Fire` has the advantage of automatically turning a function into a command-line util (neat magic!), but I think it comes with several problematic deficiencies:

1. `-h` and `--help` are not autodetected, we needed to work around it.
2. Extra/wrong arguments/switches are not detected properly, we needed to work around it.
3. We have to do a bunch of `str` mangling to deal with the fact that Fire converts stuff like `--subject=123` into an `int`, `optparse` makes this stuff explicit
4. It does not support `store_true`-style switches like just doing `--interactive` to enable interactive mode -- you have to pass `--interactive=1` which is bad/non-standard Unix command-line interfacing
5. `Fire` is nowhere near as standard as `optparse`, so it takes some time/effort to understand what it's doing, and fix it when it's doing something wrong.

This PR fixes all of these issues by just using `optparse`. The result is one fewer dependency, and more deletions (-93) than insertions (+70) to get there. It also adds some advantages that we get immediately from `optparse`, like making it easy to provide the config as an argument instead of as `--config`, so now you can do `mne_bids_pipeline whatever_config.py` directly (just like we do with `mne_browse_raw --fif file.fif` and `mne_browse_raw file.fif`), which I've implemented here already.

### DIFF
--- a/mne_bids_pipeline/run.py
+++ b/mne_bids_pipeline/run.py
@@ -155,7 +155,8 @@ def _get_script_modules(
 
 
 def main():
-    parser = optparse.OptionParser()
+    from . import __version__
+    parser = optparse.OptionParser(version=f'%prog {__version__}')
     parser.add_option(
         '-c', '--config', dest='config', default=None, metavar='FILE',
         help='The path of the pipeline configuration file to use.')
--- a/mne_bids_pipeline/run.py
+++ b/mne_bids_pipeline/run.py
@@ -1,18 +1,15 @@
 #!/usr/bin/env python
 
+import optparse
 import os
 import sys
 import pathlib
 import logging
+from textwrap import dedent
 import time
 from typing import Union, Optional, Tuple, List, Dict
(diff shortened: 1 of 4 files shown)


larsoner:
... and added a commit to trivially add `mne_bids_pipeline --version` support, which is nice, too

hoechenberger:
@larsoner It seems the tests didn't run?

larsoner:
I `ci skip`ped the `--version` commit because it seemed unnecessary (worked fine locally), the commit before that was green but CircleCI had a bug where it didn't report the status

https://app.circleci.com/pipelines/github/mne-tools/mne-bids-pipeline/3411/workflows/90e6d74e-06e4-469a-ba4e-d09fb3a984ba

hoechenberger:
@larsoner Thanks! I merged `main` and enabled automerge.

Issue #646, opened 2022-11-08
ENH: Namespace validation


larsoner:
I just had an error in my config because I put `session = ...` instead of `sessions = ...`. This is a pretty general problem I think.

One way to solve this would be:

1. Document/standardize around the idea that, as much as possible, "public" variable names in `config.py` files should be reserved for standard config values; other variables (for example functions you want to run on import, etc.) should be prefixed by `_` or be very dissimilar from config value names
2. On reading the custom `config`, iterate over our standard variable names and do `get_close_matches` to the custom config variable names (except those prefixed by `_`)
3. Add a `config_validation = 'raise' | 'warn' | 'ignore'` to decide what to do when we find any close matches, with a default of `'raise'`

I don't think this will lead to too many false alarms, and is easily worked around when there are false alarms at the user's end by prefixing the problematic variable with `_` or renaming it, or by changing the validation to `warn` or `ignore`.

larsoner:
... this approach seems cleaner and simpler to me than #360 / requiring users to `del`. To me that's too much work with little benefits over the `get_close_matches` approach outlined above

hoechenberger:
I think we should make use of pydantic, https://pydantic-docs.helpmanual.io/

Let's discuss on Friday

larsoner:
From a very quick look, that seems like a way to make sure variables are the right type / value etc. using type annotations. I'm not sure it would help with `session` vs `sessions`, so seems mostly separable from the problem I'm trying to solve here.

larsoner:
In other words, I think we face two issues:

1. how do we avoid people's misspellings silently causing config vars to be ignored/unset
2. how do we validate the config vars that users actually set

My proposal here is solely to deal with (1), and I think pydantic would solely deal with (2).

One way to make the solution to (2) also potentially a solution for (1) is what's discussed in #360, namely passing all variables in `config` to some class constructor, thereby requiring users to `del`. But as I allude to above this seems like a lot of extra work for users -- do we really want to require users to `del np, Path, ...`, and any time they iterate over something in their `config` now they need to `del ii` or whatever as well? To me it's a lot of extra work/requirement without much (any in practice?) benefit over the `get_close_matches` approach.

hoechenberger:
Agreed, `get_close_matches()` seems to be the way to go for what you have in mind

Pydantic would then be a second step, where we can actually validate the input

PR #650, opened 2022-11-09
MAINT: Move config validation out of config.py

larsoner:
Let's get `config.py` to be a clean, simple file containing just config vars + descriptions!

1. Move all validation and private vars out of `config.py` to `_config_utils.py`
2. Move custom-config loading to `_config_utils.py`
3. Remove many `import config` calls by using `script_module.main(config=config)` instead
4. Replace remaining `import config` calls with `_import_config()`, which does the default load, user load, and validation steps

This is all internal, so no changelog necessary I think. This will make it clear where/when things like #646 should be done in the code.


hoechenberger:
@larsoner I've moved the config import to a new module and added a few type annotations to the function signatures, let's see if CI likes it… WDYT?

larsoner:
Works for me -- I'll resolve conflicts, push, and mark for merge

PR #651, opened 2022-11-09
Don't spit out traceback when CLI parameters are missing

hoechenberger:
`main`:
```
❯ mne_bids_pipeline
  File "/Users/hoechenberger/mambaforge/envs/mne/bin/mne_bids_pipeline", line 8, in <module>
    sys.exit(main())
  File "/Users/hoechenberger/Development/mne-bids-pipeline/mne_bids_pipeline/run.py", line 71, in main
    raise ValueError(bad_msg)
ValueError: You must specify a configuration file as a single argument or with --config.
```
PR:
```
❯ mne_bids_pipeline
[2022-11-09 22:14:59] │ 😵  You must specify a configuration file as a single argument or with --config.
```

### Before merging …


mne_bids_pipeline/_logging.py
```diff
@@ -69,7 +69,6 @@ def gen_log_kwargs(
     session: Optional[Union[str, int]] = None,
     run: Optional[Union[str, int]] = None,
     emoji: str = '⏳️',
-    script_path: Optional[PathLike] = None,
```
hoechenberger, line 72: This parameter was unused, so I removed it

mne_bids_pipeline/_logging.py
```diff
@@ -80,8 +79,12 @@ def gen_log_kwargs(
 
     message = f' {message}'
 
-    script_path = pathlib.Path(os.environ['MNE_BIDS_STUDY_SCRIPT_PATH'])
-    step_name = f'{script_path.parent.name}/{script_path.stem}'
+    script_path = os.environ.get('MNE_BIDS_STUDY_SCRIPT_PATH')
```
hoechenberger, line 82: The env var gets set once we're running a processing script; it however run.py wants to log a message, this variable is not set, so we have to handle this

pyproject.toml
```diff
@@ -66,7 +66,7 @@ tests = [
 ]
 
 [project.scripts]
-mne_bids_pipeline = "mne_bids_pipeline.run:main"
+mne_bids_pipeline = "mne_bids_pipeline.run:main_cli"
```
hoechenberger, line 69: the new behavior only affects the CLI; if instead one programmatically calls `run.main()`, everything remains unchanged and one gets the full traceback

hoechenberger:
@larsoner WDYT?

hoechenberger:
The only annoying thing is the **two** spaces after the emoji!

larsoner:
I think the standard behavior here might be to show the equivalent of `--help` when no arguments or switches are provided. But even in that case if they do something like `mne_bids_pipeline --debug` maybe we want this message, not sure. So I think this is okay/explicit enough

hoechenberger:
I'm not sure I fully understand – are you okay with these changes?

pyproject.toml
larsoner: I think in practice nobody will programatically call `run.main()` (and it should be private).

Have you checked to make sure `run_tests.py` with `--debug` for example give reasonable outputs? And with `on_error = 'abort'` this doesn't now print redundant information or anything? If so then I'm okay with this

hoechenberger:
@larsoner I've checked and the behavior for everything happening inside the processing scripts is unchanged; this here only affects errors raised in what was previously called `run.py` -- which I have now renamed to `_main.py`

PR #652, opened 2022-11-09
ENH: Report along the way

### LINKED ISSUE #648: ENH: Improve report generation and caching
Currently the flow is: run all steps, then generate report. An alternative equivalent flow is: run each step, and within each step, add to the report any relevant/updated information.

Changing to this new flow will make report generation *much faster*, because sections only need to be updated/reproduced when things in them actually change. In other words, we get caching for free by piggybacking report generation within individual steps. It also helps with namespaces / encapsulation because it's easier to see/know what needs to go and be used in each report section. If you add something new to a given step -- or want more information from it in the report -- the relevant report code now lives in the given step, rather than in the monolithic `_01_make_report.py`.

In practical terms, I think this "just" requires using the Report save/load functionality to persist sections to disk (and save/write the HTML at the conclusion of each step, which should be fast enough), and then moving a bunch of code lines to the correct scripts. I think as long as we use the correct/consistent section names. And then hopefully `_01_make_report.py` goes away completely. At least in principle I think this is all that will be necessary.

larsoner:
### Before merging …

Adds a dependency on `filelock` but I don't think this is too bad since a) it looks maintained, and b) it was already on my system without me manually installing it, suggesting that some other dependency already uses it.

I doubt this will succeed on the first try, but let's see!

Closes #648


larsoner:
All green, but not ready yet since some sections get replicated

https://output.circle-artifacts.com/output/job/c5e6c589-bf89-4fc2-ab35-bc3012de2060/artifacts/0/site/examples/ERP_CORE/sub-015_ses-ERN_task-ERN_report.html

larsoner:
This PR is working but should wait for #658

hoechenberger:
@larsoner We'vegot merge conflicts now 😵‍💫

hoechenberger:
I enabled automerge. Thanks @larsoner!

PR #653, opened 2022-11-09
Add CLI option to generate a template configuration the user can edit freely

hoechenberger:
### Before merging …


hoechenberger:
@larsoner This is WIP, what do you think of the idea?

larsoner:
I like the idea -- but to follow how Unix config files work, instead of `copyfile` eventually we should read and then write while commenting out the Python code entries, so a user starts out with a no-op config when they then uncomment the things they want. But we could do that in a follow-up PR

hoechenberger:
@larsoner I've implemented  a very lazy approach to commenting out (most) parts of the template configuration file. It now looks like this:

```python
# """Default settings for data processing and analysis.
# """

from typing import Optional, Union, Iterable, List, Tuple, Dict, Callable

from numpy.typing import ArrayLike

import mne
from mne_bids import BIDSPath
import numpy as np

from mne_bids_pipeline._typing import PathLike, ArbitraryContrast, Literal


# Config parameters
# -----------------

# study_name: str = ''
# """
# Specify the name of your study. It will be used to populate filenames for
# saving the analysis results.

# ???+ example "Example"
#     ```python

#     study_name = 'my-study'
#     ```
# """

# bids_root: Optional[PathLike] = None
```

I think we can always make it nicer later on

I have also updated the installation docs, recommending to install from PyPI. Time to finally make a release!!!

hoechenberger:
@larsoner I've also made `config.py` private

mne_bids_pipeline/tests/test_documented.py
```diff
@@ -26,7 +26,7 @@ def test_options_documented():
     settings_path = root_path.parent / "docs" / "source" / "settings"
     assert settings_path.is_dir()
     in_doc = set()
-    key = "::: mne_bids_pipeline.config."
+    key = "::: mne_bids_pipeline._config."
```
larsoner, line 29: This probably means that all `docs/` options need to be updated to reflect this, and all `changelog` entries...

Maybe it's worth it to make this private?

larsoner: Oh you've done this already above. I guess it's okay then

hoechenberger: I didn't update any changelog entries though 🤔

larsoner: Yes one problem with `mkdocs` is that there is no way AFAIK to treat warnings as errors, so I only notice bad links when I build locally and see the warnings

hoechenberger: I jist pushed those changes

mne_bids_pipeline/_main.py
```diff
@@ -60,10 +68,40 @@ def main():
...
+        config: List[str] = []
+        with open(config_source_path, 'r', encoding='utf-8') as f:
+            for line in f:
+                line = (
+                    line if line.startswith(('#', '\n', 'import', 'from'))
+                    else f'# {line}'
+                )
+                config.append(line)
+
+        config_target_path.write_text(''.join(config), encoding='utf-8')
```
larsoner, line 90: Can you put this in a helper function in `mne_bids_pipeline/_*.py` somewhere? It helps keep `_main` just about triaging args and control flow, especially once we decide to make this more sophisticated...

hoechenberger: Done

larsoner:
Last thing I think is just some simple pytest in `mne_bids_pipleine/tests/test_*` somewhere. Basically I think you can use `tmp_path` and write a config there, then maybe `importlib.import` it and make sure that it has no entries. This ensures both that it's actually valid Python, and that it has no entries in it by default...

larsoner:
I'll add this test quickly

larsoner:
Marking for merge, thanks in advance @hoechenberger !

PR #665, opened 2022-11-15
ENH: Add misspelling check

### LINKED ISSUE #646: ENH: Namespace validation

larsoner:
### Before merging …

Closes #646

### DIFF
--- a/docs/source/changes.md
+++ b/docs/source/changes.md
@@ -293,6 +293,8 @@ authors:
 - Environment variables are no longer used to control execution and variables,
   use command-line switches instead
   ({{ gh(663) }} by {{ authors.larsoner}} )
+- Misspellings are now checked in config
+- ({{ gh(665) }} by {{ authors.larsoner }} )
 
 ### Code health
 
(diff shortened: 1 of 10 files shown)


larsoner:
All green, good to go here I think @hoechenberger !

agramfort (approved):

hoechenberger:
Thanks @larsoner

PR #666, opened 2022-11-16
MAINT: Split bad-finding step

### LINKED ISSUE #657: Reports do not contain unfiltered data
Today I showed a report to a colleague of mine and we figured it didn't contain the raw data before filtering. I believe we used to include this... So this seems like a bug to me that it's missing

larsoner:
### Before merging …

- [x] Separate bad-finding into its own step
- [x] Add raw data plots to report (before preproc)
- [x] Fix logic with `exec_params` by including it in the kwargs, but excluding it from cache via `ignore` arg in `memory.cache(...)`.
- [x] Restore grand-average source space report that was missing/dropped accidentally
- [x] Add bad-channel finding and use for empty-room data

Closes #657

It (hopefully) simplifies a bit the file logic. Let's see what the CIs say.

### DIFF
--- a/docs/source/changes.md
+++ b/docs/source/changes.md
@@ -205,6 +205,9 @@ authors:
 - Make report generation happen within relevant steps instead of at the end
   of all steps
   ({{ gh(652) }} by {{ authors.larsoner }})
+- Initial raw data plots are now added to reports and bad channel detection
+  is executed in a dedicated step
+  ({{ gh(666) }} by {{ authors.larsoner }})
 
 ### Behavior changes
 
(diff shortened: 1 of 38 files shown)


larsoner:
Okay @hoechenberger this one is finally green and ready to go I think.

Some stuff might be simplified by having `_01_data_quality.py` write out a new `_raw.fif` with the break annotations, bads, etc. applied to it. But I wasn't sure if the extra storage space required would be worth it. Maybe in some follow-up PR...

hoechenberger:
> Some stuff might be simplified by having `_01_data_quality.py` write out a new `_raw.fif` with the break annotations, bads, etc. applied to it. But I wasn't sure if the extra storage space required would be worth it.

We could use `write_raw_bids()` with `symlink=True` but yes, that's probably something for a followup PR!

hoechenberger:
Thanks @larsoner!

PR #670, opened 2022-11-30
ENH: Check for removed names

larsoner:
### Before merging …

More validation -- look for at least one removed name.

@hoechenberger there are probably lots of others to add here. Feel free to push directly based on memory if you want, or you can do it in a follow-up PR. This one just adds `debug`->`on_error='debug'`.


larsoner:
This is the last blocker for 1.0.0 and seems pretty straightforward so I'll merge. We can add more names later I think, as they could easily be in a 1.0.1.

PR #672, opened 2022-11-30
MAINT: Remove URL-based reqs

larsoner:
I tried to upload to PyPI and got:
```
Invalid value for requires_dist. Error: Can't have direct dependency: "autoreject @ https://api.github.com/repos/autoreject/autoreject/zipball/master ; extra == 'dev'"
```
It seems that PyPI does not want any requirements to be URL-based, so this PR removes the `dev` target that contained all URL-based `main` requirements.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -45,7 +45,8 @@ jobs:
           name: Get Python running
           command: |
             pip install --upgrade --progress-bar off pip setuptools
-            pip install -ve .[tests,dev]
+            pip install --upgrade --progress-bar off "autoreject @ https://api.github.com/repos/autoreject/autoreject/zipball/master" "mne[hdf5] @ https://api.github.com/repos/mne-tools/mne-python/zipball/main" "mne-bids[full] @ https://api.github.com/repos/mne-tools/mne-bids/zipball/main"
+            pip install -ve .[tests]
             pip install PyQt6
       - run:
           name: Check Qt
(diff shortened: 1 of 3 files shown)


larsoner:
... whereas on this PR:

https://pypi.org/project/mne-bids-pipeline/

hoechenberger:
Do we need to cut a new release of autoreject?

hoechenberger:
Oh never mind, I misunderstood what was going on here. Mmh that's a bummer really!

hoechenberger (approved):

hoechenberger:
> I tried to upload to PyPI and got:

Allegedly, `twine check` should catch this; we should add this to our workflow

larsoner:
> Allegedly, twine check should catch this; we should add this to our workflow

Maybe it won't actually? I didn't read the whole thing but this is still open: https://github.com/pypa/twine/issues/430

But in any case, I now added:

1. A wheel build action, including a `twine check --strict`
2. A PyPI-upload action with proper PyPI var
3. `setuptools_scm` var `tag_regex` to strip the `v` from tags so that `v1.0.0` can be the name of our tag in our first real release (following MNE-Python convention)

larsoner:
> Do we need to cut a new release of autoreject?

This would be a good idea actually, since the latest release there was some fix of `nan` handling I think?

hoechenberger:
Thanks @larsoner!

PR #674, opened 2022-12-01
DOC: Changes

larsoner:
- Future-proof the changelog at least a little bit by using `include`s, set up versions:

    Screenshot from 2022-12-01 13-04-46

- Change site colors to better match MNE ecosystem that uses pydata-sphinx-theme (asthethically I like it slightly less, but I think the improved user consistency is worth it):

     | MNE | M-B-P |
     | -- | -- |
     | Screenshot from 2022-12-01 13-00-41 | Screenshot from 2022-12-01 13-00-30 |
- Add Mastodon link
- Fix deprecated section use


larsoner:
Also fixed the `[circle ds000117]` support. I used it which is why there is only a report for that dataset in this CircleCI output, but you can see the relevant changes to the theme + changelog:

https://output.circle-artifacts.com/output/job/39e48f69-010b-468b-b531-9b5e8db1b082/artifacts/0/site/changes.html

@hoechenberger okay to merge? If so feel free and I'll push out v1.0.0

docs/mkdocs.yml
```diff
@@ -15,11 +15,16 @@ theme:
         - navigation.tabs
         - navigation.instant  # Load pages via XMLHttpRequest (XHR)
         - navigation.tracking
+    palette:
+        primary: white
```
larsoner, line 19: @hoechenberger FYI these are the two lines we'll need to revert if the theme-toward-pydata is too much :)

I'll actually merge as-is as we can always fix it easily afterward, I don't think it'll matter for cutting the release

PR #688, opened 2022-12-13
MAINT: Black, ruff, pre-commit

### LINKED ISSUE #686: Try out ruff?
As a replacement for flake8:

https://pypi.org/project/ruff/

WDYT @larsoner?

larsoner:
Closes #686

- Format with black
- Add reformatting commit to ignore-revs
- Change from flake8 to ruff
- Add to CIs
- Add to pre-commit config

Should be merged rather than squash-and-merged so that the `black` reformatting commit is ignored properly.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -33,8 +33,6 @@ check:
 trailing-spaces:
 	find . -name "*.py" | xargs perl -pi -e 's/[ \t]*$$//'
 
-flake: ruff
-
 ruff:
 	ruff .
 	@echo "ruff passed"
--- a/Makefile
+++ b/Makefile
@@ -33,9 +33,11 @@ check:
 trailing-spaces:
 	find . -name "*.py" | xargs perl -pi -e 's/[ \t]*$$//'
 
-flake:
-	flake8
-	@echo "flake8 passed"
+flake: ruff
+
+ruff:
+	ruff .
+	@echo "ruff passed"
 
 codespell:  # running manually; auto-fix spelling mistakes
 	@codespell --write-changes $(CODESPELL_DIRS)
--- a/mne_bids_pipeline/_config.py
+++ b/mne_bids_pipeline/_config.py
@@ -1,8 +1,7 @@
 """Default settings for data processing and analysis.
 """
 
-from typing import (
-    Optional, Union, Iterable, List, Tuple, Dict, Callable, Literal)
+from typing import Optional, Union, Iterable, List, Tuple, Dict, Callable, Literal
 
 from numpy.typing import ArrayLike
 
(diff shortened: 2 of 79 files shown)


larsoner:
I'm going to assume this is some weird GitHub problem and try again tomorrow...

larsoner:
Ahh, GH is degraded at the moment:

> Investigating - We are investigating reports of degraded performance for Git Operations and Pull Requests.
> Dec 13, 2022 - 17:39 UTC

Makefile
```diff
@@ -33,9 +33,11 @@ check:
 trailing-spaces:
 	find . -name "*.py" | xargs perl -pi -e 's/[ \t]*$$//'
 
-flake:
-	flake8
-	@echo "flake8 passed"
+flake: ruff
```
hoechenberger, line 36: I think this should simply be killed.

mne_bids_pipeline/_config.py
```diff
@@ -452,7 +452,7 @@
     ```
 """
 
-min_break_duration: float = 15.
+min_break_duration: float = 15.0
```
hoechenberger, line 455: Hah, nice!!!

mne_bids_pipeline/_config.py
```diff
@@ -933,12 +933,7 @@
     ```
 """
 
-contrasts: Iterable[
-    Union[
-        Tuple[str, str],
-        ArbitraryContrast
-    ]
-] = []
+contrasts: Iterable[Union[Tuple[str, str], ArbitraryContrast]] = []
```
hoechenberger, line 936: So difficult to read. Time to drop support for Python <3.10 so we can use `X | Y` instead of `Union[X, Y]` 😎

No gut really, here we'd really gain a usability boost from a switch... maybe next spring or so??

larsoner: My vote is to go with whenever MNE-Python switches to 3.10 as the min based on EOL

larsoner:
I temporarily enabled merge commits for the repo so we can merge this one rather than rebase+merge or squash+merge. I also removed the `make flake` alias. Okay to "merge commit" when green?

drammock:
why not split out the `black` commit as a separate PR?  and/or rebase to squash all the other commits in this PR into one, and then rebase-merge?

hoechenberger:
Yes, good idea

hoechenberger:
> I temporarily enabled merge commits for the repo so we can merge this one rather than rebase+merge or squash+merge. I also removed the `make flake` alias. Okay to "merge commit" when green?

Okay with me

hoechenberger:
> why not split out the `black` commit as a separate PR?  and/or rebase to squash all the other commits in this PR into one, and then rebase-merge?

This might actually be the better idea 🤓

larsoner:
> why not split out the black commit as a separate PR? and/or rebase to squash all the other commits in this PR into one, and then rebase-merge?

It's certainly another way to do it, but more work at this point. But since @hoechenberger is also for it, I'll do it that way

PR #702, opened 2023-01-11
Add notch filter

### LINKED ISSUE #701: Notch filter
Hi !

I am using the mne-bids-pipeline on my MEG dataset. I have seen that, at the moment, there is no possibility to add a notch filter in the preprocessing. I was wondering if this feature will be added to the pipeline soon and if not, if there is a possibility as a user to integrate it for our needs? I would like to remove the power line noise but I would like to avoid deleting all frequencies after 60 Hz as I might used them later when doing time-frequency analysis.

Thanks a lot!

hoechenberger:
Fixes #701

### Before merging …

### DIFF
--- a/mne_bids_pipeline/_config.py
+++ b/mne_bids_pipeline/_config.py
@@ -683,19 +683,19 @@ End time of the interpolation window in seconds.
 l_freq: Optional[float] = None
 """
 The low-frequency cut-off in the highpass filtering step.
-Keep it ``None`` if no highpass filtering should be applied.
+Keep it `None` if no highpass filtering should be applied.
 """
 
 h_freq: Optional[float] = 40.0
 """
 The high-frequency cut-off in the lowpass filtering step.
-Keep it ``None`` if no lowpass filtering should be applied.
+Keep it `None` if no lowpass filtering should be applied.
 """
 
 notch_freq: Optional[Union[float, Iterable[float]]] = None
 """
 Notch filter frequency. More than one frequency can be supplied, e.g. to remove
-harmonics. Keep it ``None`` if no notch filter should be applied.
+harmonics. Keep it `None` if no notch filter should be applied.
 
 Note: Note
     The notch filter will be applied before high- and lowpass filtering.
--- a/mne_bids_pipeline/_config.py
+++ b/mne_bids_pipeline/_config.py
@@ -725,11 +725,14 @@ lowpass filter. By default it's `'auto'` and uses default MNE
 parameters.
 """
 
-notch_trans_bandwidth: Union[float, Literal["auto"]] = "auto"
+notch_trans_bandwidth: float = 1.0
 """
-Specifies the transition bandwidth of the
-notch filter. By default it's `'auto'` and uses default MNE
-parameters.
+Specifies the transition bandwidth of the notch filter. The default is `1.`.
+"""
+
+notch_widths: Optional[Union[float, Iterable[float]]] = None
+"""
+Specifies the width of each stop band. `None` uses the MNE default.
 """
 
 raw_resample_sfreq: Optional[float] = None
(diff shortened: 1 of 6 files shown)


mne_bids_pipeline/steps/preprocessing/_03_frequency_filter.py
```diff
@@ -89,10 +116,6 @@ def filter(
         h_freq=h_freq,
         l_trans_bandwidth=l_trans_bandwidth,
         h_trans_bandwidth=h_trans_bandwidth,
-        filter_length="auto",
-        phase="zero",
-        fir_window="hamming",
-        fir_design="firwin",
```
hoechenberger, line 95: MNE defaults

hoechenberger:
@agramfort @larsoner I'm not sure when I'll be able to get back to this, so please feel free to take over if you have a couple minutes to spare! Or drop a comment on whether you agree with the approach taken here (adding a new function and thus some redundancy, instead of amending the existing `filter()` function)

mne_bids_pipeline/_config.py
```diff
@@ -683,13 +683,32 @@
 l_freq: Optional[float] = None
 """
 The low-frequency cut-off in the highpass filtering step.
-Keep it None if no highpass filtering should be applied.
+Keep it ``None`` if no highpass filtering should be applied.
```
hoechenberger, line 686:
```suggestion
Keep it `None` if no highpass filtering should be applied.
```

mne_bids_pipeline/_config.py
```diff
@@ -683,13 +683,32 @@
...
 """
 
 h_freq: Optional[float] = 40.0
 """
 The high-frequency cut-off in the lowpass filtering step.
-Keep it None if no lowpass filtering should be applied.
+Keep it ``None`` if no lowpass filtering should be applied.
```
hoechenberger, line 692:
```suggestion
Keep it `None` if no lowpass filtering should be applied.
```

mne_bids_pipeline/_config.py
```diff
@@ -683,13 +683,32 @@
...
+"""
+
+notch_freq: Optional[Union[float, Iterable[float]]] = None
+"""
+Notch filter frequency. More than one frequency can be supplied, e.g. to remove
+harmonics. Keep it ``None`` if no notch filter should be applied.
```
hoechenberger, line 698:
```suggestion
harmonics. Keep it `None` if no notch filter should be applied.
```

larsoner (commented):
I think you just need to

1. Add to documentation (see failing CI)
2. Add to some example script (maybe increase the `l_freq` of some dataset and add a `notch_freq` to it so we can see that it works?

hoechenberger:
Yes!! I really just pushed "what I had", it's by no means finished! Feel free to take over if you want. Otherwise I can do it tomorrow

mne_bids_pipeline/_config.py
```diff
@@ -706,6 +725,16 @@
 parameters.
 """
 
+notch_trans_bandwidth: float = 1.0
```
hoechenberger, line 728: This is now inconsistent with the other trans bandwidth settings. Is that intentional?

larsoner: Yes there is no `'auto'` for `notch_filter`

https://mne.tools/dev/generated/mne.io.Raw.html#mne.io.Raw.notch_filter

hoechenberger: Oh, right!

hoechenberger:
Thanks @larsoner!

hoechenberger:
@larsoner Lets cut a new release, WDYT?

larsoner:
I guess this would be 1.1 then, sure!

PR #718, opened 2023-03-01
BUG: Fix bug with dev version numbering

larsoner:
With this change I get `1.2.0.dev4+g12735039.d20230301` instead of `1.1.1.dev4+g12735039.d20230301` as the version (note the proper minor version bump).

Also a tiny change where `__version__` always gets set.

No need for `changelog` update I think since this is a dev-facing change


hoechenberger:
Perfect!

PR #720, opened 2023-03-08
Caching mne.get_datatypes to speed up _get_ignore_datatypes

apmellot:
### Before merging …

### DIFF
--- a/docs/source/v1.2.md.inc
+++ b/docs/source/v1.2.md.inc
@@ -2,6 +2,8 @@
 
 ### :new: New features & enhancements
 
+- mne.get_datatype is now cached to reduce computation time when large datasets are being processed. (#720 by @apmellot)
+
 [//]: # (- Whatever (#000 by @whoever))
 
 [//]: # (### :warning: Behavior changes)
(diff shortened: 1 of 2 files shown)


hoechenberger (commented):
Looks good! This performance improvement deserves to be credited in our changelog, don't you think? :)

apmellot:
Like that? It's the first time I update a changelog haha

docs/source/v1.2.md.inc
```diff
@@ -2,6 +2,8 @@
 
 ### :new: New features & enhancements
 
+- mne.get_datatype is now cached to reduce computation time when large datasets are being processed. (#720 by @apmellot)
```
hoechenberger, line 5: Very nice! I always like to have things a little bit less technical so users will get a feel for whether a change is going to help them or not. How about this:

```suggestion
- We improved caching to reduce initialization time when large datasets are being processed. (#720 by @apmellot)
```

hoechenberger (approved):
Thanks @apmellot 

I'll let you merge @agramfort

PR #732, opened 2023-04-04
BUG: Fix bug with sys info addition

### LINKED ISSUE #728: Add version info to mne-bids-pipeline
> it would be helpful if the version of the bids-pipeline used per step was to be found in the output for reporting in a paper and replicability

_Originally posted by @SophieHerbst in https://github.com/mne-tools/mne-bids-pipeline/issues/609#issuecomment-1482981862_

larsoner:
### Before merging …

Closes #728

### DIFF
--- a/docs/source/v1.3.md.inc
+++ b/docs/source/v1.3.md.inc
@@ -12,6 +12,6 @@
 
 [//]: # (- Whatever (#000 by @whoever))
 
-[//]: # (### :bug: Bug fixes)
+### :bug: Bug fixes)
 
-[//]: # (- Whatever (#000 by @whoever))
+- Fix bug with `mne.sys_info` insertion in reports (#732 by @larsoner)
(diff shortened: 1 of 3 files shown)


docs/source/v1.3.md.inc
```diff
@@ -12,6 +12,7 @@
 
 [//]: # (- Whatever (#000 by @whoever))
 
-[//]: # (### :bug: Bug fixes)
+### :bug: Bug fixes)
 
-[//]: # (- Whatever (#000 by @whoever))
+- Fix pandas 2.0 compatibility (#732 by @larsoner)
```
hoechenberger, line 17: I'd add this to the code health section instead as

```
- Support report generation with pandas 2.0 (#732 by @larsoner)
```
WDYT? Since it's not a bug

larsoner: To me it was a bug that we were knowingly using a deprecated API, and users will see it as an error (no pipeline can even run with latest pandas). So to me it's above the level of "code health" which sounds like stuff maintainers should mostly care about

hoechenberger:
Thanks @larsoner

PR #733, opened 2023-04-05
Always generate grand-averaged CSP decoding results if CSP was used on single-subject level

### LINKED ISSUE #731: Setting decode_csp does run CSP, but doesn't create grand average CSP results when running the group average step
The reason is that we decide to run CSP if `decoding_csp is True`, but when generating the grand average, we check for the combination of **two** configuration values:

https://github.com/mne-tools/mne-bids-pipeline/blob/05b72e82d4c36f1a81bd6424a4301974cd6d7f3b/mne_bids_pipeline/steps/sensor/_99_group_average.py#L668

This should probably be changed to just
```python
 if cfg.decoding_csp: 
```

hoechenberger:
Fixes #731

### Before merging …

### DIFF
--- a/docs/source/v1.3.md.inc
+++ b/docs/source/v1.3.md.inc
@@ -16,3 +16,4 @@
 
 - Fix pandas 2.0 compatibility (#732 by @larsoner)
 - Fix bug with `mne.sys_info` insertion in reports (#732 by @larsoner)
+- Always generate CSP decdoing grand average analysis if CSP decoding was used on the single-subjec level (#733 by @hoechenberger)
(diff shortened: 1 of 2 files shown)


docs/source/v1.3.md.inc
```diff
@@ -16,3 +16,4 @@
 
 - Fix pandas 2.0 compatibility (#732 by @larsoner)
 - Fix bug with `mne.sys_info` insertion in reports (#732 by @larsoner)
+- Always generate CSP decdoing grand average analysis if CSP decoding was used on the single-subjec level (#733 by @hoechenberger)
```
larsoner, line 19:
```suggestion
- Always generate CSP decdoing grand average analysis if CSP decoding was used on the single-subject level (#733 by @hoechenberger)
```

larsoner:
Thanks in advance @hoechenberger !

PR #734, opened 2023-04-05
Provide a more helpful log message if the CSP decoding step is being skipped

### LINKED ISSUE #729: When CSP is skipped, we don't provide the correct log message
We **do** prepare a helpful log message but don't actually use it:
https://github.com/mne-tools/mne-bids-pipeline/blob/05b72e82d4c36f1a81bd6424a4301974cd6d7f3b/mne_bids_pipeline/steps/sensor/_05_decoding_csp.py#L541-L549

The bug is in line 547, which should simply be deleted

cc @SophieHerbst

hoechenberger:
Fixes #729

### Before merging …

### DIFF
--- a/docs/source/v1.3.md.inc
+++ b/docs/source/v1.3.md.inc
@@ -1,6 +1,8 @@
 ## v1.3.0 (unreleased)
 
-[//]: # (### :new: New features & enhancements)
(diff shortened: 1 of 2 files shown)


larsoner:
Thanks in advance @hoechenberger !

PR #736, opened 2023-04-06
MAINT: Dont use deprecated book API

### LINKED ISSUE #721: mne_bids_pipeline/_run.py:317: FutureWarning: Setting the `book` attribute is not part of the public API
I'm receiving a `FutureWarning`, probably related to Pandas? In any case, we should mask it – it currently appears in the output on the console, which is not nice.
```
mne_bids_pipeline/_run.py:317: FutureWarning: Setting the `book` attribute is not part of the public API, usage can give unexpected or corrupted results and will be removed in a future version
  writer.book = book
```

larsoner:
I think this preserves the original intent: append anything that's new (`mode='a'`) but replace the given sheet name if it exists. `if_sheet_exists` is available as of Pandas 1.3 (July 2021) so it seems fine to use.

### Before merging …

Closes #721

### DIFF
--- a/docs/source/v1.3.md.inc
+++ b/docs/source/v1.3.md.inc
@@ -10,9 +10,9 @@
 
 [//]: # (- Whatever (#000 by @whoever))
 
-[//]: # (### :medical_symbol: Code health)
+### :medical_symbol: Code health
 
-[//]: # (- Whatever (#000 by @whoever))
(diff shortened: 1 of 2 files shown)


hoechenberger (approved):
LGTM! Thanks for this simplification, Eric

PR #737, opened 2023-04-06
MAINT: Use rich instead of coloredlogs

### LINKED ISSUE #726: Switch from coloredlogs to rich
Coloredlogs seems pretty much dead – last commit almost 2 years ago:
https://github.com/xolox/python-coloredlogs

A well-maintained, extremely popular, and more versatile alternative is rich:
https://github.com/Textualize/rich

We should consider switching our logging system to rich

It can even render tracebacks beautifully:

Thoughts?

larsoner:
### Before merging …

For now I tried to keep the styling pretty similar. One immediate advantage is that `rich` autodetects filenames and makes them magenta, which is cool!

Follow-up PRs could use better box drawing, titling/horizontal ruling, etc. I'll leave that to you @hoechenberger :)

Closes #726


hoechenberger:
Thanks @larsoner!

PR #741, opened 2023-06-15
MAINT: Use a large resource class to speed up build

larsoner:
Interested in trying a `large` resource class since ERP_CORE_* (e.g. [ERN](https://app.circleci.com/pipelines/gh/mne-tools/mne-bids-pipeline/3731/workflows/37f200bd-f3ac-45c9-98bb-469faeb19aff/jobs/45831/resources)) appears to be CPU-bound with 2 CPUs. First I'm pushing an empty commit just to make sure CIs are green and replicate the ~19 minute timing. Then we'll try a large resource class to see if it helps!

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -306,6 +306,7 @@ jobs:
 
   test_ds000117:
     <<: *imageconfig
+    resource_class: large
     steps:
       - attach_workspace:
           at: ~/
@@ -434,6 +435,7 @@ jobs:
 
   test_ds000246:
     <<: *imageconfig
+    resource_class: large
     steps:
       - attach_workspace:
           at: ~/
@@ -501,6 +503,7 @@ jobs:
 
   test_ds000248_base:
     <<: *imageconfig
+    resource_class: large
     steps:
       - attach_workspace:
           at: ~/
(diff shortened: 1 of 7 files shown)


larsoner:
That would be a bit harder. We could run with `ulimit` maybe. Let's see if the speed gains are there first. Then we can decide if the tradeoff of increased memory limit is worth it.

Actually thinking about it the n_jobs to dask have a fixed limit of 2gb in their config so it should be safe enough in those jobs at least -- and those are the ERP CORE ones that will probably benefit most!

larsoner:
Okay that did take it from 35 minutes down to 30, which is nice. Looks like all the ERP_CORE got faster, and it should be memory-safe to increase the resource class here since (for parallel at least) we use a memory limit on the dask workers:

no-op commit | resource-class: large
--|--
Screenshot from 2023-06-15 13-33-08 | Screenshot from 2023-06-15 13-33-18

There wasn't much effect for the others (e.g., ds000248) though, so I'll revert those then merge.

larsoner:
... actually 1810 went from 12 to 8 min which seems worth it, too. So the effect is that just one of the tests (1810) is run no longer as memory limited, which seems worth it

Issue #574, opened 2022-07-28
ENH: Add movement compensation and related functions


larsoner:
Potential plan (feel free to comment @agramfort @hoechenberger @drammock):

- [ ] BIDS-ify dataset from https://github.com/ilabsbrainteam/2022-Best-Practices-Infant-MEG and upload to OpenNeuro
- [ ] Add test where this dataset is processed using the existing pipeline *without* movement compensation
- [ ] Add head position estimation + cHPI filtering + movement compensation capability
- [ ] Add annotation of bad segments
- [ ] Add movement and cHPI SNR plots to reports

This would more or less cover the pediatric portion of the EOSS4 grant I think.

larsoner:
For the first step I think I got the dataset converted, but the validator complained:
```
	1: [WARN] The Authors field of dataset_description.json should contain an array of fields - with one author per field. This was triggered based on the presence of only one author field. Please ignore if all contributors are already properly listed. (code: 102 - TOO_FEW_AUTHORS)
```
... which was good, because I had created a dataset with no authors (whoops!).

This prompted https://github.com/mne-tools/mne-bids/issues/1024. Once that's sorted I'll continue!

drammock:
xref to #1043 / #1055 which added TWA-hp. mf-bads (#760) is now the last remaining item before we can close this issue

larsoner:
Okay let's close this in favor of #760 then since it has more detailed info

PR #746, opened 2023-06-26
MAINT: DRY kwarg passing

larsoner:
### Before merging …

I went to work on #574 and the sheer volume of `kwargs` in `get_config` that must be used to `import_*_data` made it really hard to start from one of the `steps/preprocessing/*.py` files to add a new step. This PR is a no-op from the user standpoint but refactors to:

1. `_bids_kwargs` function that returns some standard BIDS kwargs like `cfg.space`. Eventually we should use this everywhere probably but I'm just starting with `steps/preprocessing` for now.
2. `_import_data_kwargs` function that gets the `config` params needed for reading raw data

Should be a lot more red than green here. I don't expect the first version to pass -- locally I've only run `ds004229` -- but hopefully it won't require many more iterations to get right!

### DIFF
--- a/docs/source/v1.4.md.inc
+++ b/docs/source/v1.4.md.inc
@@ -8,9 +8,9 @@
 
 [//]: # (- Whatever (#000 by @whoever))
 
-[//]: # (### :medical_symbol: Code health)
+### :medical_symbol: Code health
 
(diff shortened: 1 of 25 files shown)


hoechenberger (approved):
I love it! Great work

PR #747, opened 2023-06-27
ENH: Add movement compensation

larsoner:
### Before merging …

Adds basic movement compensation capability. Part of #574, which I've fleshed out into additional steps

The `if run == runs[0]:` conditional in `_get_raw_paths` that makes it add `noise` and `rest` runs really makes the logic here convoluted. In my next PR I'll try to split that out entirely so that it's cleaner. I thought about doing it here but I think it would make it harder to review, and it'll be nice to have the `ds004229`-with-MC case in place already.


larsoner:
Okay this finally works, here is the report (which is much better than `main` where the ERF just shows an artifact!):

- https://output.circle-artifacts.com/output/job/d0fc174f-aa0b-498b-8e7a-4b741d05465f/artifacts/0/site/examples/ds004229/sub-102_task-amnoise_report.html

Ready for review/merge from my end.

agramfort (approved):

larsoner:
Approved so I'll merge and keep moving!

PR #755, opened 2023-07-05
MAINT: Cleanup dependencies and doc build

### LINKED ISSUE #754: Make some dependencies explicit
We depend on

  - numpy
  - scipy
  - matplotlib
  - nibabel

(we import these packages explicitly) but they're not specified in `pyproject.toml`

larsoner:
### Before merging …

1. Take care of warnings reported in https://github.com/mne-tools/mne-bids-pipeline/issues/752#issuecomment-1619848546
2. Use `strict=True` to prevent future link regressions
3. Add template `vX.Y.md.inc` to use to facilitate next release (can `cp` more easily like in MNE-Python)
4. Specify versions of numpy, scipy, matplotlib, and nibabel in `pyproject.toml` with versions matching those of MNE-Python

Ready for review/merge from my end @hoechenberger 

Closes #754


larsoner:
> do we actually need to match the versions to MNE's?

In practice ours will probably be the same as MNE-Python's. So it seems like if we're going to add them we might as well have a pin in there. But I might be missing the point of having them in the first place, so if having un-pinned dependencies makes sense (e.g., for conda-forge ease) then that's okay with me

hoechenberger:
We include these departments because we make explicit use of these packages -- just to be in the clean side of things 

However the work we do in MNE-BIDS-Pipeline with these packages does not demand a specific version (schrammt there MIGHT be lower bounds but I'm not aware of any at this time)

PR #757, opened 2023-07-05
ENH: Add annotation of head velocity

larsoner:
### Before merging …

With updated vals for `ds004229`:
```
mf_mc_rotation_velocity_limit = 30.0  # deg/s for annotations
mf_mc_translation_velocity_limit = 20e-3  # m/s
```
During the run we see:
```
[14:47:31] │ ⏳️ preprocessing/_03_maxfilter sub-102 Translation velocity exceeded 0.02 m/s limit for 5.5 s (1.8%)
[14:47:31] │ ⏳️ preprocessing/_03_maxfilter sub-102 Rotation velocity exceeded 30.0 °/s limit for 4.0 s (1.3%)
```
And in the report we get:

Screenshot from 2023-07-05 14-49-41

### DIFF
--- a/docs/source/v1.5.md.inc
+++ b/docs/source/v1.5.md.inc
@@ -2,7 +2,7 @@
 
 ### :new: New features & enhancements
 
-- Added support for annotating bad segments based on for head movement velocity (#757 by @larsoner)
+- Added support for annotating bad segments based on head movement velocity (#757 by @larsoner)
 
 [//]: # (### :warning: Behavior changes)
 
--- a/docs/source/v1.5.md.inc
+++ b/docs/source/v1.5.md.inc
@@ -1,8 +1,8 @@
 ## v1.5.0 (unreleased)
 
-[//]: # (### :new: New features & enhancements)
+### :new: New features & enhancements
 
-[//]: # (- Whatever (#000 by @whoever))
+- Added support for annotating bad segments based on for head movement velocity (#757 by @larsoner)
 
 [//]: # (### :warning: Behavior changes)
 
--- a/mne_bids_pipeline/_config.py
+++ b/mne_bids_pipeline/_config.py
@@ -707,6 +707,18 @@ mf_mc_dist_limit: float = 0.005
 Minimum distance (m) to accept for cHPI position fitting.
 """
 
+mf_mc_rotation_velocity_limit: Optional[float] = None
+"""
+The rotation velocity limit (degrees/second) to use when annotating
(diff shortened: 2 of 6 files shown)


docs/source/v1.5.md.inc
```diff
@@ -1,8 +1,8 @@
 ## v1.5.0 (unreleased)
 
-[//]: # (### :new: New features & enhancements)
+### :new: New features & enhancements
 
-[//]: # (- Whatever (#000 by @whoever))
+- Added support for annotating bad segments based on for head movement velocity (#757 by @larsoner)
```
hoechenberger, line 5: There's something wrong with this sentence :)

mne_bids_pipeline/_config.py
```diff
@@ -707,6 +707,18 @@
 Minimum distance (m) to accept for cHPI position fitting.
 """
 
+mf_mc_rotation_velocity_limit: Optional[float] = None
+"""
+The rotation velocity limit (degrees/second) to use when annotating
+movement-compensated data. If None, no annotations will be added.
```
hoechenberger, line 713:
```suggestion
movement-compensated data. If `None`, no annotations will be added.
```

mne_bids_pipeline/_config.py
```diff
@@ -707,6 +707,18 @@
...
+"""
+
+mf_mc_translation_velocity_limit: Optional[float] = None
+"""
+The translation velocity limit (meters/second) to use when annotating
+movement-compensated data. If None, no annotations will be added.
```
hoechenberger, line 719:
```suggestion
movement-compensated data. If `None`, no annotations will be added.
```

hoechenberger:
@larsoner I'm wondering if we shouldn't add this upstream to `mne.Report`? I got this impression when I saw you injecting custom HTML. – It's an open question, I'm not entirely sure about the best approach myself. WDYT?

agramfort (approved):

larsoner:
Indeed in principle people could want to automatically gain some notes about `raw` annotations. In principle we could maybe add some logic to `report.add_raw` to autodetect the built-in annot types we have, but for example it would miss the limit that I inject here. Or if users have done their own annotations they'll have to write custom code anyway. So there is a deeper API discussion to be had. My inclination is to add stuff here first and if people think it's useful generally then we can add it upstream and remove our "custom" code here

hoechenberger:
Thanks @larsoner

PR #764, opened 2023-07-17
ENH: More compact logging

larsoner:
### Before merging …

@hoechenberger feel free to review/merge if you're happy.

I have been bothered by too much redundancy and overuse of horizontal space in our logging recently. Now that we use `rich` we can make better use of `rule` to avoid redundancy. In particular, currently we print the name of the step at each logging line and it takes up almost a third of the horizontal space (here shown in 132x43 terminal, the largest in the Ubuntu Linux default Terminal drop-down):

Screenshot from 2023-07-17 13-35-08

With `.rule` we can cleanly log the start of each step as a title and link everything using a consistent color instead (green):

Screenshot from 2023-07-17 14-17-41

So I think overall this becomes both easier to follow and more compact / less redundant.

Also adds `[link]` support to make the Report HTML Files clickable.

### DIFF
--- a/mne_bids_pipeline/_main.py
+++ b/mne_bids_pipeline/_main.py
@@ -193,22 +193,19 @@ def main():
         # them twice.
         step_modules = [*STEP_MODULES["init"], *step_modules]
 
-    msg = "Welcome aboard the MNE BIDS Pipeline!"
-    logger.info(**gen_log_kwargs(message=msg, emoji="👋", box="╶╴", step=""))
+    logger.title("👋 Welcome aboard the MNE BIDS Pipeline! 👋")
     msg = f"Using configuration: {config}"
-    logger.info(**gen_log_kwargs(message=msg, emoji="🧾", box="╶╴", step=""))
+    logger.info(**gen_log_kwargs(message=msg, emoji="🧾"))
+    logger.end()
 
     config_imported = _import_config(
         config_path=config_path,
         overrides=overrides,
     )
-    for si, step_module in enumerate(step_modules):
+    for step_module in step_modules:
         start = time.time()
         step = _short_step_path(pathlib.Path(step_module.__file__))
-        if si == 0:
-            logger.rule()
-        msg = "Now running  👇"
-        logger.info(**gen_log_kwargs(message=msg, box="┌╴", emoji="🚀", step=step))
+        logger.title(title=f"🚀 {step}")
         step_module.main(config=config_imported)
         elapsed = time.time() - start
         hours, remainder = divmod(elapsed, 3600)
@@ -221,6 +218,4 @@ def main():
             elapsed = f"{minutes}m {elapsed}"
         if hours:
             elapsed = f"{hours}h {elapsed}"
-        msg = f"Done running  👆 [{elapsed}]"
-        logger.info(**gen_log_kwargs(message=msg, box="└╴", emoji="🎉", step=step))
-        logger.rule()
+        logger.end(f"[prefix]🎉 done[/] {elapsed}")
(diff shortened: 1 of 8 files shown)


agramfort (approved):

larsoner:
The horizonal dividers (rules) are still there. Can you clarify what you mean?

larsoner:
... you mean without the title in the rule itself? Yes we can do that instead of having the title in the rule. It's a little less compact vertically but it'll work. I'll try that and attach a screenshot to see if it's more what you're looking for

larsoner:
1. Current solution:

   Screenshot 2023-07-18 at 7 13 49 AM

2. Remove emoji from title in rule:

   Screenshot 2023-07-18 at 7 16 42 AM

3. Rule all the way across:

   Screenshot 2023-07-18 at 7 25 18 AM

I prefer (1) or (2) over (3), but if you feel strongly about it I can live with (3). Or maybe @drammock can weigh in and decide!

drammock:
Of the options shown I prefer 2. I would further do the same thing with "done" lines that was done in titles (remove emoji and set "done (58 s)" in green within a rule. And (no surprise here) I'd remove all emojis (what's the value / added information of seeing a stack of hourglass emojis?). My 2 cents

larsoner:
> And (no surprise here) I'd remove all emojis (what's the value / added information of seeing a stack of hourglass emojis?).

In typical (re-)runs they do add information and are not always hourglasses (see below)

>  I would further do the same thing with "done" lines that was done in titles (remove emoji and set "done (58 s)" in green within a rule. 

This part I'm not a big fan of because I think it's a bit distracting having what ends up being double separated lines (not 100% to spec but you get the idea):

Screenshot 2023-07-18 at 8 02 47 AM

But a varaint of this would be to draw "over to" the `done` line using a box char which helps end the section (now my fav I think):

Screenshot 2023-07-18 at 8 08 32 AM

To me it's very easy to see where each section begins and ends, doesn't waste space, etc. But I can live with the double-bar separation if people want.

mne_bids_pipeline/_main.py
```diff
@@ -193,22 +193,19 @@ def main():
         # them twice.
         step_modules = [*STEP_MODULES["init"], *step_modules]
 
-    msg = "Welcome aboard the MNE BIDS Pipeline!"
-    logger.info(**gen_log_kwargs(message=msg, emoji="👋", box="╶╴", step=""))
+    logger.title("Welcome aboard the MNE BIDS Pipeline! 👋")
```
hoechenberger, line 196: we already dropped the article everywhere else:
```suggestion
    logger.title("Welcome aboard MNE-BIDS-Pipeline! 👋")
```

hoechenberger:
In that case, could we leave the hourglasses for the time being? I'd like to revisit this later after doing some research regarding what's possible with `rich`. I believe it does have support for "parallel logging", but I've never used it. (But I might be mistaken)

larsoner:
> In that case, could we leave the hourglasses for the time being?

Yes and to me there is information that might justify keeping it even with live updating. Differnt step *execution types* -- skipped, cached, cached-but-forced-to-rerun, running, etc. -- have different icons so in reruns you can quickly see what recomputed. Maybe at some point clear enough text can replace this but in the meantime the emojis make it easy to group steps / see what was run versus skipped/cached quickly.

larsoner:
(but @drammock let me know if you really want the double-separator and I can unmark, or we can do it in a follow-up PR. Or @drammock could be a good PR for you, too -- hopefully it's easy enough to see how to tweak stuff given this PR :) )

drammock:
the double-line all the way across is not necessary.  I'm happy with the line connecting to "done" and the "done" being green.

PR #768, opened 2023-07-19
MAINT: Test MD5 hashing too

larsoner:
### Before merging …

Looking at coverage I noticed that our MD5 hashing isn't used anywhere. This should increase our coverage a bit and make sure it actually works (including on re-run).


larsoner:
Trivial change so I'll self-merge!

PR #779, opened 2023-08-14
ENH: Validate inputs with pydantic

### LINKED ISSUE #766: MAINT: Validate inputs
Pydantic would then be a second step, where we can actually validate the input

_Originally posted by @hoechenberger in https://github.com/mne-tools/mne-bids-pipeline/issues/646#issuecomment-1309260129_

In other words, we should use `pydantic` (maybe) to validate inputs to the extent possible. Ideally we could get rid of `_validate_type` and `_check_option` usage, too.

larsoner:
### Before merging …

- [x] Validate inputs with `pydantic` after adding as a dependency (2.0 seems to be the min useable version for us based on local testing)
- [x] Fix bugs where defaults for `decoding_csp_times` and `decoding_csp_freqs` didn't do what we said they did (they were set once statically based on the defaults rather than being computed dynamically based on the variable values)

Locally it seems to work -- adding `config_validation = "foo"` to `config_ds003392.py` for example produces:
```
$ pytest mne_bids_pipeline -xk 003392
...
mne_bids_pipeline/_config_import.py:410: in _pydantic_validate
    raise ValueError(str(err)) from None
E   ValueError: 1 validation error for user configuration from /home/larsoner/python/mne-bids-pipeline/mne_bids_pipeline/tests/configs/config_ds003392.py
E   config_validation
E     Input should be 'raise', 'warn' or 'ignore' [type=literal_error, input_value='foo', input_type=str]
E       For further information visit https://errors.pydantic.dev/2.1/v/literal_error
```
or when setting `mf_destination = "foo"` in `config_ds003392.py` which allows a specific Literal or a FloatArrayLike:
```
$ pytest mne_bids_pipeline -xk 004229
...
mne_bids_pipeline/_config_import.py:410: in _pydantic_validate
    raise ValueError(str(err)) from None
E   ValueError: 2 validation errors for user configuration from /home/larsoner/python/mne-bids-pipeline/mne_bids_pipeline/tests/configs/config_ds004229.py
E   mf_destination.literal['reference_run']
E     Input should be 'reference_run' [type=literal_error, input_value='foo', input_type=str]
E       For further information visit https://errors.pydantic.dev/2.1/v/literal_error
E   mf_destination.function-plain[assert_float_array_like()]
E     Value error, could not convert string to float: 'foo' [type=value_error, input_value='foo', input_type=str]
E       For further information visit https://errors.pydantic.dev/2.1/v/value_error
```
Because of the `strict=True` I suspect there will be some `List`s that we will want to change to `FloatArrayLike` (or create a generic `CheckedArrayLike` that allows arbitrary dtypes or something) but this is maybe a good enough start!

Closes #766

### DIFF
--- a/mne_bids_pipeline/tests/configs/config_ERP_CORE.py
+++ b/mne_bids_pipeline/tests/configs/config_ERP_CORE.py
@@ -24,7 +24,6 @@ components from a total of 6 experimental tasks:
                 [https://doi.org/10.1016/j.neuroimage.2020.117465](https://doi.org/10.1016/j.neuroimage.2020.117465)
 """
 import argparse
-import mne
 import sys
 
 study_name = "ERP-CORE"
@@ -50,7 +49,7 @@ raw_resample_sfreq = 128
 # Suppress "Data file name in EEG.data (sub-019_task-ERN_eeg.fdt) is incorrect..."
 read_raw_bids_verbose = "error"
 
-eeg_template_montage = mne.channels.make_standard_montage("standard_1005")
+eeg_template_montage = "standard_1005"
 eeg_bipolar_channels = {
     "HEOG": ("HEOG_left", "HEOG_right"),
     "VEOG": ("VEOG_lower", "FP2"),
--- a/mne_bids_pipeline/tests/configs/config_ERP_CORE.py
+++ b/mne_bids_pipeline/tests/configs/config_ERP_CORE.py
@@ -24,6 +24,7 @@ components from a total of 6 experimental tasks:
                 [https://doi.org/10.1016/j.neuroimage.2020.117465](https://doi.org/10.1016/j.neuroimage.2020.117465)
 """
 import argparse
+import mne
 import sys
 
 study_name = "ERP-CORE"
@@ -49,7 +50,7 @@ raw_resample_sfreq = 128
 # Suppress "Data file name in EEG.data (sub-019_task-ERN_eeg.fdt) is incorrect..."
 read_raw_bids_verbose = "error"
 
-eeg_template_montage = "standard_1005"
+eeg_template_montage = mne.channels.make_standard_montage("standard_1005")
 eeg_bipolar_channels = {
     "HEOG": ("HEOG_left", "HEOG_right"),
     "VEOG": ("VEOG_lower", "FP2"),
--- a/docs/source/v1.5.md.inc
+++ b/docs/source/v1.5.md.inc
@@ -9,6 +9,7 @@
 - Output logging spacing improved (#764 by @larsoner)
 - Added caching of sensor and source average steps (#765 by @larsoner)
 - Improved logging of coregistration distances (#769 by @larsoner)
+- Input validation has been improved by leveraging [pydantic](https://docs.pydantic.dev) (#779 by @larsoner)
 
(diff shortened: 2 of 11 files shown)


larsoner:
Already found either a usage error or documentation / type-hinting error:
```
mne_bids_pipeline/_config_import.py:411: in _pydantic_validate
    raise ValueError(str(err)) from None
E   ValueError: 1 validation error for user configuration from /home/circleci/project/mne_bids_pipeline/tests/configs/config_ERP_CORE.py
E   eeg_template_montage
E     Input should be a valid string [type=string_type, input_value=<DigMontage | 0 extras (h...fiducials, 343 channels>, input_type=DigMontage]
E       For further information visit https://errors.pydantic.dev/2.1/v/string_type
```
Changing to just be `eeg_template_montage = "standard_1005"` but maybe someday we want to extend support to arbitrary DigMontage...

mne_bids_pipeline/tests/configs/config_ERP_CORE.py
```diff
@@ -50,7 +49,7 @@
 # Suppress "Data file name in EEG.data (sub-019_task-ERN_eeg.fdt) is incorrect..."
 read_raw_bids_verbose = "error"
 
-eeg_template_montage = mne.channels.make_standard_montage("standard_1005")
+eeg_template_montage = "standard_1005"
```
hoechenberger, line 52: This change would disable testing with a `DigMontage`.

Instead of doing this, I'd suggest to fix the type annotation in our config to accept `str | mne.channels.DigMontage`

hoechenberger:
Very nice otherwise!!

mne_bids_pipeline/tests/configs/config_ERP_CORE.py
larsoner: > This change would disable testing with a DigMontage.

This was never officially documented though. Do we really need to / want to support it?

larsoner: see for example:

https://mne.tools/mne-bids-pipeline/1.4/settings/general.html#mne_bids_pipeline._config.eeg_template_montage

agramfort (commented):
Nice !

mne_bids_pipeline/tests/configs/config_ERP_CORE.py
hoechenberger: I would almost assume that we added this test for a reason ... @agramfort @dengemann Do you know if and why we required setting a montage from a DigMontage instance instead of just passing the template montage name as a string? Was this something you needed at some point?

larsoner: Ahh, doing a bit of `git` digging I can see https://github.com/mne-tools/mne-bids-pipeline/pull/407 by @dengemann so it was indeed intentional... I'll add support back in

dengemann:
I don’t remember. Probably it fixed something with some version of MNE. You
can remove it or keep it and we see what happens :)

mne_bids_pipeline/tests/configs/config_ERP_CORE.py
hoechenberger: Great, thanks!

PR #809, opened 2023-11-03
ENH: Table sorting and ds004229 updates

larsoner:
- [x] Update config for ds004229 v1.0.3
- [x] Add [table sorting](https://squidfunk.github.io/mkdocs-material/reference/data-tables/#sortable-tables) so if you want to find all examples that for example use ICA you can sort by the ICA column and sort it by checked/unchecked.

Someday I dream of having a more compact table layout and sticky first row/col but after some time trying several CSS solutions I was unable to do so!


docs/mkdocs.yml
```diff
@@ -65,6 +65,10 @@ copyright: Copyright &copy; MNE-BIDS-Pipeline authors
 
 extra_css:
     - css/extra.css
+# https://squidfunk.github.io/mkdocs-material/reference/data-tables/
+extra_javascript:
+  - https://unpkg.com/tablesort@5.3.0/dist/tablesort.min.js
```
hoechenberger, line 70: We should vendor this file instead of pulling it from the CDN

hoechenberger:
> Someday I dream of having a more compact table layout and sticky first row/col but after some time trying several CSS solutions I was unable to do so!

@larsoner The tables we use for metadata are quite nice IMHO

https://output.circle-artifacts.com/output/job/ae9d8f27-6bde-4b79-b1ba-6f020efbc2ea/artifacts/0/reports/ERP_CORE/sub-015_ses-MMN_task-MMN_report.html#Epochs__before_cleaning-Metadata

docs/mkdocs.yml
hoechenberger: Then again, I think we're not vendoring our emojis either, so …

larsoner: Let's wait to vendor until later, then. Maybe in the end we don't need to?

larsoner:
Nicely sortable now

https://output.circle-artifacts.com/output/job/cea69984-fa63-49ea-b485-fc15c9c90724/artifacts/0/site/examples/examples.html#demonstrated-features

hoechenberger:
Nice! But I do t get any visual indicator of this functionality until I actually start clicking on the header cells. So this feature is not easily discoverable 🤔

agramfort (approved):

hoechenberger:
> Nice! But I don't get any visual indicator of this functionality until I actually start clicking on the header cells. So this feature is not easily discoverable 🤔

Ah, it happens on hover! Fair enough

PR #810, opened 2023-11-06
Add support for `autoreject` (local) before running ICA

hoechenberger:
### Before merging …


mne_bids_pipeline/steps/preprocessing/_06a_run_ica.py
```diff
@@ -390,33 +391,50 @@ def run_ica(
...
+        msg = "Using autoreject to find and repair bad epochs"
+        logger.info(**gen_log_kwargs(message=msg))
 
-    msg = f"Using PTP rejection thresholds: {ica_reject}"
-    logger.info(**gen_log_kwargs(message=msg))
+        ar = autoreject.AutoReject(
+            n_interpolate=cfg.autoreject_n_interpolate,
+            random_state=cfg.random_state,
+            n_jobs=exec_params.n_jobs,
+        )
```
hoechenberger:
```suggestion
        ar = autoreject.AutoReject(
            n_interpolate=cfg.autoreject_n_interpolate,
            random_state=cfg.random_state,
            n_jobs=exec_params.n_jobs,
            verbose=False,
        )
```

hoechenberger:
Report:

https://output.circle-artifacts.com/output/job/586ce7bb-89b6-466c-84ab-23277c1ce14c/artifacts/0/reports/ERP_CORE/sub-015_ses-ERN_task-ERN_proc-ica+components_report.html#Epochs__Autoreject_cleaning


Tests are still taking too long, will look into this before we merge

mne_bids_pipeline/tests/configs/config_ERP_CORE.py
```diff
@@ -75,17 +75,22 @@
...
 else:
+    spatial_filter = "ica"
     ica_reject = dict(eeg=350e-6, eog=500e-6)
     reject = "autoreject_global"
 
-    spatial_filter = "ica"
-    ica_max_iterations = 1000
-    ica_eog_threshold = 2
-    ica_decim = 2  # speed up ICA fitting
+# These settings are only used for the cases where spatial_filter is not None
```
hoechenberger, line 88:
```suggestion
# These settings are only used for the cases where spatial_filter="ica"
```

larsoner:
Do you need both `local` and `global` options at both stages (before and after ICA), or can we get away with / is it advisable to use one or the other at one stage vs the other? Even if it's easy to add both options both places, if it's unlikely (or typically unwise) to use one or the other then it might be better to omit it for now and add it only when someone needs it.

Is there any autoreject doc / tutorial we can link to in order to help people decide?

hoechenberger:
> Do you need both `local` and `global` options at both stages (before and after ICA), or can we get away with / is it advisable to use one or the other at one stage vs the other? Even if it's easy to add both options both places, if it's unlikely (or typically unwise) to use one or the other then it might be better to omit it for now and add it only when someone needs it.
> 
> Is there any autoreject doc / tutorial we can link to in order to help people decide?

For ICA, the recommended way is using AR local before ICA (on the epochs that will be used for fitting), then clean the data with ICA, and running AR local on those cleaned epochs

I think we didn't add AR global for pre-ICA cleaning in the past because it would reject just too many epochs. I could remove that option

hoechenberger:
Official docs on AR & ICA:
https://autoreject.github.io/stable/auto_examples/plot_autoreject_workflow.html

hoechenberger:
> > Do you need both `local` and `global` options at both stages (before and after ICA), or can we get away with / is it advisable to use one or the other at one stage vs the other? Even if it's easy to add both options both places, if it's unlikely (or typically unwise) to use one or the other then it might be better to omit it for now and add it only when someone needs it.
> > Is there any autoreject doc / tutorial we can link to in order to help people decide?
> 
> For ICA, the recommended way is using AR local before ICA (on the epochs that will be used for fitting), then clean the data with ICA, and running AR local on those cleaned epochs
> 
> I think we didn't add AR global for pre-ICA cleaning in the past because it would reject just too many epochs. I could remove that option

That said, I've used the existing AR global rejection after ICA successfully for a long time. So we should offer

```
ica_reject: None | dict[str, float] | Literal["autoreject_local"]
reject: None | dict[str, float] | Literal["autoreject_local", "autoreject_global"]
```
WDYT?

hoechenberger:
@agramfort What's your feeling here? We didn't support AR global before ICA in the past. Maybe it should stay that way?

But AR local before ICA makes total sense, I'm getting really nice results and much better decompositions than when using fixed global thresholds with my data

hoechenberger:
@larsoner I've removed support for global autoreject before ICA.

larsoner:
Nice @hoechenberger !

hoechenberger:
Thanks @larsoner

PR #816, opened 2023-11-13
Do not interpolate bad segments using autoreject before ICA

hoechenberger:
This is a fixup for #810

There, not only did we remove bad epochs, but also interpolated data before submission to ICA fitting. However, this is not what is recommended in the autoreject documentation: there, the suggestion is to [only remove the bad epochs](https://autoreject.github.io/stable/auto_examples/plot_autoreject_workflow.html#ica), not interpolate.

This PR fixes the behavior accordingly and updates the documentation and log messages.

I tested on a set of noise EEG data and I'm actually getting better results with this approach for the majority participants .

### Before merging …


larsoner:
Makes sense, thanks @hoechenberger

PR #817, opened 2023-11-13
Omit bad channels when gathering data used for decoding

hoechenberger:
Addresses https://github.com/mne-tools/mne-python/issues/12197

### Before merging …


larsoner (approved):

larsoner:
I'm going to remove the merge req that branches be up to date to reduce CI burden. It should be rare that PRs are incompatible and when it does happen `main` will fail quickly for us

PR #825, opened 2023-12-08
Switch build backend to hatchling

hoechenberger:
### Before merging …

cc @ofek in case you're interested :)


pyproject.toml
```diff
@@ -83,19 +84,21 @@ homepage = "https://mne.tools/mne-bids-pipeline"
...
+raw-options = { version_scheme = "release-branch-semver" }
 
-[tool.setuptools.package-data]
-"mne_bids_pipeline.steps.freesurfer.contrib" = ["version"]
+[tool.hatch.build]
+exclude = [
+    "/.*",
+    "/codecov.yml",
+    "**/tests",
+    "/docs",
```
hoechenberger, line 96: @ofek I'm explicitly excluding `/docs` here, but `docs/source/examples/gen_examples.py` is still being packaged when I build the sdist. Any idea why that's happening?

hoechenberger: Ah! There's an entry in `.gitignore`:

```
!docs/source/examples/gen_examples.py
```

Wonder why we have that one … but this explains it.

hoechenberger:
@larsoner i think this is good to go

larsoner:
Nice @hoechenberger !

PR #836, opened 2024-01-24
MAINT: Test result caching

### LINKED ISSUE #814: Caching seems not to be working on macOS
_I cannot provide sufficient information now – this issue is supposed to serve as a reminder._

I'm under the impression that caching is not working on my macOS machine – it seems that all steps are always run again, regardless of the caching setting. I also tried the `"hash"` method, but I'm not seeing any change. Needs to be investigated.

larsoner:
### Before merging …

Closes #814

Let's see if we can replicate the issue! *EDIT: Spoiler alert, we cannot.*


larsoner:
@hoechenberger this appears headed for green. I was unable to replicate the caching issue on macOS, both datasets tested here appear to process in < 15s the second time. Ready for review/merge from my end. Happy to adjust CI requirements if you're happy (I folded the Style check into the doc consistency one since the only step we really needed there now was `codespell`, and knew we'd be adjusting requirements here anyway!)

hoechenberger:
Thanks! I also realized I cannot reliably reproduce on macOS. Sometimes the problem occurs and sometimes it doesn't. I usually work inside of a OneDrive folder, maybe this has implications...

larsoner:
> I usually work inside of a OneDrive folder, maybe this has implications...

It seems like it could for `mtime` but for `hash` it would be weird. It really seems like the file hash shouldn't change!

hoechenberger (approved):

Issue #831, opened 2024-01-16
ENH: Preprocessed raw data


larsoner:
Ideally MNE-BIDS-Pipeline would be usable for all possible analyses. However, currently (and for the foreseeable future) it's used for event-based designs. For things like naturalistic speech or resting state analyses (or other types of analyses probably, too) it would be useful to have not just preprocessed *epochs* but also preprocessed *raw* data available for downstream analyses not yet implemented. So assuming we want to add it, we have two options I think:

1. Always preprocess the raw data the same way we do for epochs, and save the resulting files
2. Make saving preprocessed raw data optional / opt-in

I'm inclined toward (1) but it could be wasteful for some people in terms of disk usage. Thoughts?

larsoner:
... in other words, after creating `sub-01_task-whatever_run-01_proc-filt_raw.fif` create something like a `sub-01_task-whatever_run-01_proc-clean_raw.fif` or similar

PR #837, opened 2024-01-25
ENH: Add artifact regression

### LINKED ISSUE #832: ENH: TSPCA or regress_artifact
We currently don't offer any (?) reference-sensor-based preprocessing. meegkit has [TSPCA](https://nbara.github.io/python-meegkit/modules/meegkit.tspca.html) for this. I haven't used it much but some collaborators with a KIT system have had good results. Any objection to me adding it as a step?

As part of this, I'd need to find a dataset that has reference sensors... not sure if we have any so far.

larsoner:
### Before merging …

Adds `config.regress_artifact`, used in my pipeline to regress reference sensor signals:
```
regress_artifact = dict(picks="meg", picks_artifact=["MISC 001", "MISC 002", "MISC 003"])
```
Draft because I need to find a good KIT dataset to process to show this stuff, any ideas @agramfort ? I saw [MASC-MEG](https://osf.io/ag3kj/) but it's continuous speech (which we probably don't handle well) and huge.

I might try converting https://mne.tools/dev/auto_tutorials/inverse/95_phantom_KIT.html to BIDS since it works to demonstrate the processing.

Closes #832


mne_bids_pipeline/_import_data.py
```diff
@@ -449,7 +449,6 @@ def import_er_data(
             cfg=cfg,
             bids_path_bads=bids_path_er_bads_in,
         )
-    raw_er.pick("meg", exclude=[])
```
larsoner, line 452: For some reason we were only picking MEG channels in empty room data. To me this step seems unnecessary, and also breaks artifact regression, so I think we should drop it.

mne_bids_pipeline/_report.py
```diff
@@ -69,14 +69,13 @@ def _open_report(
             yield report
         finally:
             try:
-                msg = "Adding config and sys info to report"
-                logger.info(**gen_log_kwargs(message=msg))
```
larsoner, line 73: Moved this later in the file where we can make sure we emit it exactly once per `mne_bids_pipeline` call (other calls should be fast and it's not helpful to have dozens of these messages in the logs)

mne_bids_pipeline/steps/preprocessing/_04_frequency_filter.py
```diff
@@ -190,6 +193,8 @@ def filter_data(
 
     out_files[in_key] = bids_path_in.copy().update(
         root=cfg.deriv_root,
+        subject=subject,  # save under subject's directory so all files are there
+        session=session,
```
larsoner, line 197: Previously the `filt` step for SSS processing would output to `sub-01/*` type paths, whereas if SSS was off it would output to `sub-emptyroom/*` type paths. This unifies it so everything lands in subject-specific paths. This means that after the `filt` step (e.g., during regression, SSP, and ICA) we can assume all files for each subject come from that subject's path rather than potentially `sub-emptyroom`. This may mean a few redundant files but simplifies logic and should help with #831.

mne_bids_pipeline/steps/preprocessing/_04_frequency_filter.py
```diff
@@ -199,6 +204,18 @@ def filter_data(
         check=False,
     )
 
+    if cfg.regress_artifact is None:
```
larsoner, line 207: Need to filter the data and regression channels the same way

mne_bids_pipeline/steps/preprocessing/__init__.py
```diff
@@ -4,22 +4,24 @@
...
+from . import _08b_apply_ssp
+from . import _09_ptp_reject
 
 _STEPS = (
     _01_data_quality,
     _02_head_pos,
     _03_maxfilter,
     _04_frequency_filter,
-    _05_make_epochs,
+    _05_regress_artifact,
```
larsoner, line 20: Previously the order was SSS, filt, epoch, SSP/ICA, clean epochs. Now the order is SSS, filt, SSP/ICA, epoch, clean epochs. This makes it clearer that the epochs *are not* used for SSP and ICA, which is helpful in terms of understanding the flow in that we basically stay working with raw data ountil step _07 now.

larsoner (commented):
Leaving some inline comments to facilitate eventual review

mne_bids_pipeline/_config.py
```diff
@@ -202,7 +213,7 @@
 covariance (via `noise_cov='rest'`).
 """
 
-ch_types: Iterable[Literal["meg", "mag", "grad", "eeg"]] = []
+ch_types: Annotated[Sequence[Literal["meg", "mag", "grad", "eeg"]], Len(1, 4)] = []
```
larsoner, line 216: I errantly had `ch_types = "meg"`. By changing this from `Iterable` to `Sequence`, I got a nice error message saying it couldn't just be a string. Then we can use `pydantic` to further validate the length such as non-emptiness using `Len` from `annotated_types`, which is required by `pydantic`. So the deletions in `_config_import.py` below are now handled by pydantic for us!

.circleci/config.yml
```diff
@@ -48,7 +48,7 @@ jobs:
           name: Get Python running
           command: |
             pip install --upgrade --progress-bar off pip
-            pip install --upgrade --progress-bar off "autoreject @ https://api.github.com/repos/autoreject/autoreject/zipball/master" "mne[hdf5] @ git+https://github.com/mne-tools/mne-python@main" "mne-bids[full] @ https://api.github.com/repos/mne-tools/mne-bids/zipball/main" numba
+            pip install --upgrade --progress-bar off "autoreject @ https://api.github.com/repos/autoreject/autoreject/zipball/master" "mne[hdf5] @ git+https://github.com/larsoner/mne-python@regress" "mne-bids[full] @ https://api.github.com/repos/mne-tools/mne-bids/zipball/main" numba
```
larsoner, line 51: Temporarily use my branch until https://github.com/mne-tools/mne-python/pull/12389 lands

larsoner:
... @hoechenberger I think this is ready for review. If you're happy I'll wait until https://github.com/mne-tools/mne-python/pull/12389 lands, revert the change to use my branch, then merge

larsoner:
Seems to work, nice low-freq noise suppressed by regressing the ref sensors!

https://output.circle-artifacts.com/output/job/d92c5960-9b7e-4cb2-ad98-f9586ccb1092/artifacts/0/site/examples/MNE-phantom-KIT-data/sub-01_task-phantom_report.html#Epochs__before_cleaning

I'll work on fixing the conflicts then should be good to go

agramfort (approved):
LGTM

hoechenberger:
Thanks @larsoner!

PR #838, opened 2024-01-26
MAINT: Enable more ruff rules

larsoner:
### Before merging …

Get our formatting rules up to date with MNE-Python. We should merge then add to `.git-blame-revs-ignore` (can do in #837 since that will probably take longer to get merged).

### DIFF
--- a/mne_bids_pipeline/_config.py
+++ b/mne_bids_pipeline/_config.py
@@ -1,7 +1,6 @@
 # Default settings for data processing and analysis.
 
-from collections.abc import Iterable
-from typing import Callable, Literal, Optional, Union
+from typing import Callable, Iterable, Literal, Optional, Union
 
 from mne import Covariance
 from mne_bids import BIDSPath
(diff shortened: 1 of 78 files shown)


mne_bids_pipeline/_config.py
```diff
@@ -101,7 +100,7 @@
 BIDS dataset.
 """
 
-exclude_runs: Optional[Dict[str, List[str]]] = None
+exclude_runs: Optional[dict[str, list[str]]] = None
```
hoechenberger, line 103: I think this is only supported in Python 3.10+

larsoner: Looks like 3.9+ I think which we have in our reqs?

https://docs.python.org/3.9/library/typing.html#type-aliases

larsoner: I think I found the PEP that says 3.9 https://peps.python.org/pep-0585/

hoechenberger: Oh, great! 😊 all good then!

PR #839, opened 2024-01-30
BUG: Fix several bugs

larsoner:
### Before merging …

I went to work on #831 but wanted to sort out a few little other bugs (and one minor enhancement) first:

1. Fix bug where using `mf_reference_run = None` could change the empty room file matched from run-to-run of `mne_bids_pipeline` due to use of `set`, whose order is not guaranteed. @hoechenberger I suspect this could be the actual root cause of #814 
2. Output the `reject` params in the reject step -- though this can be set in the config, it's sometimes calculated on-the-fly by `autoreject` and not saved anywhere. Even in the manual `dict` case it's nice to have it near the drop logs in the report in case you want to think about adjusting your thresholds:

   Screenshot from 2024-01-30 12-00-44

4. Fix bug where `reject` was getting restricted incorrectly for M/EEG channels
5. Prefer `stat` to `lstat` because I'm pretty sure we *don't* want to use the symlink's own `mtime` but rather the `mtime` of whatever it points to. (This is apparently the only difference between `lstat` and `stat`.) Doubt anyone has actually hit this so no changelog needed, but better to have it in there I think.
6. Fix bug with `-regress` saving where instead of writing one for each run, a single filename was written to by every run (within-release change so no changelog needed).
7. Slightly better report tags for the cleaned epochs to make it clear that what's happened is the `reject` step

### DIFF
--- a/docs/source/v1.6.md.inc
+++ b/docs/source/v1.6.md.inc
@@ -20,6 +20,7 @@
 - Fix minor issues with path handling for cross-talk and calibration files (#834 by @larsoner)
 - Fix bug where EEG `reject` params were not used for `ch_types = ["meg", "eeg"]` (#839 by @larsoner)
 - Fix bug where implicit `mf_reference_run` could change across invocations of `mne_bids_pipeline`, breaking caching (#839 by @larsoner)
+- Fix bug where `--no-cache` had no effect (#839 by @larsoner)
 
 ### :medical_symbol: Code health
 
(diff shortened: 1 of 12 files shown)


hoechenberger:
Thanks @larsoner!

I think the value of the "reject" parameter should be included via upstream changes to mne.Report.add_epochs(). But we can merge here and move the feature to MNE-Python at a later time. Would be good to create an issue as a reminder

larsoner:
Found another bug! `--no-cache` didn't work. Pushing now. Also nesting the rejection threshold stuff under the `Epochs: after cleaning` section of the report. Marking for merge-when-green but let me know if you'd like me to hold off @hoechenberger !

hoechenberger (approved):

PR #840, opened 2024-01-30
ENH: Write out raw data and SSP events

### LINKED ISSUE #831: ENH: Preprocessed raw data

larsoner:
### Before merging …

Closes #831

Also does some cleanup of `processing=` where we errantly used some uncleaned data -- not just no SSP (which would be taken into account during inversion at least), but no ICA, no regression, and no filtering it seems.


larsoner:
All green, ready for review/merge from my end @hoechenberger @agramfort !

I pushed an empty commit just to remove an errant line from `v1.6.md` so no need to run CIs again, can just be merged without waiting for requirements if there are no review comments.

larsoner:
... you can see an example of the `Raw (clean)` here:

https://output.circle-artifacts.com/output/job/5a50001f-886a-48c7-8e3f-cbcec122019e/artifacts/0/site/examples/ds000248_base/sub-01_task-audiovisual_report.html#Raw__clean___run__01_

I think it's useful enough to have these raw runs for comparison with the original and filtered ones in terms of time traces and PSDs to just have this always on (rather than add an option for it, which was a potential idea in #831)

mne_bids_pipeline/steps/preprocessing/_08b_apply_ssp.py
```diff
@@ -81,13 +71,85 @@ def apply_ssp(
...
+    return _prep_out_files(exec_params=exec_params, out_files=out_files)
+
+
 def get_config(
     *,
     config: SimpleNamespace,
+    subject: str,
 ) -> SimpleNamespace:
     cfg = SimpleNamespace(
+        processing="filt" if config.regress_artifact is None else "regress",
```
drammock, line 150: this is the third time I've seen the same inline "filt-or-regress" if-else. Makes me worry about forgetting to copy it where it's needed, when adding new code. Not sure if there is a sensible way to reduce that cognitive overhead / duplication that wouldn't feel over-engineered, though.

drammock (commented):
@larsoner I took a look as requested. Changes look reasonable (with the caveat that, as you know, I'm not super familiar with this codebase yet).  Just one comment.

PR #854, opened 2024-02-16
ENH: Better event count rep and noise_cov_method

### LINKED ISSUE #846: ENH: Improve table formatting of events
An example:

Screenshot from 2024-02-06 14-29-41

1. If `NaN` / zero, I think we should just omit it. It'll make clearer where there actually are events.
2. Event counts should be integers, i.e., `24` not `24.0`

larsoner:
### Before merging …

Closes #846

Alongside https://github.com/mne-tools/mne-bids/pull/1227 and https://github.com/mne-tools/mne-python/pull/12447 produces for example:

| [main](https://mne.tools/mne-bids-pipeline/dev/examples/MNE-phantom-KIT-data/sub-average_task-phantom_report.html#Event_counts) | PR |
| -- | -- |
| Screenshot from 2024-02-16 16-20-29 | Screenshot from 2024-02-16 16-30-59 |

It's more impressive on other examples but you can at least see that it turns it into a datatable and also fixes the overflow.

Also:

- Snuck in a `noise_cov_method = "shrunk"` option, too. It's useful when debugging to set this to `empirical` for example for speed.
- Added a `--pdb` alias for `--debug` to mirror the `pytest` option by the same name (I mistakenly use `--pdb` *a lot* and I don't think there's any issue with having an alias)

### DIFF
--- a/docs/source/settings/source/inverse.md
+++ b/docs/source/settings/source/inverse.md
@@ -10,5 +10,6 @@ tags:
         - depth
         - inverse_method
         - noise_cov
+        - noise_cov_method
         - source_info_path_update
         - inverse_targets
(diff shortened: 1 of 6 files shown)


drammock (approved):

hoechenberger:
Thank you, very nice indeed

Issue #848, opened 2024-02-08
Adjust vmin/vmax parameters in epochs image in report to rejection limits


SophieHerbst:
When checking the reports for the cleaned epochs, I would like to set the scale limits  to the defined epochs rejection thresholds.
Otherwise, when the scale limits are set automatically  it is impossible to interpret the colors and figure out quickly whether the thresholds were high enough. 
If I am not mistaken, this requires an adjustement in mne.report.add_epochs, to hand down the rejection limits?
In the pipeline, I would suggest to make it the default that the epochs images are scaled to the rejection limits when clean epochs are displayed in the report.
WDYT?

larsoner:
One disadvantage is that if people have their thresholds set very high because they know they only want to exclude artifacts are huge (or because autoreject sets them that way), but their signals of interest are small (and SNR high other than the large artifacts that get rejected), then the range of the epochs images will end up being compressed (all white)

SophieHerbst:
I see your point. Is there a custom way for me to adjust them in the report created by the pipeline to have the same limits for all my participants?

SophieHerbst:
any insights on how to do this @hoechenberger ?

hoechenberger:
@SophieHerbst It would require a change to MNE-Python and subsequently to the Pipeline. Nothing difficult, but still a bit of work:

- First, add a new parameter `image_kwargs` to `Report.add_epochs()` here: https://github.com/mne-tools/mne-python/blob/70455902ba834bec75ae1fd0f7f5b2fc48455cf8/mne/report/report.py#L1095 (consistent with the existing `topomap_kwargs`)
- Add and pass it down to `Report._add_epochs()`, https://github.com/mne-tools/mne-python/blob/70455902ba834bec75ae1fd0f7f5b2fc48455cf8/mne/report/report.py#L1141
- Inside `_add_epochs()`, pass it to `plot_image()`: https://github.com/mne-tools/mne-python/blob/70455902ba834bec75ae1fd0f7f5b2fc48455cf8/mne/report/report.py#L3939

Then we could do `Report.add_epochs(..., image_kwargs={"vmin": ..., "vmax": ...})`

Once we have this, we'd need to expose this functionality (or a more restricted subset of it) via the Pipeline configuration.

SophieHerbst:
Mh, ok, it sounds like a bit of work, but I would find it very useful. Shall I give it a try?

hoechenberger:
> Shall I give it a try?

[...]

PR #859, opened 2024-02-23
add image_kwargs to report

SophieHerbst:
Responds to issue #848

Now that mne has been updated, we can pass image_kwargs to report.add_epochs. 
It works when I set the limits to the rejection limits in the config, but I cannot get it right when setting image_kwargs to None as per default. 

```
 File "/home/sh254795/Documents/REPOS/mne-study-template/mne_bids_pipeline/_config_import.py", line 340, in _default_factory
    idx = allowlist.index(val)
ValueError: {'image_kwargs': None} is not in list

    assert val == typ(), (key, val)
AssertionError: ('report_add_epochs_kwargs', {'image_kwargs': None})
```


mne_bids_pipeline/_config.py
```diff
@@ -2131,6 +2131,20 @@ def noise_cov(bids_path):
     ```
 """
 
+report_add_epochs_image_kwargs: Optional[dict] = dict(image_kwargs=None)
```
larsoner, line 2134:
```suggestion
report_add_epochs_image_kwargs: Optional[dict] = dict()
```

mne_bids_pipeline/_config.py
```diff
@@ -2131,6 +2131,20 @@ def noise_cov(bids_path):
...
+Specifies the limits for the color scales of the epochs_image in the report.
+If `None`, it defaults to the current default in MNE-Python.
+
+???+ example "Example"
+    Set vmin and vmax to the epochs rejection thresholds:
+    ```python
+    report_add_epochs_kwargs = dict(image_kwargs=dict(
+        grad = dict(vmin=-reject['grad'], vmax=reject['grad']), 
+        mag=dict(vmin=-reject['mag'],vmax=reject['mag'])))
+    ```
```
larsoner, line 2145: I would expect:
```suggestion
    report_add_epochs_kwargs = dict(
        grad = dict(vmin=-reject['grad'], vmax=reject['grad']), 
        mag=dict(vmin=-reject['mag'],vmax=reject['mag']))
    ```
```

mne_bids_pipeline/steps/preprocessing/_07_make_epochs.py
```diff
@@ -256,6 +256,7 @@ def run_epochs(
             psd=psd,
             drop_log_ignore=(),
             replace=True,
+            image_kwargs=cfg.report_add_epochs_image_kwargs["image_kwargs"],
```
larsoner, line 259:
```suggestion
            image_kwargs=cfg.report_add_epochs_image_kwargs,
```

mne_bids_pipeline/_config.py
hoechenberger: Should be `None`, yes

mne_bids_pipeline/_config.py
hoechenberger: Why the love for `dict`? I think `{}` is much easier to read and more versatile:

```python
    report_add_epochs_kwargs = {
        "grad": {
            "vmin": -reject["grad"],
            "vmax": reject["grad"]
        }, 
        "mag": {
            "vmin": -reject["mag"],
            "vmax": reject["mag"]
        }
    }
```

SophieHerbst:
Also, it turns out that in order to use the rejection criteria from my config, I had to deal with the different scaling between rejection and plotting commands:
# # add image kwargs for report
# report_add_epochs_image_kwargs = {
#     "grad": {"vmin": 0, "vmax": reject["grad"]*1e15},
#     "mag": {"vmin": 0, "vmax": reject["mag"]*1e15},
# }

larsoner:
For the consistency we should merge https://github.com/mne-tools/mne-bids-pipeline/pull/863 first probably, it should fix things for you I think

We should maintain compat with latest MNE stable, so you'll need to `inspect.getfullargspec(report.add_epochs)` to determine whether or not to set `image_kwargs` I think

SophieHerbst:
I think I need some help here

larsoner (approved):
@SophieHerbst pushed a few commits to DRY (don't repeat yourself) the code a bit, and also test that the option works -- locally the report seems to reflect `reject`:

image

Can you look and make sure you're still happy?

SophieHerbst:
Thank you @larsoner I ran it again on my data and I am very happy with it now.

hoechenberger:
Thanks @SophieHerbst and @larsoner!!!

Issue #862, opened 2024-02-26
DOC: Document steps where config options are used


larsoner:
> Used as in informing the documentation?

Yes

We'd basically have to iterate over all steps and grep the arguments used in `get_config()` and then put that in a table or so…

_Originally posted by @hoechenberger in https://github.com/mne-tools/mne-bids-pipeline/issues/860#issuecomment-1963909677_

larsoner:
> (I learned about this [adding params to documentation] part last week, when I forgot to mention the image_kwargs for the report there; easy to make mistakes as it needs to be added manually)

Yeah we at least added a CI job (I think!) that checks to make sure all `_config.py` options are in the docs, so CIs yell at you if you don't add them so we stay consistent even if it takes a few commits to get there.

In principle instead we could think about trying to mark up the default `_config.py` with some `# ## Preprocessing` headings for example that get found with `ast`, parsed, and automatically populate `.md` files. It wouldn't be too difficult but not sure if it would cover everything.

> We'd basically have to iterate over all steps and grep the arguments used in get_config() and then put that in a table or so…

Yeah or in addition to or instead of a monolithic table we could figure out how to document it inline with the param itself, like for example

![Screenshot from 2024-02-26 09-11-42](https://github.com/mne-tools/mne-bids-pipeline/assets/2365790/05a8d1f7-e96d-41bc-ae80-5a30ba6bab97)

Could have a list of all three decoding steps right at the end, since they are all *directly* affected by the choice (assuming that's the complete set anyway!). Complementary to a table I guess since when looking at a config option you probably wonder "what steps will this option affect?" in which case the inline list is useful; but when looking at a *processing step* you might wonder "what options are used in this step?" in which case a table is more useful.

larsoner:
... for the table half, we already have a step table in https://mne.tools/mne-bids-pipeline/dev/features/steps.html , perhaps we could add config options to each table entry. The table would get a bit long but I think that's okay.

hoechenberger:
> Yeah or in addition to or instead of a monolithic table we could figure out how to document it inline with the param itself, like for example

I wasn't thinking of a table to include it in the docs, but what I had it mind was a look-up table (LUT) we could then use to enrich the docs; my idea would be that every parameter gets an Info button or something that, when clicked on, displays which steps will be affected

PR #863, opened 2024-02-26
DOC: Automatically document settings

larsoner:
This PR automatically adds configuration options from `_config.py` to `docs/source/settings/`. It's a first step toward #862. To accomplish this, I wrote code to parse header levels `#` , `##`, `###` and map those onto our current hierarchy. I made a few opinionated choices:

1. Make a new `Execution` section that has stuff like `n_jobs`. It seems useful to separate those from stuff like `bids_root` which are more dataset-specific and actually affect *what gets generated* as opposed to *how it gets generated on machines / jobs etc.*.

   image

3. Chose the website docs as the source of truth for parameter order. This means the `_config.py` diff is bigger, but the `.md` diffs are smaller.
4. Chose to add the `tags` to the `_config.py`. This was a bit of a toss-up for me -- if it is not useful there I can easily just move those to a dict in `gen_settings.py`.

For now I have left the generated `.md` files in source control *but this is just to facilitate review to see how the docs have changed*. Once people are happy I will `rm` and add to `.gitignore`.

Link to rendered doc: https://output.circle-artifacts.com/output/job/0a5e0fe7-0af8-4717-80f5-e33f99967966/artifacts/0/site/settings/general.html

### Before merging …

- [x] Remove `.md` files that are autogenerated and add to `.gitignore`


mne_bids_pipeline/_config.py
```diff
@@ -95,6 +94,11 @@
 The task to process.
 """
 
+task_is_rest: bool = False
+"""
+Whether the task should be treated as resting-state data.
+"""
```
larsoner, line 100: All changes like this are just copy-paste moves to match orders in `.md` files above, no content changes.

mne_bids_pipeline/_config.py
```diff
@@ -426,9 +422,40 @@
...
+
+# ## Break detection
+#
+# ---
+# tags:
+#   - preprocessing
+#   - artifact-removal
+#   - raw
+#   - events
+# ---
```
larsoner, line 458: @hoechenberger this is my one open question -- useful to have the `tags` in the `_config.py` or no? If not I'll move to the parsing `.py` file instead.

mne_bids_pipeline/_config.py
```diff
@@ -527,10 +554,30 @@
...
+#   - preprocessing
+#   - raw
+#   - bad-channels
+# ---
+#
+# !!! warning
+#     This functionality will soon be removed from the pipeline, and
+#     will be integrated into MNE-BIDS.
+#
+# "Bad", i.e. flat and overly noisy channels, can be automatically detected
```
larsoner, line 571: Blocks of text like this were taken from the `.md` files and moved to the `_config.py`, which I think is a quality-of-life improvement for end users. We had little bits of this stuff in some places and some outdated info, and now it's documented nicely in the code and in the website the same way.

mne_bids_pipeline/_config.py
```diff
@@ -1,4 +1,4 @@
-# Default settings for data processing and analysis.
+"""Default settings for data processing and analysis."""
```
larsoner, line 1: ... and I'm going to revert this since it injects this line at the top of each rendered `.md` file

mne_bids_pipeline/_config.py
hoechenberger: How would you add them to the Python files? What would that look like?

larsoner: In the parsing gen file something like
```
tags = {
    "break detection": ("preprocessing", "artifact-removal", "raw", "events"),
    ...
```
One tuple of tags per option category, then the parser adds them to the .md files in the right place

hoechenberger: I don't have a clear preference here, I suppose I'm fine with either

larsoner: Thinking about it more I decided to move it to `gen_settings.py` -- that way someday we could consider auto-generating / adding tags. And I think it keeps `_config.py` a little bit more compact which is nice.

larsoner:
Even with all the parsing code we end up with a net-negative line count in the diff `+511 −689` which is nice! Ready for review/merge from my end @hoechenberger

hoechenberger:
@larsoner I won't be able to thoroughly review for now; please go ahead and merge if you're happy and i'll start complaining sometime later 😁

larsoner:
> i'll start complaining sometime later 😁

Complaints always welcome, but even more welcome if they're in the form of a PR to improve wordings etc. :smile:

SophieHerbst:
In my view, it would be helpfull to explain the memory-related arguments, and more generally the caching behavior in a dedicated page, like the basic usage

hoechenberger:
> In my view, it would be helpfull to explain the memory-related arguments, and more generally the caching behavior in a dedicated page, like the basic usage

i agree!!

PR #869, opened 2024-03-01
ENH: Doc config in sheet

### LINKED ISSUE #868: ENH: Simplify config in logs
If we follow this premise, we could create a single Excel sheet where one row corresponds to one line or one setting in the config file that was used

_Originally posted by @hoechenberger in https://github.com/mne-tools/mne-bids-pipeline/pull/865#discussion_r1508941664_

FWIW I did start looking into adding just a `config` sheet with something like:
```
@@ -317,10 +292,18 @@ def save_logs(*, config: SimpleNamespace, logs: list[pd.Series]) -> None:
             mode="a" if append else "w",
             if_sheet_exists="replace" if append else None,
         )
+        assert isinstance(config, SimpleNamespace), type(config)
+        cf_df = dict()
+        for key, val in config.__dict__.items():
+            cf_df[key] = json_tricks.dumps(val, sort_keys=False)
+        cf_df = pd.DataFrame(cf_df)
         with writer:
+            # Config first then the data
+            cf_df.to_excel(writer, sheet_name="config", index=False)
             df.to_excel(writer, sheet_name=sheet_name, index=False)
```
but it fails because the `config` at this level isn't trimmed to the only relevant entries. But this should be fixable in principle if we want to go this route.

larsoner:
### Before merging …

Closes #868

Result:

@hoechenberger in principle we *could* split this into multiple rows, but I think it's perhaps nicer this way -- if you want to see a multiline value you click in the cell and you can see it, select, copy-paste etc. If we put in multiple rows it might not be as easy to do that.

image

The truncation code is still present, but it really only applies to the `DigMontage` because its `json_tricks` dump is so long (contains all channel locations, coord frame, etc from `DigMontage`.).

Also this contains *all* config values, both those supplied by the user and those that are defaults. I think it's nice because it shows all values that are used. If people want to know which values they supplied, that complementary information is in the `_report.html` files.

### DIFF
--- a/docs/source/v1.6.md.inc
+++ b/docs/source/v1.6.md.inc
@@ -11,6 +11,7 @@
 - Added [`noise_cov_method`][mne_bids_pipeline._config.noise_cov_method] to allow for the use of methods other than `"shrunk"` for noise covariance estimation (#854 by @larsoner)
 - Added option to pass `image_kwargs` to [`mne.Report.add_epochs`] to allow adjusting e.g. `"vmin"` and `"vmax"` of the epochs image in the report via [`report_add_epochs_image_kwargs`][mne_bids_pipeline._config.report_add_epochs_image_kwargs] (#848 by @SophieHerbst)
 - Split ICA fitting and artifact detection into separate steps. This means that now, ICA is split into a total of three consecutive steps: fitting, artifact detection, and the actual data cleaning step ("applying ICA"). This makes it easier to experiment with different settings for artifact detection without needing to re-fit ICA. (#865 by @larsoner)
+- The configuration used for the pipeline is now saved in a separate spreadsheet in the `.xlsx` log file (#869 by @larsoner)
 
 [//]: # (### :warning: Behavior changes)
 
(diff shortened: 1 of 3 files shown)


larsoner:
And here's how I'm getting it to deal with callables by adding the path + function name:

image

larsoner:
... @hoechenberger should we cut 1.6 after this PR?

drammock:
> ... @hoechenberger should we cut 1.6 after this PR?

I know you're not asking me, but I would appreciate a new release, as I'm about to start in on a new project in the coming days (thus freezing the environment; would be nice to be on stable and have all the great additions you two have done lately)

hoechenberger:
yes let's cut one, i'm generally a big friend of frequent releases

hoechenberger:
@larsoner We depend on MNE 1.7 though 

we should mention this in the changelog 


not sure how to handle this properly, since 1.7 isn't out yet

larsoner:
> @larsoner We depend on MNE 1.7 now, though

For which features? I made the `add_epochs_image_kwargs` optional. Right now it will silently ignore them if you aren't on 1.7+ but I could make it raise an error or emit a warning instead.

Was there something else that required 1.7+?

hoechenberger:
> > @larsoner We depend on MNE 1.7 now, though
> 
> 
> 
> For which features? I made the `add_epochs_image_kwargs` optional. Right now it will silently ignore them if you aren't on 1.7+ but I could make it raise an error or emit a warning instead.
> 
> 
> 
> Was there something else that required 1.7+?

ah ok! this was the only feature I had in mind. We should at least document in the changelog that this feature will only work with MNE dev or 1.7 once released

larsoner:
Okay pushed a commit with the doc requirement, good to merge this one?

hoechenberger:
good to go for me!

larsoner:
@hoechenberger okay for me to cut a basic release on GH and you can edit the notes afterward when you get a chance?

hoechenberger:
yes let's do it 👍

PR #872, opened 2024-03-03
Enable stricter linting

hoechenberger:
### Before merging …


mne_bids_pipeline/_config.py
```diff
@@ -1,12 +1,13 @@
...
-from typing import Annotated, Any, Callable, Literal, Optional, Sequence, Union
+from collections.abc import Sequence
+from typing import Annotated, Any, Callable, Literal, Optional, Union
 
 from annotated_types import Ge, Interval, Len
 from mne import Covariance
 from mne_bids import BIDSPath
 
-from mne_bids_pipeline.typing import (
+from mne_bids_pipeline._types import (
```
larsoner, line 10: Speaking of public vs private, this one was meant to be public like [numpy.typing](https://numpy.org/devdocs/reference/typing.html) because people might want to annotate with some of our types like ArbitraryContrast

larsoner: Like `from mne_bids_pipeline.typing import ArbitraryContrast` or similar

hoechenberger: I will revisit this tomrrow

hoechenberger: @larsoner I have reverted this change

PR #888, opened 2024-03-14
Remove ancient config parameter `shortest_event`

hoechenberger:
### Before merging …

- [x] Changelog has been updated (`docs/source/changes.md`)

PR #889, opened 2024-03-14
MAINT: Check that all options are used

larsoner:
### Before merging …

Follow-up to #888:

1. Adds test that we do use all options somewhere
2. Remove `study_name` which we don't use anywhere (even though it's our first listed config option!)


mne_bids_pipeline/_docs.py
```diff
@@ -0,0 +1,261 @@
...
+                            "_import_data_kwargs",
+                            "get_runs",
+                            "get_subjects",
+                            "get_sessions",
+                        ):
+                            raise RuntimeError(
+                                f"{where} cannot handle call {keyword.value.func.id=} "
+                                f"for {key}"
+                            )
+                        # Get the source and regex for config values
```
larsoner, line 208: This file is copy-paste then refactored to have constants at the top, plus added regex for finding `config.` stuff in functions like `_bids_kwargs` and `get_runs` etc.

hoechenberger:
very nice!

hoechenberger:
> 2\. Remove `study_name` which we don't use anywhere (even though it's our first listed config option!)

🫣😅

PR #892, opened 2024-03-19
MAINT: Print longer traceback on report error

larsoner:
Don't think this one needs a changelog update, but I was getting an error related to NumPy import and just printing the `exc` itself wasn't enough to see it.


hoechenberger:
Thanks Eric!

PR #908, opened 2024-03-27
Drop support for Python 3.9, drop `Optional` and `Union` from type hints

### LINKED ISSUE #906: Drop support for Python 3.9?
I would love to drop support for Python 3.9, which would allow us to write types such as

```python
Optional[Union[str, float]]
```
as
```python
None | str | float
```
which is much more readable. I'm mostly thinking about the configuration file here, which is immediately user-facing.

Thoughts?

hoechenberger:
Fixes #906

@larsoner I simply had to change the min. Python version in `pyproject.toml` and run `ruff`; only two additional manual interventions were required after that. Super nice.

### Before merging …

### DIFF
--- a/docs/source/settings/gen_settings.py
+++ b/docs/source/settings/gen_settings.py
@@ -101,14 +101,18 @@ prefix = """\
 assign_re = re.compile(
     # Line starts with annotation syntax (name captured by the first group).
     r"^(\w+): "
-    # Then the annotation can be ...
-    "("
+    # Then the annotation can be (in a non-capturing group) ...
+    "(?:"
     # ... a standard assignment ...
     ".+ = .+"
     # ... or ...
     "|"
-    # ... the start of a multiline type annotation like "a: Union["
-    r"(Union|Optional|Literal)\["
+    # ... the start of a multiline type annotation like "a: Literal[" ...
+    r"Literal\["
+    # ... or ...
+    "|"
+    # ... the start of a multiline type annotation like "a: ("
+    r"\("
     # To the end of the line.
     ")$",
     re.MULTILINE,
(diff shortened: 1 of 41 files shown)


hoechenberger:
Diff for the config file really shows how much more readable things are now:
https://github.com/mne-tools/mne-bids-pipeline/pull/908/files#diff-ab728fad8c0aee89735be8e90dda245e871f17892fc41d1b3a7377e3eccf58f7

hoechenberger:
@larsoner I assume this line / regexp needs to be changed:

https://github.com/mne-tools/mne-bids-pipeline/blob/ed5939a1d8c820197c0d69d5455693c5b4898e43/docs/source/settings/gen_settings.py#L111

to accommodate the way the type for `noise_cov` is specified now:
```python
noise_cov: (
    ...
) = ...
```
But I couldn't get it to work so far :(

docs/source/settings/gen_settings.py
```diff
@@ -99,18 +99,16 @@
...
+    "^"  #         The line starts, then is followed by
+    r"(\w+): "  #  annotation syntax (name captured by the first group),
+    "(?:"  #       then the rest of the line can be (in a non-capturing group):
+    ".+ = .+"  #     1. a standard assignment
+    "|"  #           2. or
+    r"Literal\["  #  3. the start of a multiline type annotation like "a: Literal["
+    "|"  #           4. or
+    r"\("  #         5. the start of a multiline 3.9+ type annotation like "a: ("
+    ")"  #         Then the end of our group
+    "$",  #       and immediately the end of the line.
```
larsoner, line 111: @hoechenberger this should work, make more sense?

larsoner:
Marking for merge-when-green, we'll see if my `re` tweak works 🤞 Thanks in advance @hoechenberger !

PR #911, opened 2024-03-29
Do not show `Annotated` types in configuration options documentation

hoechenberger:
I want to get rid of the `Annotated` bits e.g. here:
Screenshot 2024-03-29 at 14 28 15

and only include the "actual" type (the `Literal[...]` in this specific case).

### Before merging …

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1031,6 +1031,10 @@ jobs:
           at: ~/
       - bash_env
       - run:
+          name: Install dependencies
+          command: |
+            pip install -ve .[docs]
+      - run:
           name: Build documentation
           command: |
             set -eo pipefail
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -59,6 +59,14 @@ tests = [
   "pooch",
   "psutil",
   "ruff",
+  "jinja2",
+  "openneuro-py >= 2022.2.0",
+  "httpx >= 0.20",
+  "tqdm",
+  "Pygments",
+  "pyyaml",
+]
+docs = [
   "mkdocs",
   "mkdocs-material >= 9.0.4",
   "mkdocs-material-extensions",
@@ -67,13 +75,8 @@ tests = [
   "mkdocs-exclude",
   "mkdocstrings-python",
   "mike",
-  "jinja2",
   "livereload",
-  "openneuro-py >= 2022.2.0",
-  "httpx >= 0.20",
-  "tqdm",
-  "Pygments",
-  "pyyaml",
+  "black",                          # docstring reformatting
 ]
 
 [project.scripts]


larsoner:
We will lose some information here like the `Len(1, 4)` or `Ge` etc. but I guess that's okay

hoechenberger:
> We will lose some information here like the `Len(1, 4)` or `Ge` etc. but I guess that's okay

Yes but I don't think this stuff is very user-readable anyway, so I'd love to keep it out of this part of the docs. If we think we need to document it, we should add it to the docstring in "proper" words.

WDYT?

larsoner:
Okay by me. Some of it is obvious anyway like people shouldn't pass an empty list for `ch_types` etc.

hoechenberger:
https://output.circle-artifacts.com/output/job/71b0ad3a-e6cd-4551-a09f-e685b3995872/artifacts/0/site/settings/general.html#mne_bids_pipeline._config.ch_types

It's working

… and makes me wonder: huh, the default here actually doesn't match the annotation – it's an empty list, which is in violation of the annotation:

```python
ch_types: Annotated[Sequence[Literal["meg", "mag", "grad", "eeg"]], Len(1, 4)] = []
```

@larsoner Do you think we could change the default to `None` and simply use all available & supported channel types in this case? Otherwise, I don't know what a good default would look like.

larsoner:
> Do you think we could change the default to None and simply use all available & supported channel types in this case? Otherwise, I don't know what a good default would look like.

Sure we could add this as an enhancement. At the moment it's not too bad since if users don't specify this they immediately get a clear error message saying they need to specify it (here I've removed `ch_types` from the `ds004229` config):
```
$ pytest mne_bids_pipeline/ -k ds004229
...
E   ValueError: 1 validation error for user configuration from /home/larsoner/python/mne-bids-pipeline/mne_bids_pipeline/tests/configs/config_ds004229.py
E   ch_types
E     Value should have at least 1 item after validation, not 0 [type=too_short, input_value=[], input_type=list]
E       For further information visit https://errors.pydantic.dev/2.6/v/too_short
```
but `ch_types = None` meaning "autodetect and use all" seems like nicer behavior

larsoner:
Thanks @hoechenberger

larsoner:
Looks like the `pyproject.toml` reorg broke doc deployment

https://app.circleci.com/pipelines/github/mne-tools/mne-bids-pipeline/4433/workflows/70e9a5b2-103a-47ec-893c-3933adf7df01/jobs/63063

hoechenberger:
Argh… 

I really totally lost track of how all that stuff works, I find our CircleCI config extremely confusing :(
Will try to take a look at it later

PR #914, opened 2024-03-30
Restructure configuration options documentation sections

hoechenberger:
I've replaced the `Execution` section with the folliwing new sections

- Caching
- Parallelization
- Logging
- Error handling

### Before merging …

### DIFF
--- a/docs/source/v1.9.md.inc
+++ b/docs/source/v1.9.md.inc
@@ -6,17 +6,21 @@
 - The type annotations in the default configuration file are now easier to read: We
   replaced `Union[X, Y]` with `X | Y` and `Optional[X]` with `X | None`. (#908, #911 by @hoechenberger)
 
-[//]: # (- Whatever (#000 by @whoever))
-
 [//]: # (### :warning: Behavior changes)
 
 [//]: # (- Whatever (#000 by @whoever))
 
 ### :package: Requirements
 
-- We dropped support for Python 3.9. You now need Python 3.10 or newer.
+- We dropped support for Python 3.9. You now need Python 3.10 or newer. (#908 by @hoechenberger)
 
 [//]: # (- Whatever (#000 by @whoever))
+
+### :book: Documentation
+
+- We removed the `Execution` section from configuration options Documentation and
+  replaced it with new, more explicit sections (namely, Caching, Parallelization,
+  Logging, and Error handling). (#914 by @hoechenberger)
 
 ### :bug: Bug fixes
 
--- a/docs/mkdocs.yml
+++ b/docs/mkdocs.yml
@@ -103,7 +103,11 @@ nav:
             - Source space & forward solution: settings/source/forward.md
             - Inverse solution: settings/source/inverse.md
         - Report generation: settings/reports/report_generation.md
-        - Execution: settings/execution.md
+        - Caching: settings/caching.md
+        - Parallelization: settings/parallelization.md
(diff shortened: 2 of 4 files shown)


docs/source/v1.9.md.inc
```diff
@@ -6,18 +6,22 @@
...
 ### :package: Requirements
 
-- We dropped support for Python 3.9. You now need Python 3.10 or newer.
+- We dropped support for Python 3.9. You now need Python 3.10 or newer. (#908 by @hoechenberger)
 
 [//]: # (- Whatever (#000 by @whoever))
 
+### :book: Documentation
+
+- We removed the `Execution` section from configuration options Documentation and
```
hoechenberger, line 21:
```suggestion
- We removed the `Execution` section from configuration options documentation and
```

hoechenberger:
See https://output.circle-artifacts.com/output/job/af987afa-de5d-47e5-a8c0-c7d5a2fd5a1d/artifacts/0/site/settings/caching.html etc

cc @SophieHerbst

SophieHerbst:
Sorry, I am late to the party.. 
I think that the info about caching could be a bit more explicit and complete. 
Suggestion: Per default, the pipeline output is cached, meaning temporarily stored, to avoid unnecessary reruns of previously computed steps. However, for consistency, changes in configuration parameters trigger automatic reruns of previous steps. To force rerunning a given step, run the pipeline with the option: --no-cache

SophieHerbst:
Just a thought, wouldn't it make sense to set the default logging for mne to 'info' instead of error? 
`mne_log_level: Literal['info', 'error'] = 'error'`
In my view, this would remove some of the blackboxyness of the pipeline..

hoechenberger:
> I think that the info about caching could be a bit more explicit and complete. Suggestion: Per default, the pipeline output is cached, meaning temporarily stored, to avoid unnecessary reruns of previously computed steps. However, for consistency, changes in configuration parameters trigger automatic reruns of previous steps. To force rerunning a given step, run the pipeline with the option: --no-cache

That's a great idea, would you like to add this amendment? You'd need to edit this:

https://github.com/mne-tools/mne-bids-pipeline/blob/d8c1e4b683f45567f0c1aad59ab87c89355bf34b/mne_bids_pipeline/_config.py#L2174-L2178

hoechenberger:
> Just a thought, wouldn't it make sense to set the default logging for mne to 'info' instead of error? `mne_log_level: Literal['info', 'error'] = 'error'` In my view, this would remove some of the blackboxyness of the pipeline..

I like the idea of being more verbose in the sense of, communicating what we're doing.

However please note that we currently have two basically independent logging systems: the one for the pipeline itself and the one for MNE.

MNE often produces LOADS of output that just makes everything unreadable when processing multiple subjects in parallel. Even worse, since MNE uses a separate logging system, none of those MNE log messages are being formatted "the pipeline way" (no colors, no info about which subject/task/run the message is about, etc)

So usually, this creates a mess.

I would be open for this if we implement a logging system where maybe the user can expand and collapse entire log sections... this is possible via `rich`, which I believe we use for logging already, but it will be a major effort to actually get this "right".

SophieHerbst:
I see your point. Too much unformatted output will clearly not help. So +1 for the rich output in a separate file.

hoechenberger:
> So +1 for the rich output in a separate file.

Can you open a new issue / feature request for this?
We should probably have on log file per subject

PR #915, opened 2024-03-30
Improve documentation and config validation of `loose` and `depth` parameters; drop support for `loose=None`

hoechenberger:
@larsoner @SophieHerbst WDYT?

We currently don't support volumetric or mixed source spaces anyway, so I thought we could simplify things a little.

I was even considering entirely removing the `loose` and `depth` settings but then wasn't sure if some of you might be using them sometimes? Personally I never touch those.

### Before merging …

### DIFF
--- a/docs/source/v1.9.md.inc
+++ b/docs/source/v1.9.md.inc
@@ -10,7 +10,7 @@
 
 ### :warning: Behavior changes
 
-- The [`loose`][mne_bids_pipeline._config.depth] parameter doesn't accept `"auto""`
+- The [`loose`][mne_bids_pipeline._config.loose] parameter doesn't accept `"auto""`
   anymore. Please use `0.2` instead (the default value). (#915 by @hoechenberger)
 
 - The [`depth`][mne_bids_pipeline._config.depth] parameter doesn't accept `None`
--- a/docs/source/v1.9.md.inc
+++ b/docs/source/v1.9.md.inc
@@ -8,7 +8,13 @@
 
 [//]: # (- Whatever (#000 by @whoever))
 
-[//]: # (### :warning: Behavior changes)
+### :warning: Behavior changes
+
+- The [`loose`][mne_bids_pipeline._config.depth] parameter doesn't accept `"auto""`
+  anymore. Please use `0.2` instead (the default value). (#915 by @hoechenberger)
+
+- The [`depth`][mne_bids_pipeline._config.depth] parameter doesn't accept `None`
+  anymore. Please use `0` instead. (#915 by @hoechenberger)
 
 [//]: # (- Whatever (#000 by @whoever))
 
--- a/docs/source/v1.9.md.inc
+++ b/docs/source/v1.9.md.inc
@@ -10,9 +10,6 @@
 
 ### :warning: Behavior changes
 
-- The [`loose`][mne_bids_pipeline._config.loose] parameter doesn't accept `"auto""`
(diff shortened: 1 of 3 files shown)


docs/source/v1.9.md.inc
```diff
@@ -8,7 +8,13 @@
 
 [//]: # (- Whatever (#000 by @whoever))
 
-[//]: # (### :warning: Behavior changes)
+### :warning: Behavior changes
+
+- The [`loose`][mne_bids_pipeline._config.depth] parameter doesn't accept `"auto""`
```
hoechenberger, line 13:
```suggestion
- The [`loose`][mne_bids_pipeline._config.loose] parameter doesn't accept `"auto""`
```

mne_bids_pipeline/_config.py
```diff
@@ -1979,23 +1979,26 @@ def mri_landmarks_kind(bids_path):
...
-inverse solution is computed with **fixed orientation.**
-If `1`, it corresponds to **free orientation.**
-The default value, `'auto'`, is set to `0.2` for surface-oriented source
-spaces, and to `1.0` for volumetric, discrete, or mixed source spaces,
-unless `fixed is True` in which case the value 0. is used.
+A value that weights the source variances of the dipole components
+that are parallel (tangential) to the cortical surface.
+
+If `0`, then the inverse solution is computed with **fixed orientation**, i.e.,
+only dipoles perpendicular to the cortical surface are considered.
```
hoechenberger, line 1988:
```suggestion
only dipole components perpendicular to the cortical surface are considered.
```

mne_bids_pipeline/_config.py
```diff
@@ -1979,23 +1979,26 @@ def mri_landmarks_kind(bids_path):
...
+
+If `1`, it corresponds to **free orientation**, i.e., dipoles with any orientation are
```
hoechenberger, line 1990:
```suggestion
If `1`, it corresponds to **free orientation**, i.e., dipole components with any orientation are
```

larsoner:
Eventually we could pretty easily add support for volumetric source estimates I think. So I'm not sure it's worth getting rid of `'auto'` just to potentially add it back. And I think `loose` and `depth` are both useful from time to time, especially when working with template / non-individualized MRIs (you often want `loose=1.`)

SophieHerbst:
I agree about keeping the loose and depth parameters, we do use them sometimes. 
No opinion about the 'auto' option, but if mixed source spaces can be implemented easily, that would be useful.

SophieHerbst:
I have always found the options for those parameters a bit confusing, so I am fine with removing the None.

larsoner:
Okay with me to remove `None`

mne_bids_pipeline/_config.py
```diff
@@ -1979,24 +1979,33 @@ def mri_landmarks_kind(bids_path):
 
 # ## Inverse solution
 
-loose: float | Literal["auto"] = 0.2
+loose: Annotated[float, Interval(ge=0, le=1)] = 0.2
```
larsoner, line 1982: To restore old `auto` behavior, something like
```suggestion
loose: Annotated[float, Interval(ge=0, le=1)] | Literal["auto"] = 0.2
```

hoechenberger: right, forgot about that one!!

larsoner (approved):
Marking for merge when green, thanks in advance @SophieHerbst !

larsoner:
Whoops wrong PR comment, thanks @hoechenberger !

PR #919, opened 2024-04-02
change default for info to use for inverse #905

### LINKED ISSUE #905: STC saturated – rank problem?
I am not happy with the stcs of some of my subjects, they look like the screenshot attached: at one moment in time there is very high activity everywhere, which messes up the scale. I remember that a similar outcome was caused by differences in rank between the cov used for whitening, and the rank given to compute the inverse. 

With cfg.noise_cov = 'rest', the following should read the info of the resting state run to compute the inverse.

 l. 52 in  source/_05_make_inverse.py
 in_files["info"] = bids_path.copy().update(**cfg.source_info_path_update)

I cannot figure out what the **cfg.source_info_path_update does. 
Thanks for any help!

Screenshot 2024-03-27 at 17 12 03

SophieHerbst:
Change default for info given to 05_make_inverse.py (and for consistency 04_make_forward.py):
If set to `None` (default), the info will be retrieved either from the raw
file specified in `noise_cov`, or the cleaned epochs (if `noise_cov` is None or `ad-hoc`).

Closes #905

### DIFF
--- a/mne_bids_pipeline/_config.py
+++ b/mne_bids_pipeline/_config.py
@@ -2114,6 +2114,9 @@ file specified in `noise_cov`, or the cleaned epochs
                                 'suffix': 'raw',
                                 'task': 'rest'}
     ```
+    If you set `noise_cov = 'rest'` and `source_path_info = None`, then the behavior is identical
+    to that above (it will automatically use the resting state data).
+
 """
 
 inverse_targets: list[Literal["evoked"]] = ["evoked"]
--- a/mne_bids_pipeline/_config.py
+++ b/mne_bids_pipeline/_config.py
@@ -2098,7 +2098,7 @@ This parameter allows you to explicitly specify from which file to retrieve the
 `mne.Info` object. Use this parameter to supply a dictionary to
 `BIDSPath.update()` during the forward and inverse processing steps.
 If set to `None` (default), the info will be retrieved either from the raw
-file specified in `noise_cov`, or the cleaned epochs
+file specified in `noise_cov`, or the cleaned evoked
 (if `noise_cov` is None or `ad-hoc`).
 
 ???+ example "Example"
--- a/mne_bids_pipeline/_config.py
+++ b/mne_bids_pipeline/_config.py
@@ -2114,8 +2114,9 @@ file specified in `noise_cov`, or the cleaned evoked
                                 'suffix': 'raw',
                                 'task': 'rest'}
     ```
-    If you set `noise_cov = 'rest'` and `source_path_info = None`, then the behavior is identical
-    to that above (it will automatically use the resting state data).
+    If you set `noise_cov = 'rest'` and `source_path_info = None`,
+    then the behavior is identical to that above
+    (it will automatically use the resting state data).
 
 """
 
--- a/mne_bids_pipeline/_config.py
+++ b/mne_bids_pipeline/_config.py
@@ -2091,14 +2091,14 @@ of `mne.compute_covariance` for details.
 
 source_info_path_update: dict[str, str] | None = None
 """
-When computing the forward and inverse solutions, it is important to 
+When computing the forward and inverse solutions, it is important to
 provide the `mne.Info` object from the data on which the noise covariance was
-computed, to avoid problems resulting from mismatching ranks. 
+computed, to avoid problems resulting from mismatching ranks.
 This parameter allows you to explicitly specify from which file to retrieve the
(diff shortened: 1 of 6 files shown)


mne_bids_pipeline/_config.py
```diff
@@ -2089,22 +2089,31 @@ def noise_cov(bids_path):
...
     source_info_path_update = {'processing': 'clean',
                                'suffix': 'epo'}
     ```
+
+    Use the `Info` object stored in a raw file (e.g. resting state):
+    ```python
+    source_info_path_update = {'processing': 'clean',
+                                'suffix': 'raw',
+                                'task': 'rest'}
+    ```
```
larsoner, line 2116: This is true now, right?
```suggestion
    ```
    If you set `noise_cov = 'rest'` and `source_path_info = None`, then the behavior is identical
    to that above (it will automatically use the resting state data).

```

mne_bids_pipeline/steps/source/_04_make_forward.py
```diff
@@ -102,7 +102,19 @@ def get_input_fnames_forward(*, cfg, subject, session):
...
+    # for consistency with 05_make_inverse, read the info from the
+    # data used for the noise_cov
+    if cfg.source_info_path_update is None:
+        if cfg.noise_cov in ("rest", "noise"):
+            source_info_path_update = dict(
+                processing="clean", suffix="raw", task=cfg.noise_cov
+            )
+        else:
+            source_info_path_update = dict(suffix="ave")
+            # XXX is this the right solution also for noise_cov = 'ad-hoc'?
```
larsoner, line 114: Yes it should be the best we can do -- for ad-hoc we just get the rank based on the number of projectors + maxwell info.
```suggestion
            source_info_path_update = dict(suffix="ave")
```

mne_bids_pipeline/_config.py
```diff
@@ -2089,22 +2089,31 @@ def noise_cov(bids_path):
 of `mne.compute_covariance` for details.
 """
 
-source_info_path_update: dict[str, str] | None = dict(suffix="ave")
+source_info_path_update: dict[str, str] | None = None
```
larsoner, line 2092: Yikes, looks like `None` was documented as an option but never actually supported based on the code changes 🤭. Lines like `bids_path.copy().update(**cfg.source_info_path_update)` would have just died...

In any case, now that `None` is the default we can also remove this line:

https://github.com/mne-tools/mne-bids-pipeline/blob/d8c1e4b683f45567f0c1aad59ab87c89355bf34b/mne_bids_pipeline/_config_import.py#L347

mne_bids_pipeline/_config.py
```diff
@@ -2089,22 +2089,31 @@ def noise_cov(bids_path):
...
-certain situations you may wish to use a different `Info`.
-
+When computing the forward and inverse solutions, it is important to
+provide the `mne.Info` object from the data on which the noise covariance was
+computed, to avoid problems resulting from mismatching ranks.
 This parameter allows you to explicitly specify from which file to retrieve the
 `mne.Info` object. Use this parameter to supply a dictionary to
 `BIDSPath.update()` during the forward and inverse processing steps.
+If set to `None` (default), the info will be retrieved either from the raw
+file specified in `noise_cov`, or the cleaned epochs
```
larsoner, line 2101:
```suggestion
file specified in `noise_cov`, or the cleaned evoked
```

mne_bids_pipeline/steps/source/_05_make_inverse.py
```diff
@@ -48,12 +48,25 @@ def get_input_fnames_inverse(
...
+            # XXX is this the right solution also for noise_cov = 'ad-hoc'?
+    else:
+        source_info_path_update = cfg.source_info_path_update
+    in_files["info"] = bids_path.copy().update(**source_info_path_update)
     in_files["forward"] = bids_path.copy().update(suffix="fwd")
     if cfg.noise_cov != "ad-hoc":
         in_files["cov"] = get_noise_cov_bids_path(
             cfg=cfg, subject=subject, session=session
         )
+        dict(suffix="ave")
```
larsoner, line 69: cruft I think?
```suggestion
```

larsoner (commented):
Nice, just some minor cleanups!

mne_bids_pipeline/_config.py
SophieHerbst: done

larsoner (approved):
Marking for merge-when-green, thanks in advance @SophieHerbst !

PR #926, opened 2024-04-18
Don't perform ICA cleaning on raw data by default when epochs are present

### LINKED ISSUE #925: Don't perform ICA cleaning on raw data by default
I use ICA to clean my epochs. The `apply_ica` step now (I believe since #899?) cleans the raw data using ICA and adds the results to the report. I don't care for this; I only want to clean my epochs. The cleaned raws clutter my report. Could we change this behavior please?

cc @SophieHerbst

hoechenberger:
Fixes #925

### Before merging …

### DIFF
--- a/docs/source/v1.9.md.inc
+++ b/docs/source/v1.9.md.inc
@@ -2,15 +2,22 @@
 
 ### :new: New features & enhancements
 
-- Added number of subject to `sub-average` report (#902, #910 by @SophieHerbst)
+- Added number of subjects to `sub-average` report (#902, #910 by @SophieHerbst)
 - The type annotations in the default configuration file are now easier to read: We
   replaced `Union[X, Y]` with `X | Y` and `Optional[X]` with `X | None`. (#908, #911 by @hoechenberger)
+- Added a new configuration option [`spatial_filter_raw`][mne_bids_pipeline._config.spatial_filter_raw]
+  to control whether to create cleaned raw data via ICA or SSP. Previously, we'd only clean epochs.
+  (#840, #926 by @larsoner and @hoechenberger)
 
 ### :warning: Behavior changes
 
 - All ICA HTML reports have been consolidated in the standard subject `*_report.html`
   file instead of producing separate files (#899 by @larsoner).
-- Changed default for `source_info_path_update` to `None`.  In `_04_make_forward.py`
+- When using ICA or SSP with resting-state data, we now automatically produce cleaned raw data files. This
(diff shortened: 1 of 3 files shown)


hoechenberger:
I'm not sure this is the right approach (in terms of, where to do the conditional branching)… @larsoner Could you kindly take a look?

mne_bids_pipeline/steps/preprocessing/_08a_apply_ica.py
```diff
@@ -196,7 +196,12 @@ def apply_ica_raw(
     run: str,
     task: str | None,
     in_files: dict,
-) -> dict:
+) -> dict | None:
+    if cfg.spatial_filter_raw is None and not (cfg.task == "rest" or cfg.task_is_rest):
+        return
+    elif cfg.spatial_filter_raw is False:
+        return
```
larsoner, line 203: Rather than doing it at this level I would do it at the level of the `main` below. I'll push a commit

docs/source/v1.9.md.inc
```diff
@@ -2,15 +2,22 @@
...
+  (#840, #926 by @larsoner and @hoechenberger)
 
 ### :warning: Behavior changes
 
 - All ICA HTML reports have been consolidated in the standard subject `*_report.html`
   file instead of producing separate files (#899 by @larsoner).
-- Changed default for `source_info_path_update` to `None`.  In `_04_make_forward.py`
+- When using ICA or SSP with resting-state data, we now automatically produce cleaned raw data files. This
+  behavior can be controlled via the new [`spatial_filter_raw`][mne_bids_pipeline._config.spatial_filter_raw]
+  configuration option. (#840, #926 by @larsoner and @hoechenberger)
```
larsoner, line 18: The introduction of writing raw data was actually in 1.6 / #840. So the behavior change here is actually to *stop* writing files by default. So maybe we should leave the default as `True` for backward compat. Then this PR is just to add the ability to have `False | None`.

Personally I think `True` is the most reasonable default anyway. If you ever want to re-epoch your data (I often do!) then you'll want those preprocessed raw files. And you can cross-check individual runs against the produced epochs, it's more complete to have raw data report at each stage, etc. even though it makes the report longer.

larsoner (changes requested):

hoechenberger:
Thanks for your help here, @larsoner, this is much appreciated!

I'd like to hear @SophieHerbst's input as well, esp. since she said they were struggling with disk space usage and all…

PR #931, opened 2024-04-23
Fix report section on original raw data (didn't contain bad channels and subject or experimenter name)

### LINKED ISSUE #930: Report section on original raw data doesn't contain bads
In my BIDS data, 3 channels are marked as bad. These show up correctly in the HTML repr for the filtered and cleaned raw, but not for the original raw. The content of the generated `*_bads.tsv` file is:

```
name	reason
P8	pre-existing (before MNE-BIDS-pipeline was run)
T7	pre-existing (before MNE-BIDS-pipeline was run)
TP10	pre-existing (before MNE-BIDS-pipeline was run)
```

Screenshot 2024-04-23 at 13 17 29

Screenshot 2024-04-23 at 13 17 34

I tried to understand what was going on but so far I don't really have a clue.

hoechenberger:
Fixes #930

### Before merging …

### DIFF
--- a/mne_bids_pipeline/_report.py
+++ b/mne_bids_pipeline/_report.py
@@ -848,9 +848,9 @@ def _add_raw(
     cfg: SimpleNamespace,
     report: mne.report.Report,
     bids_path_in: BIDSPath,
+    raw: BaseRaw,
     title: str,
     tags: tuple = (),
-    raw: BaseRaw | None = None,
     extra_html: str | None = None,
 ):
     if bids_path_in.run is not None:
@@ -865,7 +865,7 @@ def _add_raw(
     tags = ("raw", f"run-{bids_path_in.run}") + tags
     with mne.use_log_level("error"):
         report.add_raw(
-            raw=raw or bids_path_in,
+            raw=raw,
             title=title,
             butterfly=5,
             psd=plot_raw_psd,
--- a/mne_bids_pipeline/steps/preprocessing/_01_data_quality.py
+++ b/mne_bids_pipeline/steps/preprocessing/_01_data_quality.py
@@ -180,7 +180,7 @@ def assess_data_quality(
     tsv_data.to_csv(out_files["bads_tsv"], sep="\t", index=False)
 
     # Report
-    # Restore bads to their original state before generating the report
+    # Restore bads to their original state so they will show up in the report
     raw.info["bads"] = preexisting_bads
 
     with _open_report(
--- a/mne_bids_pipeline/steps/preprocessing/_01_data_quality.py
+++ b/mne_bids_pipeline/steps/preprocessing/_01_data_quality.py
@@ -173,7 +173,7 @@ def assess_data_quality(
             if ch in bads_for_tsv:
                 continue
             bads_for_tsv.append(ch)
-            reasons.append(["pre-existing (before MNE-BIDS-pipeline was run)"])
+            reasons.append("pre-existing (before MNE-BIDS-pipeline was run)")
 
     tsv_data = pd.DataFrame(dict(name=bads_for_tsv, reason=reasons))
     tsv_data = tsv_data.sort_values(by="name")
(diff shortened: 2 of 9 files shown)


mne_bids_pipeline/steps/preprocessing/_01_data_quality.py
```diff
@@ -149,27 +149,40 @@ def assess_data_quality(
...
+                continue
+            bads_for_tsv.append(ch)
+            reasons.append(["pre-existing (before MNE-BIDS-pipeline was run)"])
 
     tsv_data = pd.DataFrame(dict(name=bads_for_tsv, reason=reasons))
     tsv_data = tsv_data.sort_values(by="name")
     tsv_data.to_csv(out_files["bads_tsv"], sep="\t", index=False)
 
     # Report
+    # Restore bads to their original state
```
hoechenberger, line 183:
```suggestion
    # Restore bads to their original state before generating the report
```

mne_bids_pipeline/steps/preprocessing/_01_data_quality.py
```diff
@@ -149,27 +149,40 @@ def assess_data_quality(
...
+                continue
+            bads_for_tsv.append(ch)
+            reasons.append("pre-existing (before MNE-BIDS-pipeline was run)")
 
     tsv_data = pd.DataFrame(dict(name=bads_for_tsv, reason=reasons))
     tsv_data = tsv_data.sort_values(by="name")
     tsv_data.to_csv(out_files["bads_tsv"], sep="\t", index=False)
 
     # Report
+    # Restore bads to their original state before generating the report
```
hoechenberger, line 183:
```suggestion
    # Restore bads to their original state so they will show up in the report
```

mne_bids_pipeline/_report.py
```diff
@@ -848,9 +848,9 @@ def _add_raw(
     cfg: SimpleNamespace,
     report: mne.report.Report,
     bids_path_in: BIDSPath,
+    raw: BaseRaw,
```
hoechenberger, line 851: This change should avoid the problem we saw here, as it now forces us to provide pre-loaded raw data (with hopefully the correct metadata set). previously this was optional. see also the changes below

hoechenberger:
Circle fails for ds000246 due to outdated cache

hoechenberger:
> I'll fix the caches in #929

Thank you!

larsoner:
> Should we only set the pre-existing bads when generating the HTML repr for the original data, or should we set the pre-existing bads + all bads from the data quality checks?

To me the optimal thing to do here I think is to have two report entries:

1. One which is the data as it comes originally from the dataset author with bads from the sidecar. It tells you what you start with before doing anything to it with the pipeline
2. Another which is the data after our autobads (this section only needs to exist if we do some autobad detection). It tells you what the automated approaches did to the data

To me it's okay for brevity to have (1) have all the nice raw data plots, and have (2) maybe have less -- like just a list of channels removed or something? I think we're already kind of close to this organization:

image

But from that data quality plot it's not immediately clear which channels were added as bad, if any. Even just a HTML entry that lists the bads (when bad enabled) and flats (when flat enabled) would be a nice help

larsoner:
... and organizationally I think it might make sense to call that section in the sidebar `Raw (autobad)` or similar. It makes it clear the parentheticals from original->autobad->maxwell->filtered->cleaned. "Data Quality" is an outlier at the moment in that it uses the step name

hoechenberger:
Thanks for sharing your ideas, they resonate well with me!

I believe then at least the bits of code that this PR touches should be good! And the rest could be addressed in future PRs

larsoner:
I merged `main` into the branch, you should be able to add to the changelog now and it should come back green

hoechenberger:
Thanks @larsoner, I pushed changelog entries and enabled auto-merge

PR #938, opened 2024-04-24
Save ECG and EOG evokeds created during ICA artifact detection to disk

hoechenberger:
I needed access to the EOG evoked for diagnostic purposes, but figured we didn't save this data anywhere … This PR now saves the evokeds of ECG and EOG data. I considered saving the epochs instead, but at least for my usecase, I would always want to average first anyway…

### Before merging …

### DIFF
--- a/docs/source/v1.9.md.inc
+++ b/docs/source/v1.9.md.inc
@@ -10,6 +10,7 @@
   [`epochs_metadata_tmax`][mne_bids_pipeline._config.epochs_metadata_tmax]. (#873 by @hoechenberger)
 - If you requested processing of non-existing subjects, we will now provide a more helpful error message. (#928 by @hoechenberger)
 - We improved the logging output for automnated epochs rejection and cleaning via ICA and SSP. (#936 by @hoechenberger)
+- ECG and EOG signals created during ICA artifact detection are now saved to disk. (#938 by @hoechenberger)
 
 ### :warning: Behavior changes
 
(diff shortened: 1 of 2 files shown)


larsoner:
Yeah my first impulse was to say "why not save the Epochs" but then realized that I, too, always work with the averages :laughing: 

So okay for me to save the evokeds for now and add epochs later if we want them!

larsoner:
I think another dataset will come back out of date based on a failure I saw recently -- if/when it does can you try bumping the cache number (e.g., `-3` to `-4` or whatever) on CircleCI? 246 will probably also be broken but we can just ignore it for now.

larsoner:
Yeah looks like 247 needs a new cache number bump

https://app.circleci.com/pipelines/github/mne-tools/mne-bids-pipeline/4527/workflows/a3292c92-606c-4f66-ba2a-7b91ddced838/jobs/65426

larsoner:
I'll be surprised if anything works at the moment given https://www.githubstatus.com/ today but I'll restart CIs once things settle a bit

hoechenberger:
@larsoner I just found myself wanting to do an ICA properties plot and this one requires epochs… 😶‍🌫️🫣

larsoner:
Hah, feel free to change it to Epochs instead

hoechenberger:
I decided that `plot_components()` was sufficient after all. Let's keep the averages until somebody complains :)

larsoner:
Just 246 that's failing so in it goes, thanks @hoechenberger !

PR #950, opened 2024-05-15
Project CSP patterns to source

SophieHerbst:
Hello! I am trying to implement Britta's and Jean-Remy's [best of both world approach ](https://github.com/britta-wstnr/source_decoding) to project the CSP decoding patterns to source. 
I am able to do it on my data in my on code, but I am struggling to extract the patterns and coefficients in the pipeline code. 
On the one hand, I am somewhat lost in the terminology between patterns, coefficients, and filters. 

On the other hand, when using Britta's code, I get an error message:
```
This estimator does not have a patterns_ attribute:
LogReg(n_jobs=1, random_state=42, solver='liblinear')
```

Can you help me understand the difference between the decoding pipeline we employ and Britta's?

```
clf = make_pipeline(StandardScaler(), LinearModel(LogisticRegression(
        solver='liblinear')))
```

### DIFF
--- a/mne_bids_pipeline/steps/sensor/_05_decoding_csp.py
+++ b/mne_bids_pipeline/steps/sensor/_05_decoding_csp.py
@@ -7,10 +7,11 @@ import matplotlib.transforms
 import mne
 import numpy as np
 import pandas as pd
-from mne.decoding import CSP
+from mne.decoding import CSP, LinearModel
 from mne_bids import BIDSPath
 from sklearn.model_selection import StratifiedKFold, cross_val_score
 from sklearn.pipeline import make_pipeline
+from sklearn.linear_model import LogisticRegression
 
 from ..._config_utils import (
     _bids_kwargs,
@@ -239,8 +240,12 @@ def one_subject_decoding(
         epochs_filt, y = prepare_epochs_and_y(
             epochs=epochs, contrast=contrast, fmin=fmin, fmax=fmax, cfg=cfg
         )
+
         # Get the data for all time points
         X = epochs_filt.get_data()
+
+        
+        
 
         cv_scores = cross_val_score(
             estimator=clf,
@@ -252,6 +257,22 @@ def one_subject_decoding(
         )
         freq_decoding_table.loc[idx, "mean_crossval_score"] = cv_scores.mean()
         freq_decoding_table.at[idx, "scores"] = cv_scores
+
+        # PATTERNS
+        csp.fit_transform(X, y)
+        sensor_pattern_csp = csp.patterns_
+
+        # COEFS
+        clf.fit(X, y)
+        weights_csp = get_coef(clf, 'patterns_', inverse_transform=True)
+
+        # save scores
+        # XXX right now this saves in working directory
+        csp_fname = cond1 + cond2 + str(fmin) + str(fmax)
+
+        np.save(csp_fname + '_patterns', sensor_pattern_csp)
+        np.save(csp_fname + '_weights', weights_csp)
+
 
     # Loop over times x frequencies
     #
@@ -306,6 +327,7 @@ def one_subject_decoding(
         # Crop data to the time window of interest
         if tmax is not None:  # avoid warnings about outside the interval
             tmax = min(tmax, epochs_filt.times[-1])
+
         X = epochs_filt.crop(tmin, tmax).get_data()
         del epochs_filt
         cv_scores = cross_val_score(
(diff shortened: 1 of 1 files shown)


britta-wstnr:
Hey @SophieHerbst, thanks for pinging me via email.
I'd have to look into this a bit more - I'll put it on my agenda but please do ping me again if I don't move on it for a while, my to do list is a bit populated right now.
One first thing to share: please note that the analyses for this paper were done quite a long time ago (around 2016, 2017 I would say). So this might also come down to changes in versions. 
But I will have a more detailed look!

mne_bids_pipeline/steps/sensor/_05_decoding_csp.py
```diff
@@ -253,6 +254,21 @@ def _fmt_contrast(cond1, cond2, fmin, fmax, freq_range_name, tmin=None, tmax=Non
         freq_decoding_table.loc[idx, "mean_crossval_score"] = cv_scores.mean()
         freq_decoding_table.at[idx, "scores"] = cv_scores
 
+        # PATTERNS
+        csp.fit_transform(X, y)
+        sensor_pattern_csp = csp.patterns_
+
+        # COEFS
+        clf.fit(X, y)
+        weights_csp = mne.decoding.get_coef(clf, "patterns_", inverse_transform=True)
```
larsoner, line 263: Typically this is all I use. Why also take the `csp.patterns_`? `csp.fit(X, y)` results anywhere else so it's a bit weird to me to add them here. It seems like we should just use these `weights_csp`. And maybe we should call them `clf_patterns_` because they're really the patterns inverse transformed from the CSP all the way back through the other steps (e.g., PCA, sensor scaling)?

larsoner:
Yes so far I have been projecting CSP `patterns_` to source space separately, the pipeline should do it for us. But given the pipeline runs all sensor analyses then all source analyses, we'll need a new source step whose job it is (just) to visualize the CSP patterns in source space

SophieHerbst:
I think I found it.
I needed to add a LinearModel wrapper in the clf

SophieHerbst:
(am trying to push, but need to resolve a conflict)

mne_bids_pipeline/steps/sensor/_05_decoding_csp.py
SophieHerbst: that is exactly where I am lost in the terminology.. 

my approach was to save them in this step to pick them up in a separate source projection step later

SophieHerbst: I wanted to save both, because in @britta-wstnr's code they are combined during beamforming:

```
stc_csp = beamform_components(weights_csp, sensor_pattern_csp, spat_filter,
                                fwd, multipliers=multiplier)
```

britta-wstnr: The idea here is to combine the coefficients and the CSP components. The (Haufe et al.) patterns are not easily beamform-able as they do not represent classical sink-source patterns anymore but refer to the decoding classes. Using the CSP components is a work-around for that.

SophieHerbst:
With the above code, I get an error in _05_decoding_csp.py, at csp.fit_transform()
Do you have any idea on how to fix this?


```
A critical error occurred. The error message was: SVD did not converge

Aborting pipeline run. The traceback is:

  File "/home/sh254795/anaconda3/envs/mne/lib/python3.10/site-packages/mne_bids_pipeline/steps/sensor/_05_decoding_csp.py", line 354, in one_subject_decoding
    csp.fit_transform(X, y)
  File "/home/sh254795/anaconda3/envs/mne/lib/python3.10/site-packages/mne/decoding/csp.py", line 250, in fit_transform
    return super().fit_transform(X, y=y, **fit_params)
  File "/home/sh254795/anaconda3/envs/mne/lib/python3.10/site-packages/mne/decoding/mixin.py", line 33, in fit_transform
    return self.fit(X, y, **fit_params).transform(X)
  File "/home/sh254795/anaconda3/envs/mne/lib/python3.10/site-packages/mne/decoding/csp.py", line 197, in fit
    self.patterns_ = pinv(eigen_vectors)
  File "/home/sh254795/anaconda3/envs/mne/lib/python3.10/site-packages/mne/fixes.py", line 869, in pinv
    u, s, vh = np.linalg.svd(a, full_matrices=False)
  File "<__array_function__ internals>", line 180, in svd
  File "/home/sh254795/anaconda3/envs/mne/lib/python3.10/site-packages/numpy/linalg/linalg.py", line 1657, in svd
    u, s, vh = gufunc(a, signature=signature, extobj=extobj)
  File "/home/sh254795/anaconda3/envs/mne/lib/python3.10/site-packages/numpy/linalg/linalg.py", line 98, in _raise_linalgerror_svd_nonconvergence
    raise LinAlgError("SVD did not converge")
```

larsoner:
I sometimes hit these and try to open issues on the relevant upstream trackers. In this case it's probably an OpenBLAS bug. You can try setting OMP_NUM_THREADS=1 in your env, sometimes it fixes it. In principle using `mne.fixes._safe_svd` could fix it but you'd have to be careful about broadcasting (it only works on arrays of shape `(M, N)` whereas `np.linalg.svd` works the last two dimensions of any array, i.e., those of shape `(..., M, N)`.

Can you run the pipeline with `--pdb`, and do a `np.save` of `a` and print some info like:
```
(pdb) p np.savez("problem.npz", a)
(pdb) p gufunc
...
(pdb) p signature
(pdb) p extobj
```
and then also do `(pdb) p np.show_config()` and look to see where BLAS/LAPACK info exists, like:
```
>>> np.show_config()
Build Dependencies:
  blas:
    ...
    openblas configuration: OpenBLAS 0.3.27  USE64BITINT DYNAMIC_ARCH NO_AFFINITY
```
and upload the `.npz` and outputs here? Then I can try to replicate and escalate to the right place

larsoner:
... there is also a chance that upgrading your OpenBLAS / NumPy could fix the issue. Or switching away from using MKL if you're using it

SophieHerbst:
Thanks @larsoner 
The system specs are: 
(mne) sh254795@narval:~$ cat /sys/devices/cpu/caps/pmu_name
skylake
(mne) sh254795@narval:~$ cat /proc/cpuinfo | grep 'model name' | uniq 
model name      : Intel(R) Xeon(R) Gold 5218 CPU @ 2.30GHz
(mne) sh254795@narval:~$ mne sys_info | grep numpy
├☑ numpy             1.23.5 (OpenBLAS 0.3.21 with 64 threads)

I updated numpy and am trying to run it again right now.

PR #951, opened 2024-05-18
MAINT: Cleaner traceback on error

larsoner:
No need for changelog update I think, just shortens the traceback and fails faster when `cross_val_score` fails for CSP just like it does [for other decoding types](https://github.com/mne-tools/mne-bids-pipeline/blob/0a870021e6ad61b2d55d1ddd8dbf6eb83bb1ba9d/mne_bids_pipeline/steps/sensor/_02_decoding_full_epochs.py#L153).


larsoner:
Error is unrelated but I'll investigate to see if I can figure out where it's coming from. Probably a `plt.close(fig)` missing somewhere!

docs/templates/python/material/attribute.html
```diff
@@ -1,9 +1,10 @@
 {# Modified from https://github.com/mkdocstrings/python/blob/master/src/mkdocstrings_handlers/python/templates/material/_base/attribute.html #}
+{# Updated 2024/05/20. See "START NEW CODE" for block that is new. #}
```
larsoner, line 2: So the real necessary change here was to `git mv` the file. But while I was at it I also took the latest version from the URL above. We'll see if it renders okay. If not we can revert

larsoner:
Comparing:

- https://mne.tools/mne-bids-pipeline/dev/settings/sensor/contrasts.html
- https://output.circle-artifacts.com/output/job/79f31986-5cbd-4fb3-a141-f302839650c6/artifacts/0/site/settings/sensor/contrasts.html

I think things look okay, so ready for review/merge from my end @hoechenberger !

mne_bids_pipeline/steps/preprocessing/_09_ptp_reject.py
```diff
@@ -204,16 +206,16 @@ def drop_ptp(
...
-                    orientation="horizontal", aspect="auto", show=False
-                ),
+                fig=fig,
                 title=f"{kind} cleaning",
                 caption=caption,
                 section=title,
                 tags=tags,
                 replace=True,
             )
+            plt.close(fig)
```
larsoner, line 218: Need this one for green CIs and this PR is pretty trivial I think, so I'll go ahead and merge. Happy to make changes if you see issues though @hoechenberger

PR #957, opened 2024-06-12
[MAINT] Switch to trusted publishing

tsbinns:
Discussed with @larsoner: https://github.com/mne-tools/mne-connectivity/pull/200#issuecomment-2163383202

Switches releases to support PyPI trusted publishing.


larsoner:
@hoechenberger I set up trusted publishing at the PyPI end so this should be good to go!

We'll find out if it works once we push another release

hoechenberger (approved):

PR #970, opened 2024-07-25
MAINT: Use Python CircleCI image

larsoner:
Using a machine runner in MNE-Python was a bit faster, let's try it here. Our builds currently take ~40m30s, let's see if this is faster!


hoechenberger:
@larsoner Unrelated, but could we maybe explore options to use GHA? The main thing that's kept us from moving was the limitation regarding viewing the artifacts (HTML reports) directly in the browser without having to download them, right?

larsoner:
> The main thing that's kept us from moving was the limitation regarding viewing the artifacts (HTML reports) directly in the browser without having to download them, right?

That was one, and I wouldn't want to have to unzip and view locally. So if that's still the case (and I think it is), -1 from me for transitioning.

At some point CircleCI allowed more parallel jobs. Looks like the limit for GHA [is now 20](https://docs.github.com/en/actions/administering-github-actions/usage-limits-billing-and-administration#usage-limits) (I think?) so maybe this isn't as much of a limitation? But if that's at an organization level (and it used to be IIRC) that would be potentially quite bad because you could have a PR in MNE-Python waiting about an hour just to get started, or have one here held up for an hour, etc.

Lastly all of the caching etc. would have to be migrated, and that's all extra work for something that might or might not end up working as well (if at all).

What's the expected benefit of GHA?

hoechenberger:
> What's the expected benefit of GHA?

Really just more consistency across our projects, and potentially the ability to test on more platforms. Nothing urgent though

larsoner:
FWIW we're fairly consistent with MNE-Python (and mne-connectivity, mne-nirs, etc.) already in that CircleCI is the CI that does the doc build and pushed to gh-pages. And we do already test one dataset on other platforms using GHA actually! I don't think we'd want to do *all datasets* on other platforms because that would take a ton of CI time.

PR #977, opened 2024-07-30
don't fail if emptyroom is missing cHPI channels

### LINKED ISSUE #974: pipeline fails if emptyroom data is missing cHPI
When using `mf_filter_chpi=True` (or `mf_filter_chpi=None`), MNEBP will attempt to filter cHPI, but this causes a crash when the emptyroom data don't have cHPI channels enabled. In such cases what is the best approach?

- should `mne.preprocessing.maxwell_filter_prepare_emptyroom` try to copy the cHPI channels from the data file? That seems... complicated, given the unequal time domains of the two files.
- Should MNEBP be lenient for emptyroom files, maybe by calling `filter_chpi(..., allow_line_only=True)`?  (I can confirm that hacking this in does allow the pipeline step to succeed)
- should `mne.filter_chpi` be refactored to make its `recon` operator available, such that the `recon` could be computed on the raw reference data and applied to the emptyroom raw?

drammock:
closes #974 

### Before merging …


docs/source/v1.10.md.inc
```diff
@@ -12,9 +12,11 @@
 
 [//]: # (- Whatever (#000 by @whoever))
 
-[//]: # (### :bug: Bug fixes)
+### :bug: Bug fixes
 
-[//]: # (- Whatever (#000 by @whoever))
+- When running the pipeline with [`mf_filter_chpi`][mne_bids_pipeline._config. mf_filter_chpi] enabled,
```
larsoner, line 17: Unless the space was added intentionally I assume it shouldn't be there. Will have to look at the website once it's built to see if it matters. (It may or may not error, I'll look by eye.)
```suggestion
- When running the pipeline with [`mf_filter_chpi`][mne_bids_pipeline._config.mf_filter_chpi] enabled,
```

larsoner (commented):
Looking at:
```
$ git grep mf_filter
...
mne_bids_pipeline/tests/configs/config_ds004229.py:mf_filter_chpi = False  # for speed, not needed as we low-pass anyway
```
maybe having this `True` would have caught the error on `main`. Do you want to check? If not I can try quickly.

Either way might be good to set to True assuming in only adds a reasonable amount of time on CircleCI (e.g., < 1 minute more is reasonable, > 3 minutes probably isn't).

docs/source/v1.10.md.inc
larsoner: Indeed it's problematic

https://output.circle-artifacts.com/output/job/62c1acc3-c225-4dcb-bdf8-94b0b70b9328/artifacts/0/site/changes.html

Screenshot 2024-07-30 at 7 16 17 PM

@hoechenberger any way to make these links loudly rather than silently fail? I'm not so familiar with the `mkdocs` settings that might be the equivalent of `-nW` in sphinx (nitpicky + warning-as-error) that would have caught this

drammock: I put in the space because I was copying the format of the first bugfix in the 1.9 changelog (which is also not rendering, now that I look at the site: https://mne.tools/mne-bids-pipeline/stable/changes.html#bug-fixes)

thought it was odd but assumed it was some fiddly requirement of mkdocs

larsoner:
> maybe having this True would have caught the error on main. Do you want to check? If not I can try quickly.

Locally
```
rm -Rf ~/mne_data/derivatives/mne-bids-pipeline/ds004229/ && time pytest mne_bids_pipeline/ -k 4229
```
took `5m50.472s`. Setting `mf_filter_chpi = True` on `main` did emit an error!

I also realized that we should filter cHPI *before* performing movement compensation, so I changed that order and added it in the changelog as a bugfix.

Marking for merge-when-green, thanks in advance @drammock !

drammock:
> took `5m50.472s`. Setting `mf_filter_chpi = True` on `main` did emit an error!

heh, you beat me to it, was just about to check for that failure on main

PR #987, opened 2024-08-25
Allow separate MRIs per session (for longitudinal studies)

drammock:
WIP. Could use some advice about the best approach here; so far it *should* work (but yet untested!) for my use case but definitely won't work for the "normal" case that is currently supported on `main`. See the (small) diff for a prominent comment explaining the issue.

### Before merging …

### DIFF
--- a/mne_bids_pipeline/steps/freesurfer/_01_recon_all.py
+++ b/mne_bids_pipeline/steps/freesurfer/_01_recon_all.py
@@ -11,25 +11,33 @@ from pathlib import Path
 
 from mne.utils import run_subprocess
 
-from mne_bids_pipeline._config_utils import get_fs_subjects_dir, get_subjects
+from mne_bids_pipeline._config_utils import (
+    get_fs_subjects_dir,
+    get_sessions,
+    get_subjects,
+)
 from mne_bids_pipeline._logging import gen_log_kwargs, logger
 from mne_bids_pipeline._parallel import get_parallel_backend, parallel_func
 
 fs_bids_app = Path(__file__).parent / "contrib" / "run.py"
 
 
-def run_recon(root_dir, subject, fs_bids_app, subjects_dir) -> None:
+def run_recon(root_dir, subject, fs_bids_app, subjects_dir, session=None) -> None:
     subj_dir = subjects_dir / f"sub-{subject}"
+    sub_ses = f"Subject {subject}"
+    if session is not None:
+        subj_dir = subj_dir / f"ses-{session}"
+        sub_ses = f"{sub_ses} session {session}"
 
     if subj_dir.exists():
         msg = (
-            f"Subject {subject} is already present. Please delete the "
+            f"Recon for {sub_ses} is already present. Please delete the "
             f"directory if you want to recompute."
         )
         logger.info(**gen_log_kwargs(message=msg))
         return
     msg = (
-        "Running recon-all on subject {subject}. This will take "
+        "Running recon-all on {sub_ses}. This will take "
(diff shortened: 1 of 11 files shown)


mne_bids_pipeline/steps/freesurfer/_01_recon_all.py
```diff
@@ -77,21 +87,34 @@ def main(*, config) -> None:
...
     subjects_dir = Path(get_fs_subjects_dir(config))
     subjects_dir.mkdir(parents=True, exist_ok=True)
 
     with get_parallel_backend(config.exec_params):
         parallel, run_func = parallel_func(run_recon, exec_params=config.exec_params)
         parallel(
-            run_func(root_dir, subject, fs_bids_app, subjects_dir)
+            # TODO How can we infer if inner `for session in sessions` loop is needed?
+            #      Should we specify in `config`? As written, it will *only* work for
+            #      cases where subjects_dir has this layout:
```
larsoner, line 106: I think it's okay to use some filesystem inspection using whatever conditionals you need to construct the subj-session pairs, and then call the `parallel`. So something like
```
subj_session = list()
for subject in subjects:
    if mri_is_root_level(subject):  # TODO: Need to write this helper to do the filesystem inspection part!
        subj_session.append((subject, None))
    else:
        subj_session.extend((subject, session) for session in sessions)
parallel(run_func(...) for subject, session in subj_session)
```
where `session=None` would mean "root-level not session specific". Does that make sense? If not maybe I misunderstand the problem.

drammock:
eh, this is not going to work as-is. There is an assumption baked-in in too many places that there's only a single folder level between `subjects_dir` and the various freesurfer files.  I'll explore making the folders something like `sub-116_ses-a/` rather than `sub-116/ses-a/`

larsoner:
Makes sense

drammock:
I think this is ready for review. Probably needs some addition to the docs so that folks other than me know how to actually make this work... I can add that along with whatever other changes y'all want.

larsoner:
I think next you'd want to add `docs/source/v1.9.md.inc` to say that per-session MRIs are now supported.

Any idea if there is a dataset on openneuro that we could add to the test suite that would hit the new code path?

drammock:
> I think next you'd want to add `docs/source/v1.9.md.inc` to say that per-session MRIs are now supported.

Done.  As for adding more thorough documentation elsewhere, I can't actually see where I would add it (since it doesn't involve a config value, and there's no dataset to add that could serve as an example).

> Any idea if there is a dataset on openneuro that we could add to the test suite that would hit the new code path?

I doubt it; OpenNeuro has zero datasets with modality=MEG and study_type=Longitudinal :(

drammock:
Just checked the EEG Longitudinal studies too; none of them involve source reconstruction, only 1 even includes MRI at all, and it's only at 1 timepoint (https://openneuro.org/datasets/ds004024/versions/1.0.1)

hoechenberger:
yes seems like it! do you want to take a look? just follow what i did in my previous bug reports :)

larsoner:
https://github.com/OpenNeuroOrg/openneuro/issues/3145 and https://github.com/hoechenberger/openneuro-py/pull/187, openneuro folks usually resolve quickly so I say let's wait a day or two then restart. If they don't respond by then let's just merge.

hoechenberger:
Issue has already been addressed upstream by OpenNeuro, restarting CI!

hoechenberger:
@larsoner This is green. i won't have time to review, please move along if you're happy

larsoner:
Thanks @drammock !

PR #1000, opened 2024-09-26
allow missing sessions

### LINKED ISSUE #992: can pipeline be more robust to handling missing data?
another "I have a longitudinal study" question here. I have some subjects with missing timepoints (AKA missing sessions).  I want the pipeline to look at the config file (which says `subjects = "all"` and `sessions = ["a", "b", "c"]`) and then *look at the bids data tree* and process all the subjectXsession combinations that are present, without crashing because *subject 101 is missing session "b"*.  In practice, such errors take the form of:

```
┌────────┬ init/_02_find_empty_room ──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
│13:31:56│ 🐛 sub-101 ses-b A critical error occurred. The error message was: Missing file: /redacted/redacted/bids-data/sub-101/ses-b/meg/sub-101_ses-b_task-redacted_split-01
```

which is confusing; it errors when looking for a split partial file, when the real problem is that there isn't even a `.../sub-101/ses-b/` directory.

#### Proposed solutions

1. just make the error message better, and do something complicated in the config where I figure out which subject-session combos I have, and run the pipeline multiple times (e.g. once for each session, with different lists of subjects each time)

2. blithely add an `allow_missing=True` here, and hope that it doesn't have catastrophic downstream effects:

https://github.com/mne-tools/mne-bids-pipeline/blob/61f29616f5e3c2e246365f9568ae2c168616617c/mne_bids_pipeline/steps/init/_02_find_empty_room.py#L45

3. it seems like there are a lot of places where we do `for subject in subjects` and `for session in sessions` where `subjects` and `sessions` are taken straight from the config file each time.  So it won't work to do a simple `Path(...).exists()` type of check to see if the entire session folder is missing, right about here:

    https://github.com/mne-tools/mne-bids-pipeline/blob/61f29616f5e3c2e246365f9568ae2c168616617c/mne_bids_pipeline/steps/init/_01_init_derivatives_dir.py#L79-L81

    But we could do that check, but then store in a private config location the subjectXsession combos that are actually present, and use that info to get subjects/sessions to loop over, throughout the codebase.

Open to other ideas!

drammock:
WIP. Closes #992 

### Before merging …

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -16,7 +16,7 @@ jobs:
     - uses: actions/checkout@v4
     - uses: actions/setup-python@v5
       with:
-        python-version: "3.11"
+        python-version: "3.12"
     - run: pip install --upgrade pip
     - run: pip install -ve .[tests] codespell tomli
     - run: make codespell-error
@@ -49,7 +49,7 @@ jobs:
           pyvista: false
       - uses: actions/setup-python@v5
         with:
-          python-version: "3.11"  # no "multidict" wheels on 3.12 yet
+          python-version: "3.12"
       - run: pip install -ve .[tests]
       - uses: actions/cache@v4
         with:
--- a/docs/source/v1.10.md.inc
+++ b/docs/source/v1.10.md.inc
@@ -3,7 +3,7 @@
 ### :new: New features & enhancements
 
 - It is now possible to use separate MRIs for each session within a subject, as in longitudinal studies. This is achieved by creating separate "subject" folders for each subject-session combination, with the naming convention `sub-XXX_ses-YYY`, in the freesurfer `SUBJECTS_DIR`. (#987 by @drammock)
-- New boolean config option `allow_missing_sessions` allows the pipeline to continue when not all sessions are present for all subjects. (#1000 by @drammock)
+- New config option [`allow_missing_sessions`][mne_bids_pipeline._config. allow_missing_sessions] allows to continue when not all sessions are present for all subjects. (#1000 by @drammock)
 
 [//]: # (### :warning: Behavior changes)
 
--- a/docs/source/v1.10.md.inc
+++ b/docs/source/v1.10.md.inc
@@ -3,7 +3,7 @@
 ### :new: New features & enhancements
 
 - It is now possible to use separate MRIs for each session within a subject, as in longitudinal studies. This is achieved by creating separate "subject" folders for each subject-session combination, with the naming convention `sub-XXX_ses-YYY`, in the freesurfer `SUBJECTS_DIR`. (#987 by @drammock)
-- New config option [`allow_missing_sessions`][mne_bids_pipeline._config. allow_missing_sessions] allows to continue when not all sessions are present for all subjects. (#1000 by @drammock)
+- New config option [`allow_missing_sessions`][mne_bids_pipeline._config.allow_missing_sessions] allows to continue when not all sessions are present for all subjects. (#1000 by @drammock)
 
 [//]: # (### :warning: Behavior changes)
 
--- a/docs/source/v1.10.md.inc
+++ b/docs/source/v1.10.md.inc
@@ -3,6 +3,7 @@
 ### :new: New features & enhancements
 
 - It is now possible to use separate MRIs for each session within a subject, as in longitudinal studies. This is achieved by creating separate "subject" folders for each subject-session combination, with the naming convention `sub-XXX_ses-YYY`, in the freesurfer `SUBJECTS_DIR`. (#987 by @drammock)
+- New boolean config option `allow_missing_sessions` allows the pipeline to continue when not all sessions are present for all subjects. (#1000 by @drammock)
 
 [//]: # (### :warning: Behavior changes)
 
--- a/mne_bids_pipeline/_config.py
+++ b/mne_bids_pipeline/_config.py
@@ -79,6 +79,12 @@ The sessions to process. If `'all'`, will process all sessions found in the
 BIDS dataset.
 """
 
+allow_missing_sessions: bool = False
+"""
+Whether to continue processing the dataset if some combinations of `subjects` and
+`sessions` are missing.
+"""
+
 task: str = ""
 """
 The task to process.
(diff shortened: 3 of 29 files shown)


larsoner (approved):
Nice, I was hoping it would look about this simple! Thanks @drammock

larsoner:
... just noticed you had WIP in the title. Let me know when I should look. In the meantime the CIs are giving it a shot but I wouldn't be surprised if they failed because of https://github.com/mne-tools/mne-bids/issues/1311

drammock:
OK, I need help. I don't really understand what is going on in `_docs.py` (more/better comments would help!) so I can't figure out what to change to get the `test_config_options_used` to pass. Tried:

- adding `allow_missing_sessions` to `_FORCE_EMPTY`
- adding `_config_utils.get_subjects_sessions` to the helper functions early in the `__init__` of `_ParseConfigSteps`
- both at once

larsoner:
Naively I would expect you to be able just to add it to the one place in `_docs.py` where you find `get_subjects` (and `get_sessions`) for example but it's probably more than that. I'd have to look. BTW you can manually run different doc-building `.py` files to replicate the error and very quickly iterate rather than doing a quick doc build.

Do you want me to take a look?

drammock:
> Naively I would expect you to be able just to add it to the one place in `_docs.py` where you find `get_subjects` (and `get_sessions`) for example but it's probably more than that.

nope. `get_subjects_sessions` only ever appears in `main()` functions, and `get_subjects` / `get_sessions` only appear in `_docs` later on when you're handling `get_config` funcs.

> BTW you can manually run different doc-building `.py` files to replicate the error and very quickly iterate rather than doing a quick doc build.

Not following what you mean here --- the failure in question is a pytest test, not doc build?

> Do you want me to take a look?

I've pushed 2 commits, one with stuff that I thought would work (or at least be needed) but wasn't after all (once I figured out the hack in the second commit). Please view the diff of those two commits separately, then comment on how I can make it less hacky. The 3rd commit (about `[None]`) is unrelated, and should fix other test failures.

larsoner:
The AST parsing I wrote wasn't good/complete enough. Needed to iterate over `call.args` in addition to `call.kwargs`. Pushed a commit to fix it.

drammock:
> The AST parsing I wrote wasn't good/complete enough. Needed to iterate over `call.args` in addition to `call.kwargs`. Pushed a commit to fix it.

OK thanks! WDYT about testing? It looks like `ds001810` has one subj w/ multiple sessions so it should work as a test case. Thinking maybe use `tmpdir` as BIDSRoot, and in it create links to just one session from that subj, but have config list 2 sessions. If you think there's a better/easier way LMK.

After that, I'll add a changelog then, and I think we'll be good!

larsoner:
> Thinking maybe use tmpdir as BIDSRoot, and in it create links to just one session from that subj, but have config list 2 sessions. If you think there's a better/easier way LMK.

Yes that could work! You don't even need to run any steps from the pipeline beyond the `init_derivatives` looks like so it could be very quick. You could run it first without the setting set, ensure it raises an error about a missing session, then modify the temp config file to be permissive, and ensure the init step runs.

larsoner:
... FWIW so far I don't think we have many (any?) tests like this so you might be starting from scratch a bit. But it will be nice to have some code for more targeted testing. So far if we add a new feature or behavior we have tried to make sure we could modify some dataset to make use of it, which isn't really sustainable as the number of options and exceptions grows.

drammock:
OK @larsoner assuming all CIs are still green, this one should be ready!

larsoner:
Ahh that CI does not have `ds001810` downloaded. Might be better to mock a little dataset rather than assume it's available. More work but will give us more flexibility down the line. I can give it a shot if you want

drammock:
> Ahh that CI does not have `ds001810` downloaded. Might be better to mock a little dataset rather than assume it's available. More work but will give us more flexibility down the line. I can give it a shot if you want

wait, which one?  The only CI failure I see now (check docs consistency) is that `Path.walk()` only exists since Py 3.12.  Can we bump that CI to 3.12 or should I swap in `os.walk` etc?

larsoner:
Oh if it's a 3.12 thing then yes feel free just to bump the version there

.github/workflows/run-tests.yml
```diff
@@ -49,7 +49,7 @@ jobs:
           pyvista: false
       - uses: actions/setup-python@v5
         with:
-          python-version: "3.11"  # no "multidict" wheels on 3.12 yet
+          python-version: "3.12"
```
drammock, line 52: should be OK since v6.0.5 (2024 Feb 1) https://github.com/aio-libs/multidict/releases/tag/v6.0.5

drammock:
> > Ahh that CI does not have `ds001810` downloaded. Might be better to mock a little dataset rather than assume it's available. More work but will give us more flexibility down the line. I can give it a shot if you want
> 
> wait, which one? The only CI failure I see now (check docs consistency) is that `Path.walk()` only exists since Py 3.12. Can we bump that CI to 3.12 or should I swap in `os.walk` etc?

ah OK, now it's showing up again :(

```
 E   FileNotFoundError: The following requested subjects were not found in the dataset: 01
```

I'll mock a dataset

drammock:
all green!

docs/source/v1.10.md.inc
```diff
@@ -3,6 +3,7 @@
 ### :new: New features & enhancements
 
 - It is now possible to use separate MRIs for each session within a subject, as in longitudinal studies. This is achieved by creating separate "subject" folders for each subject-session combination, with the naming convention `sub-XXX_ses-YYY`, in the freesurfer `SUBJECTS_DIR`. (#987 by @drammock)
+- New boolean config option `allow_missing_sessions` allows the pipeline to continue when not all sessions are present for all subjects. (#1000 by @drammock)
```
hoechenberger, line 6:
```suggestion
- New config option [`allow_missing_sessions`][mne_bids_pipeline._config. allow_missing_sessions]  allows to continue when not all sessions are present for all subjects. (#1000 by @drammock)
```

hoechenberger (approved):
otherwise, seems good to me! :) 

@larsoner I'll leave the final review / decision to you

larsoner (approved):

larsoner:
Thanks @drammock !

PR #1035, opened 2025-01-07
fix ignoring sessions that have no data

drammock:
The addition of `allow_missing_sessions` in #1000 had some problems:

1. it was storing `[None]` in the `subjs_sessions` dict in cases where a subject had no valid sessions; this was not behaving as expected
2. the use of `mne_bids.get_entity_vals` was not working as expected, because even if a subject had no *data* for a session for a given task, the presence of `coordsystem.json` or `scans.tsv` sidecars was causing it to report that the subject had *some files* for that session.

Item (2) was fixed in https://github.com/mne-tools/mne-bids/pull/1362; item (1) is fixed here. Works on my data, but will self-review to add some questions about implementation/testing.

### Before merging …

### DIFF
--- a/mne_bids_pipeline/_config_utils.py
+++ b/mne_bids_pipeline/_config_utils.py
@@ -142,9 +142,11 @@ def _get_sessions(config: SimpleNamespace) -> tuple[str, ...]:
     return tuple(str(x) for x in sessions)
 
 
-def get_subjects_sessions(config: SimpleNamespace) -> dict[str, list[str]]:
-    subj_sessions: dict[str, list[str]] = dict()
-    cfg_sessions = _get_sessions(config)
+def get_subjects_sessions(
+    config: SimpleNamespace,
+) -> dict[str, tuple[None] | tuple[str, ...]]:
+    subj_sessions: dict[str, tuple[None] | tuple[str, ...]] = dict()
+    cfg_sessions = get_sessions(config)
     # find which tasks to ignore when deciding if a subj has data for a session
     if config.task == "":
         ignore_tasks = None
@@ -165,14 +167,20 @@ def get_subjects_sessions(config: SimpleNamespace) -> dict[str, list[str]]:
             ignore_suffixes=("scans", "coordsystem"),
             ignore_datatypes=_get_ignore_datatypes(config),
         )
-        missing_sessions = set(cfg_sessions) - set(valid_sessions_subj)
-        if missing_sessions and not config.allow_missing_sessions:
-            raise RuntimeError(
-                f"Subject {subject} is missing session{_pl(missing_sessions)} "
-                f"{tuple(sorted(missing_sessions))}, and "
-                "`config.allow_missing_sessions` is False"
+        # handle datasets that don't have (named) sessions
+        if cfg_sessions == (None,):
+            assert valid_sessions_subj == (), valid_sessions_subj
+            keep_sessions = cfg_sessions
+        else:
+            missing_sessions = tuple(
+                sorted(set(cfg_sessions) - set(valid_sessions_subj))  # type: ignore
             )
-        keep_sessions = sorted(set(cfg_sessions) & set(valid_sessions_subj))
+            if missing_sessions and not config.allow_missing_sessions:
+                raise RuntimeError(
+                    f"Subject {subject} is missing session{_pl(missing_sessions)} "
+                    f"{missing_sessions}, and `config.allow_missing_sessions` is False"
+                )
+            keep_sessions = tuple(sorted(set(cfg_sessions) & set(valid_sessions_subj)))  # type: ignore
         if len(keep_sessions):
             subj_sessions[subject] = keep_sessions
     return subj_sessions
--- a/mne_bids_pipeline/_config_utils.py
+++ b/mne_bids_pipeline/_config_utils.py
@@ -146,7 +146,7 @@ def get_subjects_sessions(config: SimpleNamespace) -> dict[str, list[str]]:
     subj_sessions: dict[str, list[str]] = dict()
     cfg_sessions = _get_sessions(config)
     # find which tasks to ignore when deciding if a subj has data for a session
-    if config.task is None:
+    if config.task == "":
         ignore_tasks = None
     else:
         all_tasks = _get_entity_vals_cached(
--- a/mne_bids_pipeline/_config_utils.py
+++ b/mne_bids_pipeline/_config_utils.py
@@ -145,42 +145,42 @@ def _get_sessions(config: SimpleNamespace) -> tuple[str, ...]:
 def get_subjects_sessions(
     config: SimpleNamespace,
 ) -> dict[str, tuple[None] | tuple[str, ...]]:
-    subj_sessions: dict[str, tuple[None] | tuple[str, ...]] = dict()
-    cfg_sessions = get_sessions(config)
+    subjects = get_subjects(config)
+    cfg_sessions = _get_sessions(config)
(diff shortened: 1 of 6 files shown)


mne_bids_pipeline/_config_utils.py
```diff
@@ -159,9 +172,9 @@ def get_subjects_sessions(config: SimpleNamespace) -> dict[str, list[str] | list
                 f"{tuple(sorted(missing_sessions))}, and "
                 "`config.allow_missing_sessions` is False"
             )
-        subj_sessions[subject] = sorted(set(cfg_sessions) & set(valid_sessions_subj))
-        if subj_sessions[subject] == []:
-            subj_sessions[subject] = [None]
```
drammock, line 164: this wasn't doing what was intended: the `None` was getting interpreted as "all sessions" 🤦🏻 Now, if a subject doesn't have data for any sessions, we just omit their entry from this `subj_sessions` dict.

mne_bids_pipeline/_config_utils.py
```diff
@@ -142,14 +142,27 @@ def _get_sessions(config: SimpleNamespace) -> tuple[str, ...]:
...
+        ignore_tasks = tuple(set(all_tasks) - set([config.task]))
     for subject in get_subjects(config):
-        # Only traverse through the current subject's directory
+        # Check current subject's directory for available sessions
         valid_sessions_subj = _get_entity_vals_cached(
             config.bids_root / f"sub-{subject}",
             entity_key="session",
+            ignore_tasks=ignore_tasks,
+            ignore_acquisitions=("calibration", "crosstalk"),
+            ignore_suffixes=("scans", "coordsystem"),
```
drammock, line 165: These two lines are the important bit, for getting MNE-BIDS to only tell us if *data* files are present.
This `ignore_suffixes` line requires upstream https://github.com/mne-tools/mne-bids/commit/41ce36a4dc938e85a2f6a32ed3eeb88c732bfbae to work, so we'll need to pin MNE-BIDS to that commit, or wait for the next MNE-BIDS release before merging here. 

I'm also not really sure if `coordsystems.json` and `scans.tsv` are the only possible sidecars that could have the undesirous effect (of having `get_entity_vals` return a false positive when no *data files* are actually present for the requested session --- cf. discussion at https://github.com/mne-tools/mne-bids/issues/1351). If you can think of other sidecar suffixes to exclude, speak now.

drammock (commented):
regarding testing, @larsoner and I came up with the plan to use a small dataset we have from UW (after BIDSifying it and uploading to OSF). It could make sense to add that in this PR, or a separate one

mne_bids_pipeline/_download.py
```diff
@@ -43,16 +43,18 @@ def _download_from_web(*, ds_name: str, ds_path: Path) -> None:
 
     ds_path.mkdir(parents=True, exist_ok=True)
     path = ds_path.parent.resolve(strict=True)
-    fname = f"{ds_name}.zip"
+    ext = "tar.gz" if options.get("processor") == "untar" else "zip"
+    processor = pooch.Untar if options.get("processor") == "untar" else pooch.Unzip
```
drammock, line 47: this sets up a problem that is fixed by https://github.com/fatiando/pooch/pull/458, but I wonder if it's better to set up a warning filter here rather than pin to a pooch fork / wait for pooch PR merge or even pooch release... WDYT?

mne_bids_pipeline/tests/configs/config_funloc.py
```diff
@@ -0,0 +1,24 @@
...
+l_freq = None
+h_freq = 40.0
+use_maxwell_filter: bool = True
+
+
+# Epochs
+epochs_tmin = -0.08
+epochs_tmax = 0.18
+epochs_decim = 10  # 2000->200 Hz
+baseline = (None, 0)
```
drammock, line 18: this is all copied from a different dataset's config, may not be most suitable, TBD.  LMK if you have suggestions here.

mne_bids_pipeline/tests/test_run.py
```diff
@@ -274,3 +275,38 @@ def test_missing_sessions(
...
+def test_session_specific_mri(
+    dataset: str,
+    monkeypatch: pytest.MonkeyPatch,
+    dataset_test: Any,
+    tmp_path: Path,
+    capsys: pytest.CaptureFixture[str],
+) -> None:
+    """Test running a dataset."""
+    # TODO edit this test to monkeypatch the dataset, making it seem like
+    # there's only one subject who has different MRIs for different sessions
```
drammock, line 291: just calling out this TODO; you can mostly ignore this test function for now.

mne_bids_pipeline/_download.py
hoechenberger: In any case we should add a comment to document why we need this differentiation :)

drammock: Sure I could add a comment. I thought it would be evident from the names "Untar" and "Unzip" that the purpose was supporting different compression formats

drammock:
@larsoner @hoechenberger this one is all green and ready for review

mne_bids_pipeline/_config_utils.py
larsoner: > so we'll need to pin MNE-BIDS to that commit, or wait for the next MNE-BIDS release before merging here.

This is a tough req since MNE-BIDS does not release super frequently. And in the meantime anyone doing `pip install -ve .` will have a valid usable (sufficiently up-to-date) editable install overwritten by the pinned install etc.

Since this is a bit of a niche bug (right?), could we instead use `inspect` to see if the right `kwarg` is present and pass it if so? Then once MNE-BIDS releases we bump our min req to the one that actually has it and remove the `kwarg` thing here. If we wait to release MNE-BIDS-Pipeline until MNE-BIDS releases then users are in good shape, and in the meantime devs are also in (sufficiently) good shape. We'd "just" need to make CIs install the dev version of MNE-BIDS, which they might already be doing.

larsoner: ... i.e., we don't have to wait to merge this PR and we also don't have to explicitly pin below in `pyproject.toml`

drammock: I can do that, but if someone sets `allow_missing_sessions` (and actually *is* missing some sessions) and they don't have a new enough mne_bids, then mne_bids will report that they have data for a given subj+session that they don't actually have data for, and then the pipeline will bomb later.

Maybe I should use `inspect` here, but also check it during config import, to make sure that if they want to `allow_missing_sessions`, that they also have a sufficiently capable version of mne_bids?

larsoner: Yeah that would work, if they have it set and the `kwarg` doesn't exist then raise an error, or at least emit a warning.

Should be a transient problem anyway since hopefully we'll release *after* MNE-BIDS, and then pin in `pyproject.toml` to `>=` a released version

drammock:
well, we know the check works at least:
```
FAILED mne_bids_pipeline/tests/test_run.py::test_missing_sessions[True] - mne_bids_pipeline._config_import.ConfigError: You've requested to `allow_missing_sessions`, but this functionality requires a newer version of `mne_bids` than you have available. Please update MNE-BIDS (or if on the latest version, install the dev version).
```

larsoner:
Accidental TDD is the best kind of TDD

drammock:
> Accidental TDD is the best kind of TDD

not when you pay for your CIs 😅

drammock:
ok all green again 🙂

larsoner (approved):
Merging, thanks @drammock !

PR #1038, opened 2025-01-16
add config for passing extra kwargs to maxwell_filter

drammock:
### Before merging …

### DIFF
--- a/mne_bids_pipeline/steps/preprocessing/_03_maxfilter.py
+++ b/mne_bids_pipeline/steps/preprocessing/_03_maxfilter.py
@@ -378,11 +378,6 @@ def run_maxwell_filter(
         head_pos=head_pos,
         extended_proj=extended_proj,
     )
-    if duplicates := (set(cfg.mf_extra_kws) & set(mf_kws)):
-        raise RuntimeError(
-            f"`mf_extra_kws` contains keys {', '.join(sorted(duplicates))} that are "
-            "handled by dedicated config keys. Please remove them from `mf_extra_kws`."
-        )
     mf_kws |= cfg.mf_extra_kws
 
     logger.info(**gen_log_kwargs(message=f"{apply_msg} {recording_type} data"))
--- a/mne_bids_pipeline/steps/preprocessing/_03_maxfilter.py
+++ b/mne_bids_pipeline/steps/preprocessing/_03_maxfilter.py
@@ -378,6 +378,8 @@ def run_maxwell_filter(
         head_pos=head_pos,
         extended_proj=extended_proj,
     )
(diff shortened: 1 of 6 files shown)


mne_bids_pipeline/steps/preprocessing/_03_maxfilter.py
```diff
@@ -378,6 +378,12 @@ def run_maxwell_filter(
         head_pos=head_pos,
         extended_proj=extended_proj,
     )
+    if duplicates := (set(cfg.mf_extra_kws) & set(mf_kws)):
+        raise RuntimeError(
```
drammock, line 382: is there a better error class than `RuntimeError` that I should be using instead?  ConfigError or similar?

hoechenberger: ValueError perhaps?

hoechenberger: ValueError perhaps?

larsoner: Agreed. Would it be worth adding a `mne_bids_pipeline.ConfigError` subclass?
```
class ConfigError(ValueError):
    pass
```

Also this validation would ideally happen in with the rest of the config validation stuff in `_config_import.py` but I understand if it's a pain because we'd have to list the set of (dis)allowed `kwargs` over there probably, which isn't elegant

drammock: it wasn't hard. see [647b830](https://github.com/mne-tools/mne-bids-pipeline/pull/1038/commits/647b830c453a8c2d6d7ec1d6c11eadabc4834155)

mne_bids_pipeline/steps/preprocessing/_03_maxfilter.py
```diff
@@ -378,6 +378,7 @@ def run_maxwell_filter(
         head_pos=head_pos,
         extended_proj=extended_proj,
     )
+    mf_kws |= cfg.mf_extra_kws
```
larsoner, line 381: or similar
```suggestion
    # If the mf_kws keys above change, we need to modify our list
    # of illegal keys in _config_import.py
    mf_kws |= cfg.mf_extra_kws
```

larsoner (commented):
Other than one tiny suggestion, LGTM!

drammock:
ok, incorporated Eric's suggestion and updated the validation test to check that it works. Should be ready for merge-when-green?

larsoner:
Yep, done!

PR #1039, opened 2025-01-17
Improve (and test) support for session-specific MRIs

drammock:
~~Per conversation with @larsoner I'm splitting out the work of adding the funloc dataset (and fixing the bugs it reveals) to make the diff cleaner on #1035 (which is actually dealing with a different problem than what funloc addresses --- #1035 is about skipping missing sessions, not handling session-specific anat).~~

This PR has been rebased after #1045 was merged, so now all it does is monkeypatch the funloc dataset to test support for session-specific MRIs.

### Before merging …

### DIFF
--- a/docs/mkdocs.yml
+++ b/docs/mkdocs.yml
@@ -136,8 +136,6 @@ plugins:
     - search
     - macros
     - privacy # https://squidfunk.github.io/mkdocs-material/plugins/privacy/
-    - tags:
-        tags_file: tags.md
     - include-markdown
     - exclude:
         glob:
--- a/mne_bids_pipeline/_config_utils.py
+++ b/mne_bids_pipeline/_config_utils.py
@@ -55,6 +55,12 @@ def get_fs_subject(
         return f"sub-{subject}"
 
 
+def _has_session_specific_anat(
+    subject: str, session: str | None, subjects_dir: pathlib.Path
+) -> bool:
+    return (subjects_dir / f"sub-{subject}_ses-{session}").exists()
+
+
 @functools.cache
 def _get_entity_vals_cached(
     *args: list[Any],
--- a/mne_bids_pipeline/tests/test_run.py
+++ b/mne_bids_pipeline/tests/test_run.py
@@ -363,7 +363,6 @@ def test_session_specific_mri(
                     dst_file = dst_file.replace(subj, freesurfer_subject_mapping[subj])
                     break
             shutil.copyfile(src=walk_root / _file, dst=dst_dir / offset / dst_file)
-    # print_dir_tree(new_bids_path.root)  # for debugging
     # update config so that `subjects_dir` and `deriv_root` also point to the tempdir
     extra_config = f"""
 from pathlib import Path
@@ -408,6 +407,6 @@ deriv_root = Path("{new_bids_path.root}") / "derivatives" / "mne-bids-pipeline" 
         )
         result = pattern.search(coregs["html"])
         assert result is not None
-        assert float(result.group("dist")) < 3
+        assert float(result.group("dist")) < 3  # fit between pts and outer_skin < 3 mm
         results.append(result.groups())
     assert results[0] != results[1]  # different npts and/or different mean distance
--- a/mne_bids_pipeline/tests/test_run.py
+++ b/mne_bids_pipeline/tests/test_run.py
@@ -358,8 +358,9 @@ def test_session_specific_mri(
     # print_dir_tree(new_bids_path.root)  # for debugging
     # update config so that `subjects_dir` and `deriv_root` also point to the tempdir
     extra_config = f"""
+from pathlib import Path
 subjects_dir = "{new_bids_path.root / "derivatives" / "freesurfer" / "subjects"}"
-deriv_root = "{new_bids_path.root / "derivatives" / "mne-bids-pipeline" / "MNE-funloc-data"}"
+deriv_root = Path("{new_bids_path.root}") / "derivatives" / "mne-bids-pipeline" / "MNE-funloc-data"
 """  # noqa E501
     extra_path = tmp_path / "extra_config.py"
     extra_path.write_text(extra_config)
--- a/mne_bids_pipeline/tests/test_run.py
+++ b/mne_bids_pipeline/tests/test_run.py
@@ -9,7 +9,9 @@ from pathlib import Path
 from typing import Any, TypedDict
 
 import pytest
+from mne_bids import BIDSPath, get_bids_path_from_fname
 
+from mne_bids_pipeline._config_import import _import_config
 from mne_bids_pipeline._download import main as download_main
 from mne_bids_pipeline._main import main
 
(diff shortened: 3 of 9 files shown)


mne_bids_pipeline/tests/configs/config_funloc.py
```diff
@@ -0,0 +1,36 @@
...
+use_maxwell_filter: bool = True
+mf_st_duration = 60.0
+# SSP
+n_proj_eog = dict(n_mag=1, n_grad=1, n_eeg=2)
+n_proj_ecg = dict(n_mag=1, n_grad=1, n_eeg=0)
+
+# Epochs
+epochs_tmin = -0.2
+epochs_tmax = 0.5
+epochs_t_adjust = -4.0e-3  # TODO is there a way to do this in MBP?
```
drammock, line 26: @larsoner does MBP have a way of doing what `t_adjust` does in mnefun?

larsoner: No, this should be fixed/adjusted when bidsifying the dataset I think (the timings, event types, etc. in there should be as accurate as possible and not require fixing)

mne_bids_pipeline/tests/test_run.py
```diff
@@ -274,3 +284,101 @@ def test_missing_sessions(
         print()
         with context:
             main()
+
+
+@pytest.mark.dataset_test
+@pytest.mark.parametrize("dataset", ["MNE-funloc-data-with-session-specific-MRI"])
+def test_session_specific_mri(
```
drammock, line 291: @larsoner I need a bit of help with the CircleCI test config.

What I'd like to do is *not* run the funloc dataset as a "dataset test" and only use it for this `session_specific_mri` test. But (I think?) I had to set it up as a "dataset test" in order to get the auto-dataset-downloading to work.  Also I've hit issues a couple times with the `test_documented` test, complaining that the number of dataset tests (or caches) doesn't match the number of entries in `tests/test_run.py:TEST_SUITE`.

In a recent commit, I changed the name of the dataset here to be `MNE-funloc-data-with-session-specific-MRI` so that it would get its own derivatives folder / not overwrite or get overwritten by the "real" dataset test for funloc (which again, I'm not sure we actually need?).

What is your advice on how to handle this setup?

larsoner: > What I'd like to do is not run the funloc dataset as a "dataset test"

FWIW at some point it's probably worth adding it a proper dataset and example. It's an easy to understand dataset, has simultaneous M/EEG, source imaging, *and* multiple subjects, which we don't have so far IIRC.

> I had to set it up as a "dataset test" in order to get the auto-dataset-downloading to work. Also I've hit issues a couple times with the test_documented test, complaining that the number of dataset tests (or caches) doesn't match the number of entries in tests/test_run.py:TEST_SUITE.

So far the categories "dataset that you want to `pytest`", "dataset that is rendered as an example on the website", "dataset that is cached by CircleCI", and "dataset you want to download easily with `pytest --download`" have all been coextensive. There was a point a year or two ago maybe when we *wanted* things that way but they weren't, so I added a bunch of stuff to `test_documented` to make sure all datasets we had at the time -- or would add in the future -- would be consistently used this way. But to whatever extent we don't want those categories to be the same going forward, we can/should relax the `test_documented` test as needed.

For this case, though, it sounds like the thing to do might be to first add `funloc` as a normal, tested dataset in `main` (i.e., split off some stuff from this PR). Then you shouldn't have to mess with any of the `test_documented` etc., just make sure that it runs. I think we discussed this before and said it maybe wasn't necessary... but now it seems useful to help you (and me in review) make sure everything works with the normal dataset. Then this PR gets much simpler, it's any changes needed to the tests (including `test_documented`?) to accommodate the new session-specific anat.

But if you already created the dataset with session-specific anat and it *needs* this PR to work, then might as well do it all here!

drammock: I can split off `funloc` as a normal dataset test. That part I had working a few commits ago, so hopefully that PR will "just work".

I'm still not 100% clear on what things should look like here *after* that bit is split off, in order to (1) get `session-specific-MRI` derivatives written to their own folder, and (2) get `--download` to work with that test, and (3) how/whether to try to cache the monkeypatched dataset instead of recreating it every time.  But we can address those questions once the diff is simplified.

larsoner: For (2) --download we could add a --download-only if needed. But maybe easier to just require that people run the regular test with --download or somehow otherwise obtain the dataset before running the new one that monkey patches. 

For (3) caching hopefully we don't need it and can copy the (few?) needed files/dirs to a tmp_path and modify as needed. This modification could include modifying the config file itself to use a subdir of tmp_path for the derivatives, would would take care of (1)

larsoner: > extra_config to set the subjects_dir. The downside to writing pipeline derivs into tmp_dir is that (I think?) you won't be able to easily view the report files on CI (viewing the reports is how I confirmed locally that different session-specific MRIs were actually getting used)

IIRC on failure I think the directory actually exists / doesn't get wiped so I think you still could. Might require `--pdb` but I don't think it does?

mne_bids_pipeline/_config_utils.py
```diff
@@ -55,6 +55,12 @@ def get_fs_subject(
         return f"sub-{subject}"
 
 
+def _has_session_specific_anat(
+    subject: str, session: str | None, subjects_dir: pathlib.Path
+) -> bool:
+    return (subjects_dir / f"sub-{subject}_ses-{session}").exists()
```
drammock, line 61: This is merely moved from `_01_recon_all.py`, as it's now needed in more than one place.

mne_bids_pipeline/tests/test_run.py
```diff
@@ -278,3 +282,131 @@ def test_missing_sessions(
...
+        # for filenames that contain the subject identifier (BEM files, morph maps),
+        # we need to change the filename too, not just parent folder name
+        for _file in files:
+            dst_file = _file
+            for subj in freesurfer_subject_mapping:
+                if subj in dst_file:
+                    dst_file = dst_file.replace(subj, freesurfer_subject_mapping[subj])
+                    break
+            shutil.copyfile(src=walk_root / _file, dst=dst_dir / offset / dst_file)
+    # print_dir_tree(new_bids_path.root)  # for debugging
```
drammock, line 365:
```suggestion
```
cruft

mne_bids_pipeline/tests/test_run.py
```diff
@@ -278,3 +282,131 @@ def test_missing_sessions(
...
+    # update config so that `subjects_dir` and `deriv_root` also point to the tempdir
+    extra_config = f"""
+from pathlib import Path
+subjects_dir = "{new_bids_path.root / "derivatives" / "freesurfer" / "subjects"}"
+deriv_root = Path("{new_bids_path.root}") / "derivatives" / "mne-bids-pipeline" / "MNE-funloc-data"
+"""  # noqa E501
```
drammock, line 371: It was not working to specify `deriv_root` as a `str`, hence the (IMO awkward) need to cast it to `Path` in this stringified `extra_config` (and also hence the inclusion of `Path` in `valid_extra_names` in `_config_import.py`).  If you're bothered by this, I can pass in `deriv_root` as a `str` and add in an explicit `Path()` cast wherever it's needed (I forget now exactly where the error occurred, but not hard to find out again)

mne_bids_pipeline/tests/test_run.py
```diff
@@ -278,3 +282,131 @@ def test_missing_sessions(
...
+        coregs = next(
+            filter(lambda x: x["dom_id"] == "Sensor_alignment", report["_content"])
+        )
+        pattern = re.compile(
+            r"Average distance from (?P<npts>\d+) digitized points to head: "
+            r"(?P<dist>\d+(?:\.\d+)?) mm"
+        )
+        result = pattern.search(coregs["html"])
+        assert result is not None
+        assert float(result.group("dist")) < 3
```
drammock, line 410:
```suggestion
        assert float(result.group("dist")) < 3  # fit between pts and outer_skin < 3 mm
```
add an explainer comment

mne_bids_pipeline/tests/test_run.py
```diff
@@ -278,3 +282,131 @@ def test_missing_sessions(
...
+    # hack in the new bids_root
+    command.append(f"--root-dir={new_bids_path.root}")
+    if "--pdb" in sys.argv:
+        command.append("--n_jobs=1")
+    monkeypatch.setenv("_MNE_BIDS_STUDY_TESTING", "true")
+    monkeypatch.setattr(sys, "argv", command)
+    with capsys.disabled():
+        print()
+        main()
+    # check some things that are indicative of different MRIs being used in each session
```
drammock, line 387: this is just the first thing that occurred to me to check, but there's probably a better / simpler / more definitive thing to look at, LMK if you think of something.

drammock (commented):
OK this one is just about ready. Test passes locally. Not done is figuring out how to run `test_session_specific_mri` in an environment that actually has the Funloc dataset available (it's not there for the "doc consistency and codespell" workflow, which is what runs `pytest -m "not dataset_test"`).  Is the best approach to create a new CircleCI job to run that test, and make it depend on the funloc cache job? (and also tag it as a dataset test? Or give it a different fixture tag and use that tag to exclude it from "doc consistency and codespell"?)

larsoner:
One test fails, I think you might need to mark it a `dataset_test`?

drammock:
> One test fails, I think you might need to mark it a `dataset_test`?

did you see my message from before? 👇🏻 

>  Not done is figuring out how to run test_session_specific_mri in an environment that actually has the Funloc dataset available (it's not there for the "doc consistency and codespell" workflow, which is what runs pytest -m "not dataset_test"). Is the best approach to create a new CircleCI job to run that test, and make it depend on the funloc cache job? (and also tag it as a dataset test? Or give it a different fixture tag and use that tag to exclude it from "doc consistency and codespell"?)

mne_bids_pipeline/tests/test_run.py
```diff
@@ -278,3 +282,131 @@ def test_missing_sessions(
         print()
         with context:
             main()
+
+
+def test_session_specific_mri(
```
larsoner, line 287: I think this should make `doc` not run the test.
```suggestion
@pytest.mark.dataset_test
def test_session_specific_mri(
```
But then you'll want to modify `.github/workflows/run_tests.yml` to have another `pytest` line that *does* run this test.

larsoner: ... I think that would help us move toward "CircleCI should run all unit tests that it needs for docs" and "GHA runs non-doc unit tests, including a couple of pipeline caching tests"

drammock: ok.  IIUC though that would involve downloading the dataset from OSF into the GHA runner. Is that better than setting it up as a CircleCI job that is dependent on `cache-MNE-funloc-data`?  I guess the GHA job would be free, so that's one advantage...

larsoner: Yeah the DL and cache on GHA should be pretty quick assuming it works and doesn't hit a size limit or anything

drammock:
@larsoner after installing `pyvistaqt` (like the error message suggests) I'm still hitting the same error:

```
FAILED mne_bids_pipeline/tests/test_run.py::test_session_specific_mri - RuntimeError: Could not load any valid 3D backend
pyvistaqt: No Qt bindings could be found
notebook: No module named 'ipyevents'

 install pyvistaqt, using pip or conda:
'pip install pyvistaqt'
'conda install -c conda-forge pyvistaqt'

 or install ipywidgets, if using a notebook backend
'pip install ipywidgets'
'conda install -c conda-forge ipywidgets'
```
any tips?

larsoner:
Needed to install a qt binding, we could probably improve that error message

docs/mkdocs.yml
```diff
@@ -136,8 +136,6 @@ plugins:
     - search
     - macros
     - privacy # https://squidfunk.github.io/mkdocs-material/plugins/privacy/
-    - tags:
-        tags_file: tags.md
```
hoechenberger, line 140: why this change?

drammock:
argh now it's OSF throttling:
```
FAILED mne_bids_pipeline/tests/test_run.py::test_session_specific_mri - requests.exceptions.HTTPError: 429 Client Error: Too Many Requests for url: https://files.osf.io/v1/resources/rxvq7/providers/osfstorage/5cbdba4ef2be3c0019091890?version=2&action=download&direct
```

larsoner:
Looks good, thanks @drammock !

PR #1043, opened 2025-01-23
add framework for TWA head pos as MF destination

drammock:
Opening this for early feedback. So far it looks like it won't require a very big diff, but perhaps that's the naïvety of not having tested it yet 🙈. Could probably be tested with `funloc` once that's finished / merged.

### Before merging …

### DIFF
--- a/mne_bids_pipeline/_config.py
+++ b/mne_bids_pipeline/_config.py
@@ -644,7 +644,7 @@ options or specifying the origin manually.
     ```
 """
 
-mf_destination: Literal["reference_run"] | FloatArrayLike = "reference_run"
+mf_destination: Literal["reference_run", "twa"] | FloatArrayLike = "reference_run"
 """
 Despite all possible care to avoid movements in the MEG, the participant
 will likely slowly drift down from the Dewar or slightly shift the head
@@ -666,6 +666,9 @@ account, we are realigning all data to a single position. For this, you can:
    from mne.transforms import translation
    mf_destination = translation(z=0.04)
    ```
+3. Compute the time-weighted average head position for each run, and use that as the
+   destination coordinates for each run. This will result in a device-to-head
+   transformation that differs between runs within each subject.
 """
 
 mf_int_order: int = 8
--- a/mne_bids_pipeline/_config.py
+++ b/mne_bids_pipeline/_config.py
@@ -666,7 +666,7 @@ account, we are realigning all data to a single position. For this, you can:
    from mne.transforms import translation
    mf_destination = translation(z=0.04)
    ```
-3. Compute the time-weighted average head position for each run, and use that as the
+3. Compute the time-weighted average head position across all runs, and use that as the
    destination coordinates for each run. This will result in a device-to-head
    transformation that differs between runs within each subject.
 """
--- a/mne_bids_pipeline/_config.py
+++ b/mne_bids_pipeline/_config.py
@@ -666,9 +666,9 @@ account, we are realigning all data to a single position. For this, you can:
    from mne.transforms import translation
    mf_destination = translation(z=0.04)
    ```
-3. Compute the time-weighted average head position across all runs, and use that as the
-   destination coordinates for each run. This will result in a device-to-head
-   transformation that differs between runs within each subject.
+3. Compute the time-weighted average head position across all runs in a session,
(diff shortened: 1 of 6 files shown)


mne_bids_pipeline/_config.py
```diff
@@ -666,6 +666,9 @@
    from mne.transforms import translation
    mf_destination = translation(z=0.04)
    ```
+3. Compute the time-weighted average head position for each run, and use that as the
```
larsoner, line 669: Just to make it clearer
```suggestion
3. Compute the time-weighted average head position across all runs, and use that as the
```

mne_bids_pipeline/steps/preprocessing/_03_maxfilter.py
```diff
@@ -340,9 +340,21 @@ def run_maxwell_filter(
...
     if isinstance(destination, str):
-        assert destination == "reference_run"
-        destination = raw.info["dev_head_t"]
+        if destination == "reference_run":
+            destination = raw.info["dev_head_t"]
+        elif destination == "twa_hp":
+            assert head_pos is not None, (
+                "cannot compute time-weighted average head position without movement "
+                "compensation. Please set `mf_mc=True` in your config file."
+            )
```
larsoner, line 356: I think ideally this would happen in the config validation code I think rather than here, that way a user gets immediate feedback when starting their pipeline run rather than (perhaps much) later after some steps are potentially performed.

drammock (commented):
OK this is now ready for another set of eyes again. Still TODO is setting up a proper test for it; I haven't had time yet to see if there's a suitable dataset (funloc won't work without monkeypatching, as it doesn't have multiple runs).  That said, it appears to be working locally.

It feels like TWA-HP should actually be a separate "step" between `_02_headpos` and `_03_maxfilter`. As implemented now, if run 1 and run 2 for a given subj/sess are happening in parallel, they may both try to compute TWA-HP, where really it should only ever need to happen once.

Right now there is no logging.

I suppose it might be possible to tack TWA-HP computation onto the *end* of the headpos step instead of the beginning of the maxfilter step... then in maxfilter step we just check the config and load the destination `Transform`... that might make both logging and parallelization issues easier... WDYT?

larsoner:
> I suppose it might be possible to tack TWA-HP computation onto the end of the headpos step instead of the beginning of the maxfilter step... then in maxfilter step we just check the config and load the destination Transform... that might make both logging and parallelization issues easier... WDYT?

Yeah if you're computing time-varying head positions, I think it makes sense to always compute the time-weighted average and save it. It's very quick to compute and the resulting file is tiny. Then whether or not you use it in the maxwell_filter step depends on whether or not you set the destination to `'twa'` or whatever

drammock:
@larsoner this is ready for review. Main outstanding question is how to test it. AFAICT `ds000117` is the only dataset with multiple runs?  Locally I changed its config to include 

```py
mf_mc = True
mf_destination = "twa"
```

and it ran fine, but IIUC it's the only one that actually contains multiple runs, so making that config change would cease testing the case where `mf_reference_run` is not the first run.

If changing the `ds000117` config is not an option, I could try hacking a dataset to mock having multiple runs. That might be slighly preferable (as it would allow adding assertions about what the resulting `mf_destination` should look like) but it's obviously a lot more work than a 2-line config change to an existing dataset test.  WDYT?

larsoner:
Why do there have to be multiple runs? The time-weighted average should be computed *over all time in all runs* I think. So it shouldn't really matter if there is one run or several, that just changes the length of the `list`s passed to `mne.preprocessing.compute_average_dev_head_t`, no?

drammock:
> Why do there have to be multiple runs? The time-weighted average should be computed _over all time in all runs_ I think. So it shouldn't really matter if there is one run or several, that just changes the length of the `list`s passed to `mne.preprocessing.compute_average_dev_head_t`, no?

True, it can work with a single run. I just enabled it on ds004229

larsoner (approved):
I checked the CircleCI log and see:
```
│21:48:00│ ⏳️ sub-102 Destination is 27.3 mm and 13.0° from the original head position
```
and on `main` it's
```
│17:51:24│ ⏳️ sub-102 Destination is 17.1 mm and 19.2° from the original head position
```
so it did indeed change. The code also looks good so in it goes, thanks @drammock 
!

PR #1044, opened 2025-01-23
validate extra_config

### LINKED ISSUE #1041: strange error when loading config for new funloc dataset
locally when running `pytest -k session_specific_mri` (which is being added in #1039) I get the following traceback:

```
mne_bids_pipeline/tests/test_run.py:365: in test_session_specific_mri
    main()
mne_bids_pipeline/_main.py:221: in main
    config_imported = _import_config(
mne_bids_pipeline/_config_import.py:79: in _import_config
    config = SimpleNamespace(**{k: getattr(config, k) for k in keep_names})
E   AttributeError: 'types.SimpleNamespace' object has no attribute '_n_jobs'. Did you mean: 'n_jobs'?
```

it's the inner `getattr(config, k)` part that fails when `k` is `_n_jobs`. If I insert an error just before the problematic line, then try `pytest -k ds000247 --pdb` then `_n_jobs` isn't in the `config`... it's only added if `extra_config` is present:

https://github.com/mne-tools/mne-bids-pipeline/blob/8408697da094c04b0f05ff90a634545439aef6f7/mne_bids_pipeline/_config_import.py#L58-L66

Which I think means there's an implicit assumption that if `extra_config` is defined, that it *must* include a `_n_jobs` key.  Off the top of my head I don't see why that condition necessarily should hold, though prior to adding `funloc` it *did* incidentally hold: in `tests/test_run.py` `_n_jobs` shows up 3 times, (in the `extra_config` for ERP_CORE ERN, ds000248_base, and ds000248_ica).

AFAICT options are:

1. only insert that key into `keep_names` if it's actually present in the `extra_config`
2. put a comment near where I hit the error, saying "FYI you *must* have `_n_jobs` in there if you provide an `extra_config`"

WDYT?  Do you know/notice something else I haven't thought of?

Bonus context:
- If I hack in a `check=True` the config check (which happens before the line triggering the traceback) passes. There really isn't anything special about the new `tests/configs/config_funloc.py`

drammock:
closes #1041 

### Before merging …

### DIFF
--- a/docs/source/dev.md.inc
+++ b/docs/source/dev.md.inc
@@ -32,11 +32,9 @@
 - Bundle all previously-external JavaScript to better preserve users' privacy (#982 by @hoechenberger)
 - Document the need for offscreen rendering support when running on headless servers (#997 by @drammock)
 
-### :medical_symbol: Code health
+### :medical_symbol: Code health and infrastructure
 
 - Switch from using relative to using absolute imports (#969 by @hoechenberger)
 - Enable strict type checking via mypy (#995, #1013, #1016 by @larsoner)
-
-### :medical_symbol: Code health and infrastructure
-
 - Improve logging messages in maxwell filtering steps. (#893 by @drammock)
+- Validate extra config params passed in during testing. (#1044 by @drammock)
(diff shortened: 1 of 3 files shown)


larsoner:
Thanks @drammock

PR #1053, opened 2025-02-19
MAINT: Refactor and document CIs better

larsoner:
No need for changelog update. Will require updating merge reqs because it splits the caching test into two steps. More DRY and should overall complete a bit faster even though there is more `pip` overhead.

Just for fun I'm going to try adding an `arm` run and we can see what breaks. Then I'll remove :)

Ready for review/merge in the meantime

### DIFF
--- a/.circleci/run_dataset_and_copy_files.sh
+++ b/.circleci/run_dataset_and_copy_files.sh
@@ -27,7 +27,7 @@ else
 fi
 
 SECONDS=0
-EMPH=$(tput setaf 2 bold)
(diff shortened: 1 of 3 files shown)


larsoner:
@drammock now in https://app.circleci.com/pipelines/github/mne-tools/mne-bids-pipeline/4928/workflows/35e80018-6a3e-4248-ba58-539aa6138292/jobs/76455 for example you see:

image

...

image

Maybe would have helped?

PR #1061, opened 2025-02-26
extra keyword arguments to `find_bad_channels_maxwell`, `notch_filter`, and `filter_data`; bugfix compute_rank channel types

harrisonritz:
Sorry there are so many commits, I had copied this branch from the branch I did the other PR on (lessons learned).

- new config options that let you provide a dictionary of keyword argument to `find_bad_channels_maxwell`, `notch_filter`, and `filter_data`.
- now allow for notch_filter to have a frequency of `None`, since this is allow for `method="spectrum_fit"`
- bugfix for `compute_rank` that allows for `mag` sensors, rather than `meg` sensors (sorry this is tangential). This was breaking for OPM, which only has `mag`.
- new test for custom keyword arguments

I haven't added any checking for whether these keyword arguments overlap with existing arguments, because the MNE functions already will throw an error if you try to give them multiple values.

### Before merging …


mne_bids_pipeline/_config.py
```diff
@@ -884,6 +894,18 @@
 Specifies if the iterative version of the Zapline algorithm should be run.
 """
 
+notch_extra_kws: dict[str, Any] = {}
```
hoechenberger, line 897: Please make the default `None` rather than an empty dictionary, as one would to in a function signature. Feels cleaner and safer.

mne_bids_pipeline/_config.py
```diff
@@ -884,6 +894,18 @@
...
+"""
+A dictionary of extra kwargs to pass to `mne.filter.notch_filter`. If kwargs
+are passed here that have dedicated config settings already, an error will be raised.
+"""
+
+bandpass_extra_kws: dict[str, Any] = {}
```
hoechenberger, line 903: Same

mne_bids_pipeline/_config.py
```diff
@@ -884,6 +894,18 @@
...
+"""
+A dictionary of extra kwargs to pass to `mne.filter.filter_data`. If kwargs
+are passed here that have dedicated config settings already, an error will be raised.
```
hoechenberger, line 906: Can we not mention those explicitly?

hoechenberger (commented):
Great work, I left a few comments

mne_bids_pipeline/_config.py
harrisonritz: This was formatted to match the existing config defaults for `mf_extra_kws`

https://github.com/mne-tools/mne-bids-pipeline/blob/30c779864b3daf3d15261f786d6e5dc905aac9e0/mne_bids_pipeline/_config.py#L791

mne_bids_pipeline/_config.py
harrisonritz: Are you thinking like provide a link to the mne documentation to see the options for kwargs? I like that.

This was also based on the config for `mf_extra_kws`:

https://github.com/mne-tools/mne-bids-pipeline/blob/30c779864b3daf3d15261f786d6e5dc905aac9e0/mne_bids_pipeline/_config.py#L791-L795

mne_bids_pipeline/tests/configs/config_ds000248_extra_kws.py
```diff
@@ -0,0 +1,76 @@
...
+deriv_root = "~/mne_data/derivatives/mne-bids-pipeline/ds000248_extra_kws"
+subjects_dir = f"{bids_root}/derivatives/freesurfer/subjects"
+
+subjects = ["01"]
+rename_events = {"Smiley": "Emoji", "Button": "Switch"}
+conditions = ["Auditory", "Visual", "Auditory/Left", "Auditory/Right"]
+epochs_metadata_query = "index > 0"  # Just for testing!
+contrasts = [("Visual", "Auditory"), ("Auditory/Right", "Auditory/Left")]
+
+time_frequency_conditions = ["Auditory", "Visual"]
```
drammock, line 16: drive-by comment since I don't have time for a full review at the moment: since this test only runs `preprocessing` steps, there are some config values you could omit here, which would make the important parts of the config easier to spot. These are 3 of them

harrisonritz: good idea. done.

mne_bids_pipeline/_config.py
hoechenberger: alright, I stand corrected!

mne_bids_pipeline/_config.py
harrisonritz: ok, I added links to the documentation :)

hoechenberger (approved):
LGTM but I haven't actually looked at the test cases

docs/source/dev.md.inc
```diff
@@ -7,6 +7,7 @@
 - New config option [`mf_extra_kws`][mne_bids_pipeline._config.mf_extra_kws] passes additional keyword arguments to `mne.preprocessing.maxwell_filter`. (#1038 by @drammock)
 - New value `"twa"` for config option [`mf_destination`][mne_bids_pipeline._config.mf_destination], to use the time-weighted average head position across runs as the destination position. (#1043 and #1055 by @drammock)
 - New config options [`mf_cal_missing`][mne_bids_pipeline._config.mf_cal_missing] and [`mf_ctc_missing`][mne_bids_pipeline._config.mf_ctc_missing] for handling missing calibration and cross-talk files (#1057 by @harrisonritz)
+- New config options `find_bad_channels_extra_kws`, `notch_extra_kws`, and `bandpass_extra_kws` to pass additional keyword arguments to `mne.preprocessing.find_bad_channels_maxwell`, `mne.filter.notch_filter`, and `mne.filter.filter_data` (#1061 by @harrisonritz)
```
larsoner, line 10: Can you try using the double-bracket syntax in other lines above to get the mne-bids-pipeline config values to link to their doc pages?

mne_bids_pipeline/steps/preprocessing/_03_maxfilter.py
```diff
@@ -468,8 +468,9 @@ def run_maxwell_filter(
 
         bids_path_ref_sss = in_files.pop("raw_ref_run_sss")
         raw_exp = mne.io.read_raw_fif(bids_path_ref_sss)
-        rank_exp = mne.compute_rank(raw_exp, rank="info")["meg"]
-        rank_noise = mne.compute_rank(raw_sss, rank="info")["meg"]
+        type_sel = set(cfg.ch_types).intersection({"mag", "meg"}).pop()
```
larsoner, line 471: I think this is more correct / general
```suggestion
        if "grad" in raw_exp:
            if "mag" in raw_exp:
                type_sel = "meg"
            else:
                type_sel = "grad"
        else:
            type_sel = "mag"
```

mne_bids_pipeline/tests/configs/config_ds000248_extra_kws.py
```diff
@@ -0,0 +1,40 @@
+"""MNE Sample Data: M/EEG combined processing."""
```
larsoner, line 1: A couple of things here:

1. `ds000248` is not our fastest dataset to run tests for. For this and other similar extensions, it's best if possible to change some *existing* config if possible. If not possible (which should be fairly rare), then we should add one that executes as quickly as possible.
2. In the future, when we do want to add a new test config (since I don't think we need to here), changes need to be made to `.circleci/config.yml` to get it to actually run. You can see that the new config here wasn't run:

   image

With this in mind, I think in a previous PR [you also added another config](https://github.com/mne-tools/mne-bids-pipeline/pull/1057/files#diff-b9f1b81a3389c1c8c4aa37b686bed501e1410239a0d723d4caa94bfaec7c9933), is there a dataset that runs quicker that you could add a config for instead of `ds000248`? You can see the timings here

https://app.circleci.com/pipelines/github/mne-tools/mne-bids-pipeline/4974/workflows/c3e97eed-cc21-4945-b826-c6650ad8f9b8

and you can cross-ref with the ones listed as being MEG but *not* using maxfilter here:

https://mne.tools/mne-bids-pipeline/stable/examples/examples.html#demonstrated-features

harrisonritz:
@larsoner when I run my tests locally, this line doesn't work:

https://github.com/mne-tools/mne-bids-pipeline/blob/30c779864b3daf3d15261f786d6e5dc905aac9e0/mne_bids_pipeline/_config_utils.py#L458-L464

For `ds004229`, the path I get is: `~/mne_data/ds004229/sub-102/meg/sub-102_meg`, which doesn't match anything. 
The correct path is `~/mne_data/ds004229/sub-102/meg/sub-102_acq-crosstalk_meg.fif`.

The older method works fine:
```python
BIDSPath(subject=subject,
            session=session,
            suffix="meg",
            datatype="meg",
            root=config.bids_root,
        ).meg_calibration_fpath
```

But -- this should mean that more of the tests should be failing, because the default for failing to find this path is to raise an error. 
So I must have something weird about my install, eh? I should have the latest mne_bids (`0.17.0.dev28+g7b046947e`).

mne_bids_pipeline/tests/configs/config_ds000248_extra_kws.py
harrisonritz: this is super helpful, thanks. makes sense the that goal is coverage not isolating issues. revising this now, but running into weird issues in my local install (below).

harrisonritz:
It must just be something funny in my install; I'll reinstall my environment. Maybe worth having an  environment/requirement spec.

hoechenberger:
> Maybe worth having an environment/requirement spec.

Once we switch to uv, we will provide a lock file.

larsoner:
@harrisonritz you might want to do
```
pre-commit install --install-hooks
```
so that you get complaints and fixes locally instead of having to wait for the CI to find them

docs/source/dev.md.inc
```diff
@@ -7,6 +7,7 @@
 - New config option [`mf_extra_kws`][mne_bids_pipeline._config.mf_extra_kws] passes additional keyword arguments to `mne.preprocessing.maxwell_filter`. (#1038 by @drammock)
 - New value `"twa"` for config option [`mf_destination`][mne_bids_pipeline._config.mf_destination], to use the time-weighted average head position across runs as the destination position. (#1043 and #1055 by @drammock)
 - New config options [`mf_cal_missing`][mne_bids_pipeline._config.mf_cal_missing] and [`mf_ctc_missing`][mne_bids_pipeline._config.mf_ctc_missing] for handling missing calibration and cross-talk files (#1057 by @harrisonritz)
+- New config options [`find_bad_channels_extra_kws`][mne_bids_pipeline.find_bad_channels_extra_kws], [`notch_extra_kws`][mne_bids_pipeline.notch_extra_kws], and [`bandpass_extra_kws`][mne_bids_pipeline.bandpass_extra_kws] to pass additional keyword arguments to `mne.preprocessing.find_bad_channels_maxwell`, `mne.filter.notch_filter`, and `mne.filter.filter_data` respectively (#1061 by @harrisonritz)
```
larsoner, line 10:
```suggestion
- New config options [`find_bad_channels_extra_kws`][mne_bids_pipeline._config.find_bad_channels_extra_kws], [`notch_extra_kws`][mne_bids_pipeline._config.notch_extra_kws], and [`bandpass_extra_kws`][mne_bids_pipeline._config.bandpass_extra_kws] to pass additional keyword arguments to `mne.preprocessing.find_bad_channels_maxwell`, `mne.filter.notch_filter`, and `mne.filter.filter_data` respectively (#1061 by @harrisonritz)
```

larsoner:
pre-commit.ci run is still not happy :point_down: 

image

harrisonritz:
Ya, I'm trying to figure out why its not recognizing my most recent commit
https://github.com/mne-tools/mne-bids-pipeline/compare/main...harrisonritz:mne-bids-pipeline:extra-kws

mne_bids_pipeline/_config.py
```diff
@@ -573,6 +573,18 @@
 Auto-detect "noisy" channels and mark them as bad.
 """
 
+
+find_bad_channels_extra_kws: dict[str, Any] = {}
+
+"""
+A dictionary of extra kwargs to pass to `mne.preprocessing.find_bad_channels_maxwell`
+. If kwargs are passed here that have dedicated config settings already, an error
```
larsoner, line 581: If you click on the CI it should go [here](https://results.pre-commit.ci/run/github/167957354/1741024780.korkagCwQPWQdNh7Ez-1kw) where it says:
```
mne_bids_pipeline/_config.py:581:82: W291 Trailing whitespace
    |
579 | """
580 | A dictionary of extra kwargs to pass to `mne.preprocessing.find_bad_channels_maxwell`
581 | . If kwargs are passed here that have dedicated config settings already, an error 
    |                                                                                  ^ W291
582 | will be raised.
583 | For full documentation of the bad channel detection:
    |
    = help: Remove trailing whitespace
```
So:
```suggestion
. If kwargs are passed here that have dedicated config settings already, an error
```

harrisonritz: yup, I had fixed those, but for some reason the pushed version wasn't showing up here (and gave a 500 error on my branch).

`pre-commit run --all-files` is super helpful, I'll do that before I push from now on 🫡

mne_bids_pipeline/tests/configs/config_ds004229.py
```diff
@@ -31,6 +31,24 @@
...
+}
+notch_extra_kws = {
+    "method": "spectrum_fit",
+}
+bandpass_extra_kws = {
+    "fir_window": "blackman",
+}
+
+mf_cal_missing = "warn"
+mf_ctc_missing = "warn"
```
larsoner, line 49: The problem with adding these kwargs to this dataset is that they are *not* missing, so we never know if this actually has any effect, see here where there is no warning:

https://app.circleci.com/pipelines/github/mne-tools/mne-bids-pipeline/4980/workflows/9ebecdd5-f092-4780-8aa1-b428ed3875f2/jobs/77901

You could either find a MEG dataset that we don't currently run MF on and add that as a step. But maybe easier/better would be to mess with [ds003392](https://github.com/mne-tools/mne-bids-pipeline/blob/main/mne_bids_pipeline/tests/configs/config_ds003392.py) in the GHA run by deleting the cross-talk and fine-cal files before it actually tests the dataset. Then we should see that it emits a warning in the logs.

larsoner (changes requested):

mne_bids_pipeline/tests/configs/config_ds004229.py
harrisonritz: GHA = GitHub Actions? Don't know how to do this.

Could I delete the files using `config_ds003392.py`?  something like:
```python
import os
from mne_bids import BIDSPath

bids_path = BIDSPath(
            subject=subject,
            session=session,
            suffix="meg",
            datatype="meg",
            root=bids_root,
            ).match()

file_path = bids_path.meg_calibration_fpath
try:
    os.remove(file_path)
    print(f"File '{file_path}' successfully removed.")
except FileNotFoundError:
    print(f"File '{file_path}' not found.")
```
I assume code in the config automatically get executed?


Alternatively, maybe `warn` could print an affirmative if there is the right calibration file?

harrisonritz: ok, deleted calibration in `config_ds003392.py`

larsoner: > GHA = GitHub Actions? Don't know how to do this.

Yeah it would mean modifying https://github.com/mne-tools/mne-bids-pipeline/blob/650c2637e31c8bb1ef9d1efa487043787ece9835/.github/workflows/tests.yml#L61 with a few lines like
```
- run: |
    rm ~/mne_data/ds003392/some/path/something.fif
    rm 
  if: dataset == 'ds003392'
  name: Remove cross-talk and cal files
```
want to try it? If not I can push.

It's better than modifying the config file itself because it will change the behavior only for the GHA test, whereas the circleci run will proceed as normal with more optimal cross-talk and fine-cal when generating the example HTML

harrisonritz: cool, let me try it out

harrisonritz: alright, pushed a version using GHA

larsoner (approved):
I'll check the GHA logs to make sure things look okay but seems like they should work okay. In the future if we want to do more tweaks to datasets for GHA we can create a script but `run: |` is okay for now.

@drammock did you want to look again before I merge?

.github/workflows/tests.yml
```diff
@@ -60,6 +60,11 @@ jobs:
         id: dataset-cache
       - run: python -m mne_bids_pipeline._download ${{ matrix.dataset }}
         if: steps.dataset-cache.outputs.cache-hit != 'true'
+      - run: |
+          rm ~/mne_data/ds003392/sub-01/meg/sub-01_acq-calibration_meg.dat
+          rm ~/mne_data/ds003392/sub-01/meg/sub-01_acq-crosstalk_meg.fif
```
larsoner, line 65: Just realized we'll need a `-f` here because we cache the dataset. The next time it goes stale and re-caches, the next run the file will already be missing
```suggestion
          rm -f ~/mne_data/ds003392/sub-01/meg/sub-01_acq-calibration_meg.dat
          rm -f ~/mne_data/ds003392/sub-01/meg/sub-01_acq-crosstalk_meg.fif
```

larsoner (changes requested):
@harrisonritz looking at

https://github.com/mne-tools/mne-bids-pipeline/actions/runs/13641756928/job/38132977548?pr=1061#step:9:46

I don't see any warning despite the values being `warn`, can you replicate this locally?

harrisonritz:
this should work now -- it had issues when there was a valid bids path, but its wasn't finding the calibration files. 

IDK why this wasn't showing up before.. I reinstalled my environment today so that it better matches the online test. Maybe some of the issues I was having before were with finding the right bids path.

mne_bids_pipeline/_config.py
```diff
@@ -573,6 +573,18 @@
...
+will be raised.
+For full documentation of the bad channel detection:
+https://mne.tools/stable/generated/mne.preprocessing.find_bad_channels_maxwell
```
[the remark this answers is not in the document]
drammock: #1066

larsoner:
[Great](https://github.com/mne-tools/mne-bids-pipeline/actions/runs/13642552063/job/38135336921?pr=1061#step:9:45)!
```
│23:28:01│ ⏳️ sub-01 WARNING: Could not find Maxwell Filter calibration file from BIDSPath /Users/runner/mne_data/ds003392/sub-01/meg/sub-01_meg. Set to None.
```

larsoner:
Marking for merge when green, thanks in advance for working through this @harrisonritz !

harrisonritz:
thank you guys for the hand-holding -- not an expert in python or git. 
wanted to better understand mne/pipeline; learning a lot

PR #1062, opened 2025-02-28
allow dict for ssp_ecg_channel

drammock:
### Before merging …

### DIFF
--- a/mne_bids_pipeline/steps/preprocessing/_06b_run_ssp.py
+++ b/mne_bids_pipeline/steps/preprocessing/_06b_run_ssp.py
@@ -11,10 +11,12 @@ from mne import compute_proj_epochs, compute_proj_evoked
 from mne.preprocessing import find_ecg_events, find_eog_events
 from mne_bids import BIDSPath
 
+from mne_bids_pipeline._config_import import ConfigError
 from mne_bids_pipeline._config_utils import (
     _bids_kwargs,
     _pl,
     _proj_path,
+    get_ecg_channel,
     get_runs,
     get_subjects_sessions,
 )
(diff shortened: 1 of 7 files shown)


mne_bids_pipeline/steps/preprocessing/_06b_run_ssp.py
```diff
@@ -103,14 +105,18 @@ def run_ssp(
     rejects = dict(ecg=cfg.ssp_reject_ecg, eog=cfg.ssp_reject_eog)
     avg = dict(ecg=cfg.ecg_proj_from_average, eog=cfg.eog_proj_from_average)
     n_projs = dict(ecg=cfg.n_proj_ecg, eog=cfg.n_proj_eog)
-    ch_name: dict[str, list[str] | None] = dict(ecg=None, eog=None)
+    ch_name: dict[str, str | list[str] | None] = dict(ecg=None, eog=None)
```
drammock, line 108: this should have been the case prior to this PR, as `ch_name["ecg"]` was already getting assigned a string a few lines down. Not sure why `mypy` wasn't complaining before.

drammock:
ok @hoechenberger @larsoner this one is ready for review

larsoner (commented):
Obligatry link to https://github.com/mne-tools/mne-bids-pipeline/issues/924#issuecomment-2065293402 where this is discussed in more general terms, but this PR seems compatible with it :+1:

larsoner:
Thanks @drammock !

larsoner:
@berkgercek what version of Python are you running? IIRC f-strings became more permissive in some newer version, but maybe we don't want to require so recent a Python version (not sure if it's 3.10, 3.11, 3.12 etc)

drammock:
[it was added in Python 3.12](https://docs.python.org/3.12/whatsnew/3.12.html#pep-701-syntactic-formalization-of-f-strings). We only require 3.10. I'll make a PR to fix it.

drammock:
see #1078

berkgercek:
Thanks so much, I'm indeed running 3.10 and that was the source of the issue. Didn't think to change versions.

PR #1074, opened 2025-03-19
fixup generation of template config files

drammock:
on `main`, template config files have a couple of bad features:

these lines get commented out, making the file syntactically invalid (unclosed parens):

https://github.com/mne-tools/mne-bids-pipeline/blob/6005b52381203ed3993dc1cc4943e8fd507f993b/mne_bids_pipeline/_config.py#L11-L15

this line does not get commented out, because it starts with "from":

https://github.com/mne-tools/mne-bids-pipeline/blob/6005b52381203ed3993dc1cc4943e8fd507f993b/mne_bids_pipeline/_config.py#L348

This PR:

- [x] fixes commented-out import lines (item 1 above)
- [x] fixes missed commented-out line starting with "from" (item 2 above)
- [x] writes multiline string literals (AKA, the docstrings for each config value) as-is, so that in the template config file, the IDE will (potentially) highlight them a different color, making it much easier to see which lines are changable config values and which are supporting documentation.
- [x] removes a few unnecessary `# noqa E501` comments
- [x] changes a couple cases of `Sequence["str"]` to `Sequence[str]`
- [x] Updates the changelog (`docs/source/dev.md.inc`)

### DIFF
--- a/mne_bids_pipeline/tests/test_documented.py
+++ b/mne_bids_pipeline/tests/test_documented.py
@@ -213,13 +213,37 @@ def test_datasets_in_doc() -> None:
     assert tests == test_names, "CircleCI tests != tests/test_run.py"
 
 
+def _replace_config_value_in_file(fpath: Path, config_key: str, new_value: str) -> None:
+    """Assign a value to a config key in a file, and uncomment the line if needed."""
+    lines = fpath.read_text().split("\n")
+    pattern = re.compile(rf"(?:# )?({config_key}: .* = )(?:.*)")
+    for ix, line in enumerate(lines):
+        if pattern.match(line):
+            lines[ix] = pattern.sub(
+                rf"\1{new_value}",
+                line,  # omit comment marker, change default value to `new_value`
+            )
+            break
+    fpath.write_text("\n".join(lines))
+
+
 def test_config_template_valid(tmp_path: Path, monkeypatch: pytest.MonkeyPatch) -> None:
     """Ensure our config template is syntactically valid (importable)."""
+    monkeypatch.setenv("BIDS_ROOT", str(tmp_path))
     fpath = tmp_path / "foo.py"
     create_template_config(fpath)
-    monkeypatch.setenv("BIDS_ROOT", str(tmp_path))
-    # we need check=False because (at least):
-    #   - `ch_types` fails pydantic validation (its default is `[]` but its annotation
-    #     requires length > 0)
-    #   - `conditions` cannot be None unless `task_is_rest = True` (defaults to False)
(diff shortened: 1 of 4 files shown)


larsoner:
Could we add a smoke test that `create_template_config` actually creates an importable file? That would have at least caught the SyntaxError you mention above right?

drammock:
> Could we add a smoke test that `create_template_config` actually creates an importable file? That would have at least caught the SyntaxError you mention above right?

Good idea, will do that later today

drammock:
> > Could we add a smoke test that `create_template_config` actually creates an importable file? That would have at least caught the SyntaxError you mention above right?
> 
> Good idea, will do that later today

added in [2a358e1](https://github.com/mne-tools/mne-bids-pipeline/pull/1074/commits/2a358e1b177d3b36a9125885ae6b5657bb0d191a)

drammock:
all green!

mne_bids_pipeline/tests/test_documented.py
```diff
@@ -209,3 +211,15 @@ def ignore_unknown(self, node: yaml.Node) -> None:
...
     assert tests == dataset_names, "CircleCI tests != tests/datasets.py"
     assert tests == test_names, "CircleCI tests != tests/test_run.py"
+
+
+def test_config_template_valid(tmp_path: Path, monkeypatch: pytest.MonkeyPatch) -> None:
+    """Ensure our config template is syntactically valid (importable)."""
+    fpath = tmp_path / "foo.py"
+    create_template_config(fpath)
+    monkeypatch.setenv("BIDS_ROOT", str(tmp_path))
+    # we need check=False because (at least):
```
larsoner, line 221: So I think it's okay for these to fail. So how about:

1. Run with `check=True` inside a `pytest.raises`
2. Fix the errors
3. Run with `check=True` (smoke test complete)

If you like the idea but not motivated to do it now, let's just add this as a `TODO` comment and merge!

drammock: > it's okay for these to fail for the reasons you comment below

Agreed. But I do like the idea of catching and handling those (to make sure there aren't any other problems). To me it seems a bit redundant to then also tweak the config and run without error catching, but it's not hard or expensive so I'll add that too (Monday probably).

drammock: OK all green! And I take back what I said about redundancy; I was forgetting that with `pytest.raises`, execution will just stop once a matching error is raised (unlike, say, `warnings.filterwarnings`). So if we don't change the `ch_types` value in the file, the problem with no `conditions` defined won't actually surface in the test.

PR #1078, opened 2025-04-03
avoid backslash in fstring and frontload fsaverage download

### LINKED ISSUE #1063: RFC: should we frontload fsaverage downloading?
I've seen the download of fsaverage from OSF.io fail a couple times recently in CIs. I'm wondering if there's a way to download it earlier in the pipeline instead of waiting until the BEM or group_average steps (but only do it for datasets / tests where we know those steps are going to happen), so that if OSF flakiness is going to fail the test, it will do so earlier.

drammock:
fixes https://github.com/mne-tools/mne-bids-pipeline/pull/1062#issuecomment-2775830266

within-cycle bugfix so no changelog needed.

Closes #1063

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -29,7 +29,7 @@ jobs:
       if: success()
       name: 'Upload coverage to CodeCov'
   caching:
-    name: 'Testing and caching ${{ matrix.dataset }} on ${{ matrix.os }}'
+    name: 'Testing and caching ${{ matrix.dataset }} on ${{ matrix.os }} py${{ matrix.python }}'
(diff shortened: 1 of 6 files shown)


larsoner (approved):

agramfort (approved):

larsoner:
Pushed a change to CIs so that one job runs with Python 3.10 and another with Python 3.13, plus a warning ignore. Will wait for it to come back green, adjust merge reqs, and merge

agramfort (approved):

larsoner:
Going to push a commit to address #1063 since we just hit it again

PR #1088, opened 2025-05-05
Custom metadata

harrisonritz:
From https://github.com/mne-tools/mne-bids-pipeline/issues/1086

Allow for users to pass custom metadata, which is added to epoch info.

1. Users convert a pandas `DataFrame` into a `Dict` (eg in the config file)
2. this `Dict` is converted back to a `DataFrame` at the point of epoch creation.

e.g. in the config:
```python
import pandas as pd
import numpy as np

metadata_path =  "my_dataframe.csv"
metadata_df = pd.read_csv(metadata_path)

epochs_custom_metadata = metadata_df.replace({np.nan: None}).to_dict()

del metadata_path, metadata_df
```

While there are some upcoming methods for putting metadata into the `events.tsv`, this approach allows for analysis-specific metadata (e.g., if some metadata columns apply to specific epoch types)

### todo
- [x] write tests

### Before merging …


hoechenberger:
Hello, thanks for your effort!



> 1. Users convert a pandas `DataFrame` into a `Dict` (eg in the config file)
> 2. this `Dict` is converted back to a `DataFrame` at the point of epoch creation.

Why not simply request a `DataFrame` in the first place?

harrisonritz:
Admittedly, this was because I'm not familiar enough with the config typing to add a `DataFrame` type. 
Dicts were already supported, and the conversion is easy.

If someone could help me add `DataFrame` types to the config files (`mne_bids_pipeline/typing.py`?), then this would be easier for users.

hoechenberger:
sure, just `import pd` and type annotate ad `pd.DataFrame` :) that should do it!

harrisonritz:
Oh duh thats really easy, thanks @hoechenberger. I was trying to pass string literals.

mne_bids_pipeline/_config.py
```diff
@@ -995,6 +996,19 @@
     April 1st, 2021.
 """
 
+# epochs_custom_metadata: dict[str, Any] = dict()
+epochs_custom_metadata: pd.DataFrame = pd.DataFrame()
```
hoechenberger, line 1000: The default should probably be `None`, not an empty DataFrame

harrisonritz:
Added `None` default.

Included a test, which made me realized that this will have issues with multiple subjects/session/tasks. 
So added option to pass as a nested dictionary, e.g., `custom_metadata['sub-x']['ses-x']['task-x']`

mne_bids_pipeline/_config.py
```diff
@@ -995,6 +996,25 @@
     April 1st, 2021.
 """
 
+epochs_custom_metadata: pd.DataFrame | dict[str, Any] | None = None
```
hoechenberger, line 999:
```suggestion
epochs_custom_metadata: pd.DataFrame | dict[str, pd.DataFrame] | None = None
```

harrisonritz: I didn't do this, because its a nested dict with different possible number of levels (so eg could be `dict[str, dict]`).
Would you rather something like `dict[str, Union[dict, pd.DataFrame]]`? 

I don't think this part of the code would be compute intensive, so I lowered my typing paranoia level.

harrisonritz:
@hoechenberger does this look ok? Tricky to provide specific type `epochs_custom_metadata`, since it could be a nested dictionaries with different numbers of levels.

Don't imagine this variable will be a big factor for runtime, so maybe ok for it to have vaguer typing?

larsoner:
We treat warnings as errors and every once in awhile a new one pops up. I'll push a tiny commit to fix it

harrisonritz:
@drammock is it possible that something about the metadata dataframe is causing this (eg when it formats the metadata for the report)? Shouldn't have scalars...

```
epochs_custom_metadata = {
     "ses-anodalpost": pd.DataFrame({"ones": np.ones(253)}),
     "ses-anodalpre": pd.DataFrame({"ones": np.ones(268)}),
     "ses-anodaltDCS": pd.DataFrame({"ones": np.ones(269)}),
     "ses-cathodalpost": pd.DataFrame({"ones": np.ones(290)}),
     "ses-cathodalpre": pd.DataFrame({"ones": np.ones(267)}),
     "ses-cathodaltDCS": pd.DataFrame({"ones": np.ones(297)}),
 }  # number of rows are hand-set
 ```
 
 ref: https://github.com/harrisonritz/mne-bids-pipeline/blob/9e2b74432aeacf16a4af26e7d8cd48bceb635986/mne_bids_pipeline/tests/configs/config_ds001810.py#L28

**Comment by drammock** [comment]
```py
foo = {"ones": np.ones(3)}
type(pd.DataFrame(foo).iloc[0,0])
# numpy.float64  # <--- there's your scalar.
jt.dumps(pd.DataFrame(foo))
# '{"__pandas_dataframe__": {"column_order": ["ones"], "types": ["float64"]}, "index": [0, 1, 2], "ones": [1.0, 1.0, 1.0]}'
```
In other words, json_tricks' approach to serializing dataframes seems like it doesn't store columns as numpy arrays.

harrisonritz:
> @harrisonritz ready to go otherwise from your end?

Yup, should be good to go

larsoner:
@drammock you okay with merging as-is after one last round of review given I've opened #1094, or would you rather we fix it here? It is related to this PR but might not be a blocker for progress since it's really (I think) "just" about how things get logged in the xlsx files.

PR #1092, opened 2025-06-07
fix mf_int_order; add mf_ext_order

harrisonritz:
### Before merging …

### fix mf_int_order; add mf_ext_order

Fixed bug where `mf_int_order` wasn't being passed to the `maxwell_filter` (wasn't included in `mf_kws`).

Also added a config option to supply `mf_ext_order`. Even though `mf_ext_order` can be supplied through `mf_extra_kws`, seems like an important enough parameter to assign its own config argument.

Included non-standard internal & external orders to the `ds000117` config as a test, though we'd have to check the logs to confirm.

Could can reduce the logging of int/ext order if this is an opinionated default.

### DIFF
--- a/mne_bids_pipeline/tests/configs/config_ds000117.py
+++ b/mne_bids_pipeline/tests/configs/config_ds000117.py
@@ -20,6 +20,8 @@ process_empty_room = True
 mf_reference_run = "02"
 mf_cal_fname = bids_root + "/derivatives/meg_derivatives/sss_cal.dat"
 mf_ctc_fname = bids_root + "/derivatives/meg_derivatives/ct_sparse.fif"
+mf_int_order = 7
+mf_ext_order = 4
 
 reject = {"grad": 4000e-13, "mag": 4e-12}
 conditions = ["Famous", "Unfamiliar", "Scrambled"]
--- a/mne_bids_pipeline/_config.py
+++ b/mne_bids_pipeline/_config.py
@@ -686,9 +686,16 @@ account, we are realigning all data to a single position. For this, you can:
 
 mf_int_order: int = 8
 """
-Internal order for the Maxwell basis. Can be set to something lower (e.g., 6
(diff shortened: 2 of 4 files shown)


mne_bids_pipeline/tests/configs/config_ds000117.py
```diff
@@ -20,6 +20,8 @@
 mf_reference_run = "02"
 mf_cal_fname = bids_root + "/derivatives/meg_derivatives/sss_cal.dat"
 mf_ctc_fname = bids_root + "/derivatives/meg_derivatives/ct_sparse.fif"
+mf_int_order = 7
+mf_ext_order = 4
```
larsoner, line 24: So comparing PR to dev

1. https://output.circle-artifacts.com/output/job/4a27513d-fd31-4174-aa37-5fb4c54e3508/artifacts/0/site/examples/ds000117/sub-01_ses-meg_task-facerecognition_report.html#Decoding__full-epochs
2. https://mne.tools/mne-bids-pipeline/stable/examples/ds000117/sub-01_ses-meg_task-facerecognition_report.html#Decoding__full-epochs

decoding performance went down. So I want to try tweaking this so that it hopefully doesn't hurt. I suspect it's the internal order being too small but maybe we'll need to decrease `ext_order` as well...
```suggestion
mf_int_order = 9
mf_ext_order = 4
```

larsoner (approved):
Otherwise looks good, will mark for merge-when-green assuming the decoding perf in the second condition is better with my commit

harrisonritz:
oh shit haha, I couldn't find any additional conditions for the tests (or maybe this is for documentation). Thanks!

larsoner:
It didn't even change very much so we can probably even live with it as is, we'll see. Yes that link is for documentation really but it would be nice but to lose decoding performance! In this repo we're generally kind of stuck with our docs and tests being very closely intertwined

PR #1097, opened 2025-07-07
FIX: sanitize tags

harrisonritz:
### Before merging …

### Description
Currently, contrasts from metadata (e.g., `cond == "A"`) won't create valid report tags due to illegal characters (`"`, `'`).
PR properly sanitizes these tags. Added metadata-based decoding to test suite in `ds001810`.

edit: Wondered if this should be fixed at `mne-python` level, but really only an issue if you auto-generate tags.

### DIFF
--- a/docs/source/dev.md.inc
+++ b/docs/source/dev.md.inc
@@ -35,6 +35,7 @@
 - Fix bug where only having mag sensors would crash compute_rank during maxwell filtering or epoching (#1061 and #1069 by @harrisonritz)
 - Improvements to template config file generation (#1074 by @drammock)
 - Fix bug where `mf_int_order` wasn't passed to `maxwell_filter`. Added config option for `mf_ext_order`. (#1092 by @harrisonritz)
+- Sanitize report tags that contain `"` or `'`, e.g., for certain metadata contrasts (#1097 by @harrisonritz)
 
 ### :books: Documentation
 
--- a/mne_bids_pipeline/_report.py
+++ b/mne_bids_pipeline/_report.py
@@ -568,7 +568,7 @@ def _all_conditions(*, cfg: SimpleNamespace) -> list[str]:
 
 
 def _sanitize_cond_tag(cond: str) -> str:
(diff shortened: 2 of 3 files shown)


harrisonritz:
ya in my local version, I changed `mne-python` to just remove the invalid characters instead of throwing an error

```python
 # Check for invalid characters
    invalid_chars = (" ", '"', "\n")  # we'll probably find more :-)
    cleaned_tags = []
    for tag in tags:
        cleaned_tag = tag
        for invalid_char in invalid_chars:
            cleaned_tag = cleaned_tag.replace(invalid_char, "")
        cleaned_tags.append(cleaned_tag)

    return tuple(cleaned_tags)
```

Figured I would write the PR for `mne_bids_pipeline` instead, because the biggest risk is for auto-generated tags. But IMO better to just silently fix something low stakes like a title/tag rather than raise an error, because eg you can always revise the figure in post-production.

larsoner:
Most expedient thing to do for now would be to fix here and in MNE-Python as above to keep existing behaviors. What you propose would be a change of behavior at the MNE-Python end so we'd need to discuss it more over there. For example something like a `report.bad_tags = "raise" | "warn" | "info"` or something (where our current behavior is `"raise"` and you think it should be / at least have a way to make it `"info"`-level) could work in a backward-compatible way to get you what you want. Or you could advocate for making the default `"info"`... but I'm not totally convinced. I think it's not necessarily a bad thing for the user to be told immediately (by `ValueError`) that they have tried to do something that can't be done, rather than later wonder "why did MNE silently change my tag names?" and have to find out why via docs etc... to someone familiar with HTML maybe the reason is obvious but to an end user maybe not so much. Feel free to open a MNE-Python issue about this if you're motivated and we can get thoughts/opinions from others as well!

harrisonritz:
ya agree, better to fix the special case.
mne-python probably fine as it is.

harrisonritz:
@larsoner looks like the failed tests are unrelated to this PR
(cache download and time-out on unrelated test)

PR #1133, opened 2025-12-04
BUG: Fix bug with missing sessions

### LINKED ISSUE #1134: BUG: sub-average runs even when `--subject` supplied from command line
Found when working on #1133. `sub-average` runs even when `--subject=106` is passed, but it shouldn't. In practice this is only a small issue since you can continue to tweak one subject, then once you run `mne_bids_pipeline config.py` all subjects will run and `sub-average` will be regenerated because the set of `in_files` will change. But we should avoid the temporary inconsistency of a single-subject `sub-average` being created as an intermediate, and the associated inefficiency of calculating `sub-average` in this case.

larsoner:
Hit a minor bug where `mne_bids_pipeline am.py --subject=106` would fail because 106 only had `session-c` but `sessions = ["a", "b", "c"]` and `allow_missing_sessions=True`. This led to a corner case where `sub-average` would then fail.

Closes #1134

### DIFF
--- a/.github/workflows/autopush.yml
+++ b/.github/workflows/autopush.yml
@@ -11,7 +11,7 @@ jobs:
       contents: write
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@v6
         with:
           persist-credentials: true
           token: ${{ secrets.MNE_BOT_TOKEN }}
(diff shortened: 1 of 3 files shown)


larsoner:
... also this is quite a corner case so I don't see a good way to add a test :slightly_frowning_face:

PR #1137, opened 2025-12-08
ENH: add process_raw_clean and short circuit n_jobs=1

### LINKED ISSUE #925: Don't perform ICA cleaning on raw data by default

larsoner:
### Before merging …

I am doing a lot of single-subject processing with `n_jobs=8` in my config but don't want to keep passing `n_jobs=1`. I'd also like to save a bit of disk space. It was easy to fix #925 while fixing #1023.

Added `process_raw_clean=False` (to kind of follow `process_rest` and `process_empty_room`) to our slowest test runs where it might have an effect to hopefully speed them up.

Closes #925

### DIFF
--- a/mne_bids_pipeline/steps/freesurfer/_02_coreg_surfaces.py
+++ b/mne_bids_pipeline/steps/freesurfer/_02_coreg_surfaces.py
@@ -11,10 +11,10 @@ from types import SimpleNamespace
 import mne.bem
 
 from mne_bids_pipeline._config_utils import (
+    _get_ss,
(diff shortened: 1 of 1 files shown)


drammock (approved):
just one question

Issue #760, opened 2023-07-07
ENH: Add mf_bads option


larsoner:
One solution that should work for this would be something like:
```
mf_bads: Literal["run", "union"] = "run"
```
where changing to `"union"` it will use the union of all bads across all data that will be maxwell filtered (runs + noise + rest). This actually in some sense is a safer option because the autobad only sometimes finds bad channels that it should find, so by combining across runs with a union is probably safest / most conservative.

_Originally posted by @larsoner in https://github.com/mne-tools/mne-bids-pipeline/issues/759#issuecomment-1623806155_

hoechenberger:
Makes sense to me, but I wouldn't label it a bug fix but rather a behavior change

PR #1141, opened 2025-12-10
ENH: Allow multiple tasks to be processed jointly

### LINKED ISSUE #820: ENH: Process multiple tasks jointly?
I have a dataset that I'm BIDSifying. For each subject we have the following based on how I'm BIDSifying the data:

1. Resting state data (`task="rest"`)
2. One run of auditory localizer data (`task="localizer"`)
3. Several runs of task data (`task="speech", run="01" | "02" | ...`)

I'd like to preprocess these all at once using the same SSP or ICA. It would be nice to be able to do `task = ["rest", "localizer", "speech"]` for example. But then there's an issue where each task can have a different set of runs, so maybe we'd also need to allow `runs` to be a `dict` specifying runs on a per-task basis. Thoughts @hoechenberger ? If no objection then I can try implementing this and running it on my data.

Things will get trickier later when I might want to be able to epoch data differently for the localizer vs the speech for example, but at least getting the preprocessing done for all of these (and probably just doing epoching for localizer to start) would be a huge help.

What wouldn't work is processing the localizer separately from the speech task since the spatial processing steps (SSP/ICA) need to be identical for them, but that looks like my only option currently.

### LINKED ISSUE #1174: `get_runs()` incorrectly scans all tasks instead of only the configured task, causing errors with BIDS-compliant datasets
## Problem

The `_get_runs_all_subjects_cached()` function in `_config_utils.py` (lines ~165) scans for run entities across **all tasks** in a subject directory, rather than filtering by the configured `task` parameter. This causes the pipeline to fail when processing BIDS-compliant datasets where different tasks have different run structures.

## Expected Behavior

When `task="taskA"` is configured, the pipeline should only look for run entities in files matching `sub-*_task-taskA_*.fif`, not scan runs from other tasks.

## Current Behavior

The pipeline scans for runs across all tasks and applies them to the configured task, resulting in:
```
File does not exist:
/path/to/bids/sub-002/meg/sub-002_task-taskA_run-1_meg
Did you mean one of:
sub-002_task-taskA_meg.fif  <-- correct file (no run entity)
sub-002_task-taskB_run-1_meg.fif
```

## Minimal Reproducible Example

**Dataset structure (BIDS-compliant):**
```
sub-002/
    meg/
        sub-002_task-taskA_meg.fif        # No run entity
        sub-002_task-taskB_run-1_meg.fif  # Has run entities
        sub-002_task-taskB_run-2_meg.fif
```

**Config:**
```python
task = "taskA"
runs = "all"
subjects = ["002"]
use_maxwell_filter = True
```

**Result:** Pipeline finds runs `['1', '2']` from other tasks and tries to process `sub-002_task-taskA_run-1_meg.fif` which doesn't exist.

## Root Cause

In `_config_utils.py` at line ~165:
```python
valid_runs_subj = _get_entity_vals_cached(
    config.bids_root / f"sub-{subject}",
    entity_key="run",
    ignore_datatypes=_get_ignore_datatypes(config),
)
```

The `_get_entity_vals_cached()` call does not pass `ignore_tasks` parameter, so it returns runs from ALL tasks.

## Proposed Fix

Filter by task before searching for runs:

```python
# Filter by task to avoid picking up runs from other tasks
if config.task and config.task != "":
    all_tasks = _get_entity_vals_cached(
        config.bids_root / f"sub-{subject}",
        entity_key="task",
        ignore_datatypes=_get_ignore_datatypes(config),
    )
    ignore_tasks = tuple(set(all_tasks) - {config.task})
else:
    ignore_tasks = None
    
valid_runs_subj = _get_entity_vals_cached(
    config.bids_root / f"sub-{subject}",
    entity_key="run",
    ignore_datatypes=_get_ignore_datatypes(config),
    ignore_tasks=ignore_tasks,  # Add this parameter
)
```

## Impact

This bug prevents processing of valid BIDS datasets where:
- One task has no run entity (e.g., single recording session)
- Other tasks have run entities (e.g., multiple recording sessions)

This is a valid BIDS structure as the run entity is optional and should be used only when a task is repeated within a session.

## Environment

- mne-bids-pipeline version: 1.9.0
- Python version: 3.11
- Operating system: Linux

## Additional Context

The BIDS specification allows files without run entities when there's only one recording of that task. The pipeline should respect this and only look for runs within the specific task being processed, not across all tasks in the subject directory.

larsoner:
### Before merging …

- [x] Update type defs
- [x] Updated `docs/source/dev.md.inc`
- [x] Handle `epochs`-based `cov` somehow... probably by making sure `baseline`s are the same and compatible and if not, tell the user to use a custom function instead of `"epochs"`
- [x] Ensure report section titles only get task added if there is more than one task
- [x] Add processing of multiple tasks jointly (tested locally with dataset that needs it)
  - [x] Initial
  - [x] Preproc
  - [x] Evoked
  - [x] Source est
- [x] Ensure all report sections are tagged by task type
- [x] ~~Ensure `preprocessing/*` uses union of all bad channels across all tasks (including noise!):~~ Dropped in favor of existing #760
- [x] Get CIs green
- [ ] Add dataset that demonstrates this capability

Just opening to make sure nothing is broken yet. It shouldn't be since this should just relax type def checks in a way that shouldn't matter, but :shrug: 

Closes #820
Closes #1174

### DIFF
--- a/mne_bids_pipeline/_config_utils.py
+++ b/mne_bids_pipeline/_config_utils.py
@@ -445,10 +445,10 @@ def get_tasks(config: SimpleNamespace) -> list[str | None]:
             tasks.append(config.task)
         else:
             tasks.extend(config.task)
-        if set(tasks) - set(_valid_tasks):
+        if missing_tasks := (set(tasks) - set(_valid_tasks)):
             raise ValueError(
                 "The following requested tasks were not found in the dataset: "
-                f"{set(tasks) - set(_valid_tasks)}"
+                f"{missing_tasks}"
             )
     else:
         if not _valid_tasks:
--- a/mne_bids_pipeline/_run.py
+++ b/mne_bids_pipeline/_run.py
@@ -327,7 +327,7 @@ class ConditionalStepMemory:
 
 
 def save_logs(*, config: SimpleNamespace, logs: Iterable[pd.Series]) -> None:
-    all_tasks = "-".join(map(str, config.all_tasks))
+    all_tasks = "+".join(map(str, config.all_tasks))
     fname = config.deriv_root / f"task-{all_tasks}_log.xlsx"
 
     # Get the script from which the function is called for logging


drammock:
I won't have time to review today, but I *do* have permission to use one subject from the PREK dataset (and I have the files), so maybe worth waiting for that. I hope to get to it early next week.

larsoner:
I think two or three subjects with individual MRIs would be great, assuming it doesn't add too much to the dataset size. Would be good to create and include stuff that will be needed like BEMs and source spaces to save computation time, too. Those steps are unchanged here I think so making them be computed here would add a lot of extra recomputation for all CI runs all the time without much benefit testing-wise

drammock:
SGTM. I was already planning to do "slim" versions of the MRIs, and precompute the BEMs.

drammock:
> @drammock WDYT about reviewing and merging this without a dataset? I'm using it successfully on related data and it at least works as well on non-multi-task data. We could have multi-task data as a follow-up PR when we find time.

yeah that makes sense. I got stymied when creating the dataset (some combination of defacing + slimming the MRIs + recreating the BEMs went sideways) so better not to let that be a blocker, since you're using it locally and it works.

mne_bids_pipeline/_config_utils.py
```diff
@@ -377,62 +378,84 @@ def get_runs_tasks(
...
+    tasks: list[str | None] = list()
+    if config.task:
+        if isinstance(config.task, str):
+            tasks.append(config.task)
+        else:
+            tasks.extend(config.task)
+        if set(tasks) - set(_valid_tasks):
+            raise ValueError(
+                "The following requested tasks were not found in the dataset: "
+                f"{set(tasks) - set(_valid_tasks)}"
```
drammock: [nitpick] walrus

```suggestion
        if missing_tasks := (set(tasks) - set(_valid_tasks)):
            raise ValueError(
                "The following requested tasks were not found in the dataset: "
                f"{missing_tasks}"
```

PR #1152, opened 2026-01-09
BUG: Remove unused config vars from steps

### LINKED ISSUE #1148: BUG: Some config values are unused but we pass them anyway
Over in #1141 I discovered a whole bunch of config values that will never be used in this step:

https://github.com/mne-tools/mne-bids-pipeline/blob/2f11dc89899a6f8996c25aed41a518b489ae13ae/mne_bids_pipeline/steps/preprocessing/_06a2_find_ica_artifacts.py#L590-L596

This not only makes the code needlessly complex, it can lead to false alarm cache misses. It would be great to have some way to detect this sort of thing. Two ideas come to mind to track this "namespace coverage" issue:

1. `ast`. Parse each step, and dig into the code to see what is likely to be used. This could hopefully be fairly simple using a `NodeVisitor` since we can look for `ast.Attribute` and match potentially used names to the `config.*` names, also detected using `ast` or maybe `regex` on the source. But given how hard changes to `docs` are, I'm not sure how easy this would actually be.
2. Modify our uses of `SimpleNamespace` to add a new `SimpleNamespaceTracking(SimpleNamespace)` class whose `__getattr__` keeps track in some dunder var of which names actually get accessed. Then we can add some magic in the `_run.py` code to somehow keep track of what wasn't used by pulling it from the `cfg` in a given function (this is doable since we cache / route all calls through `StepMemory` or whatever it's called). Then somehow we have to allow for some datasets to *not* use some vars, and others to need to use those (e.g., tests that use ICA vs those that use SSP).

Given the complexity of (2), I think it's worth checking to see how feasible (1) is.

larsoner:
### Before merging …

- [x] Locally check `pytest mne_bids_pipeline -k ds000248`

Closes #1148

Turned out not to be too bad with some simple `regex` and string searches.

Will create a conflict with #1141 but I'm happy to merge `main` into that one (it'll help make that diff a tiny bit smaller anyway). I'm also locally testing `pytest mne_bids_pipeline -k ds000248` so we can safely ignore the caching issue there assuming the other CI runs come back green :crossed_fingers:


mne_bids_pipeline/steps/preprocessing/_06a1_fit_ica.py
```diff
@@ -363,7 +363,6 @@ def get_config(
         task_is_rest=config.task_is_rest,
         ica_h_freq=config.ica_h_freq,
         ica_l_freq=config.ica_l_freq,
-        h_freq=config.h_freq,
```
larsoner: Notably, all 29 red lines in these `steps` files are what the new test told me to get rid of -- so if it comes back green, it did its job well!

PR #1158, opened 2026-01-15
BEH: Use union of bads when Maxwell filtering

### LINKED ISSUE #326: ENH: issue warning if user sets derivatives path to bids root.
The name says it all.

### LINKED ISSUE #760: ENH: Add mf_bads option

### LINKED ISSUE #767: BUG: ArbitraryContrast is not tested
In working on streamlining some stuff in #765 I noticed that `ArbitraryContrast` does not seem to be used in any tests:
```
$ git grep "contrasts = "
...
mne_bids_pipeline/steps/sensor/_99_group_average.py:        decoding_contrasts = []
mne_bids_pipeline/tests/configs/config_ERP_CORE.py:    contrasts = [("unrelated", "related")]
mne_bids_pipeline/tests/configs/config_ERP_CORE.py:    contrasts = [("response/incorrect", "response/correct")]
mne_bids_pipeline/tests/configs/config_ERP_CORE.py:    contrasts = [("response/right", "response/left")]  # contralateral vs ipsi
mne_bids_pipeline/tests/configs/config_ERP_CORE.py:    contrasts = [("stimulus/deviant", "stimulus/standard")]
mne_bids_pipeline/tests/configs/config_ERP_CORE.py:    contrasts = [("stimulus/right", "stimulus/left")]  # Contralteral vs ipsi
mne_bids_pipeline/tests/configs/config_ERP_CORE.py:    contrasts = [("stimulus/face/normal", "stimulus/car/normal")]
mne_bids_pipeline/tests/configs/config_ERP_CORE.py:    contrasts = [("stimulus/target", "stimulus/non-target")]
mne_bids_pipeline/tests/configs/config_ds000117.py:contrasts = [
mne_bids_pipeline/tests/configs/config_ds000246.py:contrasts = [("deviant", "standard")]
mne_bids_pipeline/tests/configs/config_ds000248_base.py:contrasts = [("Visual", "Auditory"), ("Auditory/Right", "Auditory/Left")]
mne_bids_pipeline/tests/configs/config_ds000248_no_mri.py:contrasts = [("Auditory/Right", "Auditory/Left")]
mne_bids_pipeline/tests/configs/config_ds001810.py:contrasts = [("61450", "61511")]
mne_bids_pipeline/tests/configs/config_ds001971.py:contrasts = [("AdvanceTempo", "DelayTempo")]
mne_bids_pipeline/tests/configs/config_ds003392.py:contrasts = [("incoherent", "coherent")]
mne_bids_pipeline/tests/configs/config_eeg_matchingpennies.py:contrasts = [("raised-left", "raised-right")]
```
You can't see `ds000117` here but it's a list of tuple of str. None of these use the `TypedDict` that is `ArbitraryContrast`, so it's likely either broken or going to break.

@hoechenberger or @agramfort do you want to open a quick PR to change one of the `tests/configs/*` to use an `ArbitraryContrast` in some meaningful way? If not I can add a potentially bad/useless example somewhere...

larsoner:
### Before merging …

- [x] Use union of bads for maxwell filtering
- [x] Change `add_bads=None` to `add_bads=False` default (more explicit!)
- [x] Don't triage `add_bads` based on `_do_mf_autobad` -- even if we don't do MF autobad then the bad file that gets added matches what's read from the file anyway, so no harm in using it
- [x] Remove some `in_files.pop(<name>, None)` in favor of `if <condition>: in_files.pop(<name>)` to be more explicit about when files should be used or not
- [x] Sneak in a fix for #767 since it's trivial
- [x] Sneak in a fix for #326 since it's trivial

Closes #767
Closes #760
Closes #326

### DIFF
--- a/mne_bids_pipeline/steps/preprocessing/_03_maxfilter.py
+++ b/mne_bids_pipeline/steps/preprocessing/_03_maxfilter.py
@@ -14,8 +14,8 @@ It is critical to mark bad channels before Maxwell filtering.
 The function loads machine-specific calibration files.
 """
 
-from collections import defaultdict
 import gc
+from collections import defaultdict
 from copy import deepcopy
 from types import SimpleNamespace
 


larsoner:
@drammock this one should be good to go!

mne_bids_pipeline/steps/preprocessing/_03_maxfilter.py
```diff
@@ -48,8 +51,117 @@
...
+        bids_path_bads_in = in_files.pop(f"{in_key}-bads")
+        bads_tsv = pd.read_csv(bids_path_bads_in.fpath, sep="\t", header=0)
+        assert list(bads_tsv.columns) == ["name", "reason"], bads_tsv.columns
+        counts.append(len(bads_tsv))
+        for name, reason in bads_tsv.itertuples(index=False):
+            assert isinstance(name, str)
+            assert isinstance(reason, str)
+            bads_dict.setdefault(name, list())
+            if reason not in bads_dict[name]:
+                bads_dict[name] = bads_dict.get(name, []) + [reason]
```
drammock: [nitpick] if `bads_dict = defaultdict(list)`, then this becomes:

```suggestion
            if reason not in bads_dict[name]:
                bads_dict[name].append(reason)
```
which feels like a more natural idiom to me. You could even eliminate the conditional if it were OK to use `defaultdict(set)` (though this presumably would require later casting to list, or later changes to handle sets instead)

Feel free to ignore though; what's here should work fine

PR #1159, opened 2026-01-16
ENH: Add cov_rank option

### LINKED ISSUE #1147: ENH: Allow computation of rank for inversion
The `info` rank can sometimes be wrong when there is a lot of head movement. I find something [compute_rank](https://mne.tools/stable/generated/mne.compute_rank.html) to be more robust. I propose adding:
```
cov_rank: Literal["info"] | dict[str, Any]
```
where `"info"` is the current behavior, and `dict[str, Any]` are `kwargs` to be passed to `mne.compute_rank(inst, **kwargs)` run on the data (epochs or concatenated raw, TBD which makes more sense -- maybe the min over all tasks and runs would be best!). I would probably use for example:
```
cov_rank = dict(tol=1e-4, tol_kind="relative")
```
for most data with movement compensation.

This would be accomplished via a new step `_06_compute_rank.py` that would be very quick for `info` and slower for `compute_rank` and output a JSON dump of the returned `rank` dict. We would `git mv _06_make_cov.py _07_make_cov.py`, and use that rank during cov computation and inversion.

larsoner:
Closes #1147

### Before merging …

### DIFF
--- a/docs/source/dev.md.inc
+++ b/docs/source/dev.md.inc
@@ -20,7 +20,7 @@
 ### :warning: Behavior changes
 
 - The pipeline will now raise an error if a loaded `SourceSpaces` object has a `._subject` attribute different from what the pipeline expects / would have used if creating the `SourceSpaces` anew. (#1056 by @drammock)
-- The path passed to `noise_cov` callbles is no longer the evoked dataset but instead epochs. Note that the epochs have `processing="clean"` in the `BIDSPath`. (#1159 by @larsoner)
+- The path passed to `noise_cov` callables is no longer the evoked dataset but instead epochs. Note that the epochs have `processing="clean"` in the `BIDSPath`. (#1159 by @larsoner)
 
 [//]: # (- Whatever (#000 by @whoever))
 


larsoner:
Needs https://github.com/mne-tools/mne-python/pull/13595

docs/source/dev.md.inc
```diff
@@ -15,10 +15,12 @@
 - New config option [`process_raw_clean`] to allow not fully cleaning raw data to save computing resources (#1136 by @larsoner)
 - Automatically avoid using `joblib.Parallel` when only one iteration is needed (and `dask` is not being used), allowing multithreaded linear algebra operations to be used (#1136 by @larsoner)
 - Add support for subject and session-specific EOG channels in config and change ssp_ecg_channel to accept 'default' key when dict (#1151 by @skjerns)
+- Add support for computing data rank using `mne.compute_rank` via [`cov_rank`][mne_bids_pipeline._config.cov_rank] (#1159 by @larsoner)
 
 ### :warning: Behavior changes
 
 - The pipeline will now raise an error if a loaded `SourceSpaces` object has a `._subject` attribute different from what the pipeline expects / would have used if creating the `SourceSpaces` anew. (#1056 by @drammock)
+- The path passed to `noise_cov` callbles is no longer the evoked dataset but instead epochs. Note that the epochs have `processing="clean"` in the `BIDSPath`. (#1159 by @larsoner)
```
drammock:
```suggestion
- The path passed to `noise_cov` callables is no longer the evoked dataset but instead epochs. Note that the epochs have `processing="clean"` in the `BIDSPath`. (#1159 by @larsoner)
```

PR #1166, opened 2026-01-21
Move to lefthook, ty, and dependency groups

### LINKED ISSUE #1165: ty + lefthook?
`ty` is much faster than mypy (and even found an error `mypy` didn't), see https://github.com/openneuro-py/openneuro-py/pull/260 and https://github.com/openneuro-py/openneuro-py/pull/253. @hoechenberger worth making the switch here, too, you think?

larsoner:
### Before merging …

- [x] Move to dev groups
- [x] Move to `lefthook`
- [x] Move to `ty`

Closes #1165

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -21,8 +21,9 @@ jobs:
     - uses: actions/setup-python@v6
       with:
         python-version: "3.12"
+    - uses: astral-sh/setup-uv@v7
     - run: pip install --upgrade pip
-    - run: pip install -ve . --group dev uv "mne-bids[full] @ git+https://github.com/mne-tools/mne-bids@main" "openneuro-py @ https://api.github.com/repos/openneuro-py/openneuro-py/zipball/main" codespell tomli --only-binary="numpy,scipy,pandas,matplotlib,pyarrow,numexpr"
+    - run: pip install -ve . --group dev "mne-bids[full] @ git+https://github.com/mne-tools/mne-bids@main" "openneuro-py @ https://api.github.com/repos/openneuro-py/openneuro-py/zipball/main" codespell tomli --only-binary="numpy,scipy,pandas,matplotlib,pyarrow,numexpr"
     - run: lefthook run pre-commit --all-files  # fail early if code style issues!
     - run: pytest mne_bids_pipeline -m "not dataset_test"
     - uses: codecov/codecov-action@v5


larsoner:
@hoechenberger it occured to me that there is no auto-update for these hooks now that we've replaced `pre-commit.ci` with `lefthook`. Should we add a scheduled run that does `uv lock --upgrade` and opens a PR if anything changes? Can do it here pretty easily since `mne-bot` exists

larsoner:
@drammock looks like @hoechenberger is okay with using this here (see comment edit above), feel free to review + merge if you're happy

.github/workflows/tests.yml
```diff
@@ -22,8 +22,8 @@ jobs:
       with:
         python-version: "3.12"
     - run: pip install --upgrade pip
-    - run: pip install -ve .[tests] "mne-bids[full] @ git+https://github.com/mne-tools/mne-bids@main" "openneuro-py @ https://api.github.com/repos/openneuro-py/openneuro-py/zipball/main" codespell tomli --only-binary="numpy,scipy,pandas,matplotlib,pyarrow,numexpr" pre-commit
-    - run: pre-commit run --all  # fail early if code style issues!
+    - run: pip install -ve . --group dev uv "mne-bids[full] @ git+https://github.com/mne-tools/mne-bids@main" "openneuro-py @ https://api.github.com/repos/openneuro-py/openneuro-py/zipball/main" codespell tomli --only-binary="numpy,scipy,pandas,matplotlib,pyarrow,numexpr"
```
drammock: why `pip install uv` here instead of the uv setup action in a prior step?

.github/release.yaml
```diff
@@ -4,4 +4,5 @@ changelog:
     authors:
       - dependabot[bot]
       - pre-commit-ci[bot]
+      - mne[bot]
       - mne-bot
```
drammock: are these different?

cbrnr:
AFAIK dependabot does not only do security updates, at least with the settings I've been using. I've so far run into breaking changes due to automatic tooling updates only once, so the convenience of not having to manually do that slightly outweighs the risks IMO.

.github/workflows/tests.yml
larsoner: I figured since we're using `setup-python` already it wasn't worth having `setup-uv` in there as well but I guess we can!

pyproject.toml
```diff
@@ -1,93 +1,106 @@
-[build-system]
-requires = ["hatchling", "hatch-vcs"]
-build-backend = "hatchling.build"
```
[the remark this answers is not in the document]
drammock: ok just leave it as is for now

drammock (approved):

pyproject.toml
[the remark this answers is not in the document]
hoechenberger: I didn't experience any issues so far

PR #1170, opened 2026-01-22
MAINT: Update lefthook config

larsoner:
We should be operating in stages instead of all in parallel. Some can still be in parallel when they operate on distinct sets of files.


larsoner:
Failure is just https://github.com/OpenNeuroOrg/openneuro/issues/3726, in it goes!

PR #1184, opened 2026-02-17
Take online reference channel into account for average reference + add re-referencing for raw/continuous data

### LINKED ISSUE #1169: Re-referencing for raw (continuous) data only seems to be applied in certain cases
Hi,

I'm using the mne-bids-pipeline for the preprocessing of my current data set and noticed that the `eeg_reference` specified in the config file only seems to be applied for the epoched data in `_07_make_epochs.py` but not for the raw/continuous data: 
https://github.com/mne-tools/mne-bids-pipeline/blob/687e4d31c509ea9ea4d7db4116f77d9cfbda0e1d/mne_bids_pipeline/steps/preprocessing/_07_make_epochs.py#L204-L207

If I see it correctly, the raw/continuous data is only re-referenced (to average) in the case of using `ica_use_icalabel`:
https://github.com/mne-tools/mne-bids-pipeline/blob/687e4d31c509ea9ea4d7db4116f77d9cfbda0e1d/mne_bids_pipeline/steps/preprocessing/_08a_apply_ica.py#L217-L218

But please correct me if I'm wrong.

For my preprocessing, I plan to add code to re-reference the raw/continuous data (to the reference specified in the config file) in our fork in `_08a_apply_ica.py` after applying the ica and before saving the raw data. If this is alo of interest for you I am happy to do a small PR upstream.

Best,
Judith

jschepers:
To ensure that the correct number of channels is considered for the average re-referencing (https://github.com/mne-tools/mne-python/issues/13618) I added the following functionality:
- I introduced a function `set_initial_average_reference` that optionally adds the online reference (specified by the user in the config file) to the data before computing an average reference. The function uses an average reference projection but does not apply it yet.
- The application of the projection happens outside of the function. Afterwards, optionally one can drop the reconstructed online reference channel again. Note: if it's not dropped then there will be one more channel in the data than before.
- I also implemented that the reference specified in the config file is also applied to the raw/continuous data (not only to the epoched data). Thereby this issue fixes #1169.
- After a few discussions we decided to add the re-referencing before the application of the ICA and not do add it again afterwards because currently it's not completely clear to us whether that's desired or even appropriate.  

I'm happy to hear what you think about this approach.

Cheers,
Judith


larsoner:
Sorry for the (very) slow response here @jschepers ... yes I think in principle this is okay. Do you want to keep working on this, or would it help for me to take over to get this in?

PR #1216, opened 2026-04-16
ENH: Speed up caching checks

larsoner:
### Before merging …

| Orig | xlsx shortcut | sidecar shortcut |
| -- | -- | -- |
| 1:26 | 1:07 | 0:45 |
| image | image | image |

I expect the sidecar update to help but testing is still in progress.


larsoner:
Okay this overall improves caching speed in a real-world test by a factor of about 2. Ready for review/merge from my end @drammock !

mne_bids_pipeline/steps/init/_01_init_derivatives_dir.py
```diff
@@ -12,12 +12,15 @@
 from mne_bids_pipeline._logging import gen_log_kwargs, logger
 from mne_bids_pipeline._report import _open_report, _report_path
 from mne_bids_pipeline._run import _prep_out_files_path, failsafe_run
-from mne_bids_pipeline.typing import OutFilesT
+from mne_bids_pipeline.typing import InFilesT, OutFilesT
 
 
-@failsafe_run()
-def init_dataset(cfg: SimpleNamespace, exec_params: SimpleNamespace) -> OutFilesT:
+@failsafe_run(get_input_fnames=lambda *, cfg: dict())
```
drammock: wow, it took me a while to parse this line, something about `*, cfg` not being enclosed in parens (like it would be in a normal function def) really threw me off.  I'd suggest making the lambda into a proper function defined above, but I'm not sure what to call it. Maybe:

```py
def null_input_fnames(*, cfg):
    return dict()
```

WDYT?

drammock (approved):

mne_bids_pipeline/steps/init/_01_init_derivatives_dir.py
larsoner: Pushed a more standard `get_input_fnames_init_dataset` function. Only 3 lines and more standard. Agreed it's much more readable!

PR #1219, opened 2026-04-20
ENH: Speed up sidecar searches

larsoner:
Just a few more speed tweaks

1. Avoid `inspect.stack()`, it's expensive. Use `currentframe().f_back` instead
2. Pin to MNE-BIDS version that has `find_matching_sidecar` (3+ years old, should be fine)
3. Cache `find_matching_sidecar`
4. We don't need to `deepcopy` our `kwargs`, that's from old code (we only modify if we actually call the underlying func, which we won't do at this level of code as we know it's in the call cache)

With this PR we're down to < 30s, which is nice since we were ~90s last week before optimizations!

After this PR, the vast majority of time is now spent in `joblib` caching of function code etc. There is perhaps some way to speed this up, too, since we know that in a given `mne_bids_pipeline` invocation this code will not change... but that's probably for another day :smile: 

| main | PR |
| -- | -- |
| image | image |

### DIFF
--- a/mne_bids_pipeline/_run.py
+++ b/mne_bids_pipeline/_run.py
@@ -189,7 +189,7 @@ class ConditionalStepMemory:
                 hashes.append(hash_(k, v))
                 # also hash the sidecar files if this is a BIDSPath and
                 # MNE-BIDS is new enough
-                if not (self.sidecars and hasattr(v, "find_matching_sidecar")):
+                if not (self.sidecars and isinstance(v, BIDSPath)):
                     continue
                 # from mne_bids/read.py
                 # The v.datatype is maybe not right, might need to use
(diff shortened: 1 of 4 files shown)


mne_bids_pipeline/_run.py
```diff
@@ -189,7 +191,7 @@ def wrapper(*args: list[Any], **kwargs: dict[str, Any]) -> bool:
                 hashes.append(hash_(k, v))
                 # also hash the sidecar files if this is a BIDSPath and
                 # MNE-BIDS is new enough
```
drammock:
```suggestion
# also hash the sidecar files if this is a BIDSPath
```
guaranteed new enough now, because you pinned it

drammock (approved):

PR #1220, opened 2026-04-20
move set_eeg_montage after drop_channels

dnacombo:
to match config statement:
```
# !!! warning
#     If the data contains channel names that are not part of the template montage, the
#     pipeline run will fail with an error message. You must either pick a different
#     montage or remove those channels via
#     [`drop_channels`][mne_bids_pipeline._config.drop_channels] to continue.

```

### Before merging …


larsoner:
Looks reasonable to me! Can you add a `docs/source/dev.md.inc` line about the change in the order of these operations? Or I can do it quickly if you'd prefer. I just want some note in there in case there are some unexpected results.

And I assume this fixes things for your dataset(s)?

dnacombo:
> And I assume this fixes things for your dataset(s)?

It does.

I don't know how to finalize this. Your help is welcome. Thanks!

larsoner:
Looks good to me, fixed the merge conflict and marking for merge-when-green, thanks in advance @dnacombo !

PR #1223, opened 2026-04-20
BUG: Fix bug with Evoked plots for sub-average

larsoner:
### Before merging …

Fixes a bug introduced in 1.10.0 where the `sub-average` report didn't contain grand-average Evoked plots because these were written into the report of the last subject instead :facepalm: 

We should probably cut a release after this. @drammock if you agree, should we just make it a 1.10.1 (with the speed enhancements) or a new 1.11.0?

### DIFF
--- a/docs/source/dev.md.inc
+++ b/docs/source/dev.md.inc
@@ -14,7 +14,7 @@
 ### :bug: Bug fixes
 
 - Fixed a bug where the group average step would errantly write its sensor averaged Evoked plots to the last subject's report.
-  To fix this, delete the `derivatives/mne-bids-pipeline/sub-average/` and `derivatives/mne-bids-pipeline/sub-<LAST>` folders and rerun the pipeline.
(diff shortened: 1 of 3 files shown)


drammock:
I'd call it a point release. No significant new features / API changes

docs/source/dev.md.inc
```diff
@@ -11,7 +11,11 @@
 
 [//]: # (### :package: Requirements)
 
-[//]: # (### :bug: Bug fixes)
+### :bug: Bug fixes
+
+- Fixed a bug where the group average step would errantly write its sensor averaged Evoked plots to the last subject's report.
+  To fix this, delete the `derivatives/mne-bids-pipeline/sub-average/` and `derivatives/mne-bids-pipeline/sub-<LAST>` folders and rerun the pipeline.
```
drammock:
```suggestion
  To fix this after-the-fact, delete the `derivatives/mne-bids-pipeline/sub-average/` and `derivatives/mne-bids-pipeline/sub-<LAST>` folders and rerun the pipeline.
```

drammock (approved):

PR #1225, opened 2026-04-22
ENH: Improve possible dataset error discoverability

larsoner:
### Before merging …

Our default of `mne_log_level = "error"` I think is unsafe -- it's too easy for users to hit likely errors in their datasets and not notice. So this PR:

1. Changes the default to "warning"
2. Updates some option values that should be usable
3. Adds a `ignore_warnings` config param to make it easy to choose warning messages to suppress
4. Fixes us actually setting our own internal logging level.
5. Moves `read_raw_bids_verbose`, `mne_log_level`, and `log_level` into `exec_params`, which is the correct scope for them: they *should not* affect run caching (previously at least `read_raw_bids_verbose` would have) and outputs/results, just the messages that get printed to the console

I think evidence for the utility is already here in this PR when combined with our tests treating warnings-as-errors:

1. See the `ignore_warnings` I ended up adding for `ds0001117` -- you can see multiple issues with it as a BIDS dataset and/or our reading of it
2. See the change to `_07_make_epochs.py`, where I needed a new `_ignore_warnings` context (which caused me to have to think about whether or not it was appropriate/safe!)

I suspect as all datasets run through I'll hit a lot more warnings for them and need to add `ignore_warnings` to a number of them. But we'll see...


larsoner:
Okay I think this is ready @drammock !

mne_bids_pipeline/tests/configs/config_MNE_funloc_data.py
```diff
@@ -9,6 +9,8 @@
 task = "funloc"
 ch_types = ["meg", "eeg"]
 data_type = "meg"
+# TODO: If not supplied, should probably be the default if use_maxwell_filter is True...
+reader_extra_params = dict(allow_maxshield="yes")
```
drammock: why is this "TODO"?  in `mne_bids_pipeline/_config.py` the new config docs say it will be used automatically...

drammock (approved):

PR #1230, opened 2026-05-04
bump VTK in lockfile

drammock:
This is a band-aid: it bumps VTK in the lockfile to a Py3.14-compatible version


larsoner:
I was doing a tiny bit of docs searching for this before you said you were going to take a crack at it... for posterity can you say what command(s) you used? 

Eventually would be good to put more stuff in in https://github.com/mne-tools/mne-bids-pipeline/blob/main/.github/workflows/autopush.yml if needed

larsoner (approved):

larsoner:
(like it would be good if that action kept at the very least all the formatting stuff up to date as possible... and at least every 3-6 months did whatever was needed to keep all the rest of the stuff as up-to-date as possible ...)

drammock:
for this PR what I did was
```sh
rm -rf .venv  # probably not necessary
git restore uv.lock  # get back to state on main
uv --upgrade-package vtk
uv sync  # to get .venv back
```

PR #1241, opened 2026-05-14
Fix a small naming glitch for CSP in the report

hoechenberger:
E.g. in `main` we would produce:

```
CSP decodingresponse/incorrect vs. response/correct
```

With this PR, we produce:

```
CSP decoding: response/incorrect vs. response/correct
```

### Before merging …


larsoner:
Thanks @hoechenberger !

Issue #1229, opened 2026-05-04
reconsider lockfile / lefthook fail


drammock:
`uv.lock` specifies vtk 9.5.2, which doesn't have python 3.14 wheels.  Using any tool besides `uv` yields an environment with Python 3.14 --- which causes the `lefthook` precommit checks to fail because they can't resolve the env.  Ideally for a quick PR I shouldn't need to worry about the lockfile at all... but I also shouldn't be forced to use `uv` if I'm happier or more familiar with other tools.

(side note `#1`: this illustrates one advantage of pixi over uv: pixi lockfiles can specify the python version too)

(Side note `#2`: this really underscores the pain caused by lack of a contributor guide.  If it said somewhere "we use UV, you must use UV if you want a good dev/contrib experience" then OK; I may not be happy but at least it's a *stated* choice).

larsoner:
Also I've noticed some stuff where the pre-commit hook does not fail out if it reformats, it allows me to commit with un-formatted changes. Then the `autofix` bot pushes a commit. This is less standard / less convenient than what `pre-commit` does, which is fail the commit step, which you can then re-run to allow it to succeed.

Okay with me to switch back to `pre-commit.ci`, or fix `lefthook` + UV setup here. I vote +0.5 for trying to fix the lefthook+uv setup since if it works it should be faster and might be the way of the future.

drammock:
update: even *without* using other tools, I still can't get it to work. I get this error even with no conda/pixi env active, where the default python on the path is 3.11:

```
error: Distribution `vtk==9.5.2 @ registry+https://pypi.org/simple` can't be installed because it doesn't have a source distribution or wheel for the current platform
hint: You're using CPython 3.14 (`cp314`), but `vtk` (v9.5.2) only has wheels with the following Python ABI tags: `cp310`, `cp311`, `cp312`, `cp313`
```

so I try `rm -rf .venv && uv sync` and it says "you're using Python 3.14..." followed by the same error as above 😞 

Overhaul coming

larsoner:
Looks like for my bug / behavior all we need is 
```
  fail_on_changes: "non-ci"
```
in `.lefthook.yml`, assuming you keep it!

hoechenberger:
#1243 explicitly requests Python 3.13 for dev now.

FWIW you'll typically want to run `uv sync --locked` and not plain `uv sync` unless you're looking for a potential lockfile update.

drammock:
[...]

PR #1243, opened 2026-05-14
Pin Python version to 3.13 for dev

hoechenberger:
Ensures the `uv` venv uses Python 3.13

Addresses part of #1229, I suppose

### Before merging …


larsoner:
Given the MNE-Installers were just able to get everything installed for 3.14, we should probably transition to that anyway. But this seems like a useful fix in the meantime, thanks @hoechenberger ! I restarted the failed CI and am marking for merge-when-green

PR #1244, opened 2026-05-14
Bump UV and CircleCI to 3.14

larsoner:
I just increased the `.python-version`. I then modified `autofix.ci` to run `uv lock --upgrade` for now in this PR. Assuming it works (it should push an update to `uv.lock`), I'll then modify it to run this once a month on a schedule, and open a PR if there are updates.


larsoner:
Good, it failed successfully! It tried to open a PR to push the changes. I'll go ahead and mark this for merge-when-green, and then once it lands I'll `workflow_dispatch` it, which will then hopefully open a PR to update `uv.lock`.

PR #1266, opened 2026-07-08
Add terminal title

larsoner:
Make it easier to see at a glance where you are in the process.

Screenshot From 2026-07-08 08-28-18

### Before merging …


drammock:
how does this interact with n_jobs > 1 ? I worry it might get rather flickery

larsoner:
It operates at the step level, so it occurs once per step, regardless of `n_jobs` (the title is never set *inside* something that is parallelized)

PR #1274, opened 2026-07-22
Switch to mne-tools scripts

tsbinns:
`setup_xvfb.sh` and `check_qt_import.sh` tools are being moved from MNE-Python (https://github.com/mne-tools/mne-python/pull/13971) to MNE-Tools (https://github.com/mne-tools/mne-tools).

This PR makes it so the scripts are accessed in the CIs from their new home.


larsoner:
Beautiful, thanks @tsbinns !

PR #1282, opened 2026-08-02
Clean up documented options code and options passing

larsoner:
`_ParseConfigSteps` has steadily grown more and more unwieldy because it's a bit of a hodgepodge of functions to look at, plus which those call, etc. This started out as a plain refactor (facilitated by Claude Fable 5) to recurse through all function calls with a `seen` set. However, once that was done, it became simpler and easier to traverse stuff we ignored before, like `_bids_kwargs`. Then it became easier to split these up and limit them. So now this PR also limits the config options passed to options much better than it did before, coming closer to the ideal "options are passed to steps if and only if they are (potentially) used by the given step". This turned up a number of violations, now fixed.

In theory this should come back green... :crossed_fingers:

### DIFF
--- a/.circleci/run_dataset_and_copy_files.sh
+++ b/.circleci/run_dataset_and_copy_files.sh
@@ -47,6 +47,13 @@ else
 fi
 test $RUN_TIME -le $RERUN_LIMIT
 
+# Most of the pipeline runs in loky/dask worker processes, which write their own
+# coverage data files (see [tool.coverage.run] in pyproject.toml). Merge them into
+# the main one before the upload, or everything executed in a worker looks
+# uncovered. --append is required so the parent process's data is kept too.
+coverage combine --append || true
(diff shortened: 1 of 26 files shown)


larsoner:
More red than green ❤️ plus green CIs so I think we're good to go here!

... but I want to look to see if there is any good deduplication of `_import_data_kwargs`, `_bids_kwargs`, and `_raw_path_kwargs` that we can do. I doubt it but I want to check...

larsoner:
Okay I ended up including some coverage fixes too (ensure that coverage is collected on parallel runners). @drammock did you want to look? If not I think we can trust CIs to ensure things aren't broken (our coverage is actually quite good when you measure it properly!)

drammock:
Will try to look on Monday, if I don't manage it feel free to merge after that

PR #1284, opened 2026-08-05
Handle decoding contrasts with too few epochs

### LINKED ISSUE #1265: cross_val_multiscore can crash with too few epochs
I have a dataset with only 14 epochs and it crashes when calling `cross_val_multiscore` in `decoding_time_by_time` because there aren't enough epochs of each type:
```
(Pdb) w
...
> /opt/mne-bids-pipeline/mne_bids_pipeline/steps/sensor/_03_decoding_time_by_time.py(201)run_time_decoding()
-> scores = cross_val_multiscore(
  <decorator-gen-558>(10)cross_val_multiscore()
  /opt/mne-python/mne/decoding/base.py(880)cross_val_multiscore()
-> scores = parallel(
  /opt/mne-python/mne/decoding/base.py(881)<genexpr>()
-> p_func(
  <decorator-gen-559>(10)_fit_and_score()
  /opt/mne-python/mne/decoding/base.py(967)_fit_and_score()
-> test_score = _score(estimator, X_test, y_test, scorer)
  /opt/mne-python/mne/decoding/base.py(992)_score()
-> score = scorer(estimator, X_test, y_test)
  /opt/.../sklearn/metrics/_scorer.py(472)__call__()
-> return estimator.score(*args, **kwargs)
  /opt/mne-python/mne/decoding/search_light.py(333)score()
-> y = _fix_auc(scoring, y)
  /opt/mne-python/mne/decoding/search_light.py(771)_fix_auc()
-> raise ValueError(
(Pdb) p y
array([1., 1., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0.])
```

1. Should figure out why this step *doesn't* fail for the previous decoding step.
2. Should do something reasonable here -- probably set the scores to `nan`
3. Ensure whatever we do in (2) is handled properly by the `average` task

viranovskaya:
Closes #1265.

If either condition contains fewer epochs than `decoding_n_splits`, the subject-level step now saves all-NaN scores with the expected output shape and logs a clear warning instead of running invalid cross-validation.

At the group level, results that are entirely NaN are excluded from the statistics. The saved output and report use the effective sample size and retain the valid subject IDs. Cluster permutation testing is skipped when fewer than two valid subjects remain.

Regression tests cover:

- NaN score shapes for full-epoch, time-by-time, and time-generalization decoding;
- the sufficient-epoch path;
- exclusion of all-NaN subject results while retaining partial-NaN results;
- reports with contrast-specific effective sample sizes.

### DIFF
--- a/docs/source/dev.md
+++ b/docs/source/dev.md
@@ -15,6 +15,7 @@
 
 ### :bug: Bug fixes
 
+- Handle contrasts with too few epochs for cross-validation by saving NaN scores, excluding invalid subject-level results from group statistics, and reporting the effective sample size (#1265 by @viranovskaya)
 - Raise an informative error if [`rest_epochs_duration`][mne_bids_pipeline._config.rest_epochs_duration] is not set for resting-state data and document the parameter (#1272 by @viranovskaya)
 - Fixed bug where [`log_level`][mne_bids_pipeline._config.log_level] was not being applied to the MBPlogger (#1224 by @larsoner)
 - Corrected import order: remove channels before setting template montage as stated in [`eeg_template_montage`][mne_bids_pipeline._config.eeg_template_montage] (#1220 by @dnacombo)
--- a/mne_bids_pipeline/_decoding.py
+++ b/mne_bids_pipeline/_decoding.py
@@ -33,6 +33,27 @@ class LogReg(LogisticRegression):  # type: ignore[misc]
             return super().fit(*args, **kwargs)
 
 
+def _get_nan_decoding_scores_if_insufficient(
+    *,
+    n_cond1: int,
+    n_cond2: int,
+    n_splits: int,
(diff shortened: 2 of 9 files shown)


larsoner:
Thanks for looking into this!

400+ new lines of tests for this seems like a lot. I was hoping for more like 50-100. Do you see a way to make them more compact?

viranovskaya:
Yes, that makes sense. I replaced the large test suite with focused parameterized tests. The test file is now 84 lines instead of 469 and still covers the full-epoch, time-by-time and time-generalization score shapes, all-NaN subject filtering, and contrast-specific effective sample sizes. All CI checks pass now.

larsoner (approved):
Looks workable now, thanks for the fix @viranovskaya !

PR #1291, opened 2026-08-12
ENH: Add flow diagram to reports

larsoner:
### Before merging …

Iterated a bunch with Claude Fable 5 on a feature I've wanted [for a while](https://github.com/mne-tools/mne-bids-pipeline/issues/350#issuecomment-1307838525): a flow diagram calculated on the fly based on what actually ran:

Screenshot 2026-08-12 at 10 41 00 AM

The mechanics behind how to construct it are pretty simple:

1. Write a JSON result for each step that can then later be parsed.
2. Read those JSONs during report phase and construct a directed acyclic graph (DAG)
3. Lay that out using custom DAG layout + SVG + a little JS for interaction

<details>
<summary><strong>Why a custom-rolled graph layout</strong></summary>

(3) above turned out to be the tricky decision point, because there are libraries for this sort of thing. But there were requirements I had:

1. Right graph type: DAG + labeled edges + multi-parent nodes.
2. Self-contained, offline-safe output.
3. No new runtime burden: pip-installable pure Python for a tool users run on their own machines.

I looked at maybe a dozen other pieces of software instead, top alternative candidates were:

- ⛔ `mermaid` (js): adds ~3.5 MB of vendored JS into every report and lacks some features (like label mouseover support that SVG gives for free)
- ⛔ `grandalf` (python): only replaces the "layout" part of the code, which is not the hardest part to sort out (and adds a dep)
- ⛔ `pygal` (python): no node-link chart type
- ⛔  `chart.js` (js): no text search, no tooltips
- ❓ `d3-dag` (js): the closest. layout-only in the browser, though; trades ~240 lines of Python rendering for embedded JavaScript.

I had Claude implement `d3-dag` and `mermaid` and they at least worked, but weren't overall better (edge labels worse for d3, mermaid no mouseovers, etc.):

Screenshot 2026-08-12 at 2 13 42 PM

If our graphs get far more complex at some point it could make sense to go with `d3-dag` probably, though.

</details>

### DIFF
--- a/docs/source/dev.md
+++ b/docs/source/dev.md
@@ -6,6 +6,7 @@
 
 - Added [`ignore_warnings`][mne_bids_pipeline._config.ignore_warnings] config option to allow users to specify warnings to ignore when calling `read_raw_bids` (#1224 by @larsoner)
 - Added tracking of the current step in the terminal title (#1266 by @larsoner)
+- Added a "Pipeline flow" section to the reports with an auto-generated diagram of the steps that ran for a subject and the files they passed to one another (#1289 by @larsoner)
 
 ### :warning: Behavior changes
 
--- a/docs/source/dev.md
+++ b/docs/source/dev.md
@@ -6,7 +6,7 @@
 
 - Added [`ignore_warnings`][mne_bids_pipeline._config.ignore_warnings] config option to allow users to specify warnings to ignore when calling `read_raw_bids` (#1224 by @larsoner)
 - Added tracking of the current step in the terminal title (#1266 by @larsoner)
-- Added a "Pipeline flow" section to the reports with an auto-generated diagram of the steps that ran for a subject and the files they passed to one another (#1289 by @larsoner)
+- Added a "Pipeline flow" section to the reports with an auto-generated diagram of the steps that ran for a subject and the files they passed to one another (#1291 by @larsoner)
 
 ### :warning: Behavior changes
 
--- a/docs/source/features/gen_steps.py
+++ b/docs/source/features/gen_steps.py
@@ -4,7 +4,8 @@
 import importlib
 from pathlib import Path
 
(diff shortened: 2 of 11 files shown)


larsoner:
On [mouseover](https://output.circle-artifacts.com/output/job/b122e26d-9eec-4dee-b8a7-5d463f67ed78/artifacts/0/site/examples/ds000248_base/sub-01_report.html#Pipeline_flow):

Screenshot 2026-08-12 at 6 47 58 PM

drammock:
Love this. It would be nice to avoid how the source steps are weaving in and out of the preprocessing ones (in your mouseover screenshot). Given that (for now at least) we know the identity of all the steps, could we pre-bake a layout that (1) avoids crossing arrows, and (2) doesn't jumble up the steps from different pipeline stages?
Or, can we add layout constraints that would have that effect (at least for item 2)? One possible approach would be to define subgraphs/groups of nodes and constrain the layout of those groups relative to each other.  Another solution (which I think is available in mermaid, I've definitely encountered the option somewhere) is to constrain certain nodes to be laid out below/right-of (for vert/horiz graphs) certain other nodes, regardless of whether they are connected by an edge. But simply pre-baking the layout and omitting/graying out any nodes that weren't used might not look too bad. WDYT?

larsoner:
Okay implemented the second option to try layout constraints, seems better than hand-baking and didn't add that many more lines. Then went through a few rounds of iterations to tighten things further. Here's the `ds000248_base` [example now](https://output.circle-artifacts.com/output/job/0140d226-8134-4f49-9ad6-e39147be4a08/artifacts/0/site/examples/ds000248_base/sub-01_report.html#Pipeline_flow):

Screenshot 2026-08-15 at 7 18 27 PM

drammock:
To me that seems better! +1 for merge if you agree (or don't care)

Issue #1290, opened 2026-08-12
Implement OTP as a preprocessing step for pipeline


nordme:
It would be great to have oversampled temporal projection included as a preprocessing step covered by mne-bids-pipeline. Currently, this is not an option.

PR #1297, opened 2026-08-26
ENH: Add OTP to preprocessing steps

nordme:
Addresses issue #1290 

### Before merging …


nordme:
We're also going to fix the documentation for downloading examples to make sure that we're pulling working versions of the example datasets (specifically ds003104, since version 1.0.1 is failing on the Open Neuro end as of Aug 26, 2026).

mne_bids_pipeline/tests/configs/config_ds003392_otp_mxw.py
```diff
@@ -0,0 +1,54 @@
...
+otp_duration = 10.0
+# for OTP testing purposes, set to True
+find_flat_channels_meg = True
+find_noisy_channels_meg = True
+use_maxwell_filter = True
+mf_extra_kws = {"bad_condition": "warning"}
+ch_types = ["meg"]
+mf_esss = 1
+# translation args should be x, y, z as int
+# mf_destination = translation(z=0.04)
```
sylvchev: I commented out this line as mne.Translations.translation expects int, I may have missed the point of this translation. If we don't need it, we could remove the commented lines.

mne_bids_pipeline/_config.py
```diff
@@ -602,6 +602,36 @@
...
+# ## Oversampled Temporal Projection (OTP)
+
+use_otp_denoising: bool = False
+"""
+Whether or not to use [Oversampled Temporal Projection]
+[mne.preprocessing.oversampled_temporal_projection] to preprocess the data.
+
+!!! warning
+    If the data will be processed using Maxwell filtering, OTP must be applied first and
+    the application of OTP will alter the correlation for tSSS (0.99 or higher). Flat
```
drammock: say what the user should do (i.e., "set the tsss correlation to 0.99 or higher")

mne_bids_pipeline/_config.py
```diff
@@ -602,6 +602,36 @@
...
+    and noisy channels should be marked as bad or else the noise will be spread to all
+    channels because of OTP. Bad channels need to be marked in the raw files or set
+    through BIDS channels.tsv via the automatic bad channel detection functions
+    before applying OTP.
```
drammock: - must mark bads before OTP
- the built-in flat- and noisy-marking config options will run before OTP if they're enabled. (crossref to them)
- alternatively you can manually mark bads in `channels.tsv`

mne_bids_pipeline/tests/configs/config_ds003392_otp_mxw.py
drammock: re-enable this, float should be OK

mne_bids_pipeline/steps/preprocessing/_07_make_epochs.py
```diff
@@ -121,7 +122,11 @@ def run_epochs(
     for idx, (run, raw_fname) in enumerate(zip(cfg.runs_for_task, raw_fnames)):
         msg = f"Loading filtered raw data from {raw_fname.basename}"
         logger.info(**gen_log_kwargs(message=msg))
-        raw = mne.io.read_raw_fif(raw_fname).load_data()
+        with warnings.catch_warnings(record=True):
+            warnings.filterwarnings(
+                "ignore", ".*Internal Active Shielding", category=RuntimeWarning
+            )
+            raw = mne.io.read_raw_fif(raw_fname, allow_maxshield=True).load_data()
```
drammock, line 129: same question here, can we use `allow_maxshield="yes"` here?

mne_bids_pipeline/steps/preprocessing/_07_make_epochs.py
```diff
@@ -340,8 +345,12 @@ def _get_events(
...
         this_raw_fname = raw_fname.copy().update(run=run)
         this_raw_fname = _update_for_splits(this_raw_fname, None, single=True)
-        raw_filt = mne.io.read_raw_fif(this_raw_fname)
-        raws_filt.append(raw_filt)
+        with warnings.catch_warnings(record=True):
+            warnings.filterwarnings(
+                "ignore", ".*Internal Active Shielding", category=RuntimeWarning
+            )
+            raw_filt = mne.io.read_raw_fif(this_raw_fname, allow_maxshield=True)
+            raws_filt.append(raw_filt)
```
drammock, line 353: same question here, can we use `allow_maxshield="yes"`?

mne_bids_pipeline/_import_data.py
```diff
@@ -211,11 +212,15 @@ def annotations_to_events(*, raw_paths: list[BIDSPath]) -> dict[str, int]:
...
-        raw = mne.io.read_raw_fif(raw_fname)
-        _, event_id = mne.events_from_annotations(raw=raw)
-        for event_name in event_id.keys():
-            if event_name not in event_names:
-                event_names.append(event_name)
+        with warnings.catch_warnings(record=True):
+            warnings.filterwarnings(
+                "ignore", ".*Internal Active Shielding", category=RuntimeWarning
+            )
+            raw = mne.io.read_raw_fif(raw_fname, allow_maxshield=True)
```
drammock, line 219: @larsoner any reason we shouldn't just pass `allow_maxshield="yes"` here?

mne_bids_pipeline/steps/preprocessing/_04_frequency_filter.py
```diff
@@ -209,14 +210,15 @@ def filter_data(
     out_files = dict()
     in_key = f"raw_task-{task}_run-{run}"
     bids_path_in = in_files.pop(in_key)
-    if bids_path_in.processing == "sss":
+    if bids_path_in.processing == "sss" or bids_path_in.processing == "otp":
```
drammock:
```suggestion
    if bids_path_in.processing in ("sss", "otp"):
```

mne_bids_pipeline/steps/preprocessing/_04_frequency_filter.py
```diff
@@ -209,14 +210,15 @@ def filter_data(
...
         bids_path_bads_in = None
     else:
         bids_path_bads_in = in_files.pop(f"{in_key}-bads")
     msg, run_type = _read_raw_msg(bids_path_in=bids_path_in, run=run, task=task)
     logger.info(**gen_log_kwargs(message=msg))
-    if cfg.use_maxwell_filter:
-        raw = mne.io.read_raw_fif(bids_path_in)
+    if cfg.use_maxwell_filter or cfg.otp:
+        allow_ms = "True" if cfg.otp else "False"
+        raw = mne.io.read_raw_fif(bids_path_in, allow_maxshield=allow_ms)
```
drammock:
```suggestion
        raw = mne.io.read_raw_fif(bids_path_in, allow_maxshield=cfg.otp)
```

mne_bids_pipeline/steps/preprocessing/_03_maxfilter.py
```diff
@@ -65,6 +65,8 @@ def get_input_fnames_all_bads(
 ) -> dict[str, BIDSPath]:
     """Get paths of files required by compute_all_bads function."""
     in_files: dict[str, BIDSPath] = dict()
+    # use_kind = "otp" if cfg.otp else "orig"
+    use_kind = "orig"
```
drammock: cruft?

mne_bids_pipeline/steps/preprocessing/_03_maxfilter.py
```diff
@@ -85,20 +87,25 @@ def _get_allbads_path(
     subject: str,
     session: str | None,
 ) -> BIDSPath:
+    # use_kind = "otp" if cfg.otp else "orig"
+    # use_proc = "otp" if cfg.otp else None
+    use_kind = "orig"
+    use_proc = None
```
drammock: cruft?

drammock (commented):
reviewed collaboratively with @nordme so the comments are rather terse...

mne_bids_pipeline/tests/configs/config_ds003392_otp_mxw.py
```diff
@@ -26,8 +26,8 @@
 mf_extra_kws = {"bad_condition": "warning"}
 ch_types = ["meg"]
 mf_esss = 1
-# translation args should be x, y, z as int
-# mf_destination = translation(z=0.04)
+# translation args should be x, y, z as scaler (int or float ok)
```
drammock:
```suggestion
# translation args should be x, y, z as scalar (int or float)
```

mne_bids_pipeline/tests/test_documented.py
```diff
@@ -197,7 +197,7 @@ def test_datasets_in_doc() -> None:
     # Rather than going circle_yaml['workflows']['commit']['jobs'] and
     # make sure everything is consistent there (too much work), let's at least
     # check that we get the correct number using `.count`.
-    counts = dict(ERP_CORE=7, ds000248=6, ds003392=4)
+    counts = dict(ERP_CORE=7, ds000248=6, ds003392=3)
```
drammock, line 202:
```suggestion
    # ↓↓↓↓ number of cache restores for each dataset
    # ↓↓↓↓ (some datasets used with multiple configs)
    counts = dict(ERP_CORE=7, ds000248=6, ds003392=3)
```

drammock (approved):

nordme:
@drammock It looks like the Github actions config really wants the dataset cache name and the dataset test name to be the same. Can we just use a different dataset altogether? If I'm reading this right, the point of the two included data sets is that we need one hash-based and one mtime-based config. ds003104 uses mtime rather than hashing, and it looks relatively fast to run the tests. I think the only reason ds003392 was used originally was that it used mtime and was faster than others, but now we're loading it with three tests, it's a bad option, so maybe we should just dump it for the os testing?

.github/workflows/tests.yml
```diff
@@ -43,15 +43,17 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, ubuntu-24.04-arm, macos-latest, windows-latest]
-        dataset: [ds001971, ds003392]  # one uses "hash", the other "mtime"
+        dataset: [ds001971, ds003392_base]  # one uses "hash", the other "mtime"
```
larsoner: Change this to one that's faster if possible, and add a little comment above saying we picked these two because they are quick

.github/workflows/tests.yml
```diff
@@ -43,15 +43,17 @@ jobs:
...
         include:
           - os: ubuntu-latest
             dataset: ds001971
             python: "3.11"
           - os: ubuntu-latest
-            dataset: ds003392
+            dataset: ds003392_base
             python: "3.13"
+          - os: ubuntu-latest
+            dataset: ds003392_otp
```
larsoner: No need for an additional one I think, this action is mostly to test the hashing vs mtime

larsoner:
For conflicts, you can just take the ones from `main` and modify!

PR #1300, opened 2026-09-01
Add report_image_format option

larsoner:
This is a seemingly trivial option... add a `report_image_format` option.

It turns out `webp` compression is a non-trivial amount of time in our test suite (like over 20% for some jobs locally). If it's that way in our test suite, it's possible it is in some people's pipelines, so let's add it as an option for people. And then let's use it in most tests where a few extra MB of a report is worth having our run take less time. (And the slight increase in size in the HTMLs landing in our docs are acceptable, too, I think.)

Also tacks on a report efficiency enabled by https://github.com/mne-tools/mne-python/pull/14261


drammock (approved):

larsoner:
This sped up the 248 T1_BEM job from ~9m to ~5m (for the actual `pytest` step) because so much time was spent creating the BEM image plots.

PR #1301, opened 2026-09-01
Use gen3 workers in CircleCI

larsoner:
CircleCI has workers that are faster and cheaper (how rare!) so let's use them.


larsoner:
Oooh this one has a tricky deadlock... I'll see if I can fix it. I suspect it's report<->flow maybe but we'll see!

larsoner:
I've seen the deadlock before, I have a fix in progress in loky/joblib

PR #1305, opened 2026-09-03
Speed up caching and reruns

larsoner:
I noticed reruns could still be a bit slow, so let's:

1. Remove some unneeded I/O and use faster I/O when possible for JSON
2. Nest some heavy imports so they're only used if the step actually runs
3. Use the `webp-lossy` from https://github.com/mne-tools/mne-python/pull/14249 to reduce runtime and HTML size
4. Take care of a TODO comment about not repeating filters for CSP (speeds up that step a lot)

I'll check to make sure things still look okay in the rendered docs. Changes drafted by Claude Fable 5 and reviewed by me.

This one is a pretty simple CI speedup so should be safe enough for self-merge when green


larsoner:
The full CircleCI run is down to 14m :face_holding_back_tears:

PR #1307, opened 2026-09-08
ENH: Add support for SLURM

### LINKED ISSUE #1160: Running pipeline via dask on SLURM cluster
I'm currently trying to setup `dask` for parallelization. We have an HPC with SLURM.

As far as I can see does not seem possible to set custom options for parallelization for `dask`, is that correct?

How would I go about setting something like

```python
from dask_jobqueue import SLURMCluster
cluster = SLURMCluster(
    queue='cpu',
    cores=8,
    memory='32GB',
    walltime='02:00:00',
    interface='ib0',
    scheduler_options={
        'host': '172.xxx.xxx.xxx',
        'dashboard_address': ':8787',
        'port': 5000  # Use a fixed port
    }
)

```

Is that even possible?

larsoner:
Closes #1160

Adds support for SLURM and HPC via [`dask-jobqueue`](https://jobqueue.dask.org/en/latest/). This makes it pretty simple actually! The only real tricky part is the I/O handling to get `rich` to log the dask outputs correctly. Getting CircleCI to set up a SLURM cluster isn't too bad either (~40 lines of bash). Implemented by Claude Fable 5 and reviewed by and then iterated on by me.

### DIFF
--- a/.circleci/cgroup.conf
+++ b/.circleci/cgroup.conf
@@ -1,0 +1,2 @@
+# No resource confinement in CI; "disabled" needs slurm >= 24.05 (26.04 has 25.11)
+CgroupPlugin=disabled
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -914,8 +914,17 @@ jobs:
           command: |
             google-chrome --version
       - run:
-          name: test ERP CORE N400
+          name: Start single-node SLURM
+          command: ./.circleci/setup_slurm.sh
+      - run:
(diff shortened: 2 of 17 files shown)


larsoner:
Okay let's get this in so people can try it if they want -- I think most bugs will come out in actual use cases (and hopefully are limited since this mostly relies on `dask-jobqueue` to take care of things for us!).

larsoner:
(marking for merge-when-green after fixing a merge conflict)

